PR #3, opened 2016-11-17
Initial structure

bmallred:
The initial structure during design phase of the project working out any found kinks or roadblocks. Includes such areas as the following:

 - Docker compose
 - ReactJS front-end structure and unit testing
 - API back-end initially using Golang
 - Continuous integration scripts


api/auth.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// configureAuthentication takes a service name and configures the OAuth 2.0 with
+// appropriate endpoints and scopes.
+func configureAuthentication(service string) (*oauth2.Config, bool) {
+	ok := true
+	config := &oauth2.Config{
+		RedirectURL:  fmt.Sprintf("http://localhost:3000/auth/%s/callback", strings.ToLower(service)),
+		ClientID:     os.Getenv(fmt.Sprintf("%s_CLIENT_ID", strings.ToUpper(service))),
+		ClientSecret: os.Getenv(fmt.Sprintf("%s_CLIENT_SECRET", strings.ToUpper(service))),
```
adelevie: Still new to golang, but this is neat.

api/auth.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
```
adelevie: Interesting stuff here to use OAuth info to potentially pre-populate the form. This will likely get figured out with user research and product discussions, but we should be sure 
1) this is Ok from a policy perspective; and
2) these scopes could be made optional

api/docker-run.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh
+
+echo "--------------------------------------"
```
adelevie: We should delete this. Ok for now.

.travis.yml
```diff
@@ -1,3 +1,38 @@
...
+
+before_install:
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
+
+script:
+  - docker-compose build
+  - docker-compose run api ./docker-travis.sh
```
adelevie: Eventually, this should be changed to `docker-compose run api go test`. Ok for now.

api/docker-travis.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh
+
+echo "--------------------------------------"
```
adelevie: Should be deleted. Ok for now.

api/middleware/middleware.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	return mr
+}
+
+// Subrouter creates a new Middleware router based on the route.
+func (mr *MiddlewareRoute) Subrouter() *Middleware {
+	return New(mr.Route.Subrouter()).Inject(mr.parent.handlers...)
+}
+
+// ServeHTTP can be passed to the http.ListenAndServe() function.
```
adelevie: These code commends are very helpful.

docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+  build: .
+  dockerfile: Dockerfile.yarn
+  env_file:
+    - .env
+  volumes:
+    - .:/usr/src/app
+    - /usr/src/app/node_modules
+  command: yarn watch
+api:
+  image: golang
```
adelevie: Eventually, I think we'll want this pulled out into a separate Dockerfile (which will handle dependency installation).

PR #4, opened 2016-12-01
Release #1 12/1/2016

adelevie:


api/cf/services.go
```diff
@@ -0,0 +1,65 @@
+package cf
```
adelevie: Just learned that https://github.com/cloudfoundry-community/go-cfenv is a thing. Totally fine with this for now, but consider swapping this out in favor of `go-cfenv` in the future.

api/db/db.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"log"
+	"os"
+
+	pg "gopkg.in/pg.v5"
+)
+
+func NewDB() *pg.DB {
+	user := os.Getenv("DATABASE_USER")
```
adelevie: This will eventually need to read from VCAP_SERVICES in prod and staging (and from env vars in local/test).

api/twofactor/twofactor.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// The template is stored in markdown format to easily
+	// transform something human readable to HTML as well
+	var plain bytes.Buffer
+	err := templateEmail.Execute(&plain, code)
+	if err != nil {
+		return err
+	}
+	message.Text = plain.String()
+	unsafe := blackfriday.MarkdownCommon(plain.Bytes())
```
adelevie: What is bluemonday and blackfriday?

bin/xsd2md.go
```diff
@@ -0,0 +1,175 @@
+package main
```
adelevie: What does this file do?

db/1-schema.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE IF NOT EXISTS ACCOUNTS (
```
adelevie: We need to chat to discuss how to run "migrations" ... Ideally a schema change should be applied automatically with each new deploy. Maybe we're already set up for that?

adelevie:
LGTM. @bmallred, and anyone else, lmk if you want to take a pass at additional review. Otherwise, I'll merge.

api/db/db.go
bmallred: This will probably be completely refactored next sprint to allow for multiple database types as well.

api/twofactor/twofactor.go
bmallred: `bluemonday` is an HTML sanitizer
`blackfriday` is a Markdown processor to HTML or other outputs

bin/xsd2md.go
bmallred: This takes the XSD files given by the product owner and transforms it to easier to read Markdown as well as templates what the components may look like.

Couldn't check in the sources or artifacts due to sensitivity.

db/1-schema.sql
bmallred: Definitely needs to be considered. I think the goal was to get something out there this first cycle :)

manifest-api-staging.yml
```diff
@@ -0,0 +1,10 @@
+---
+applications:
+- name: eqip-prototype-api-staging
+  memory: 1G
```
bmallred: Definitely :) The API code base should be very light

bmallred:
@adelevie, I am pinging @floresj to take a look and add any comments if need be from our side.

api/db/db.go
floresj: Yep this will definitely change.

db/1-schema.sql
floresj: The intent first pass was to get sql artifacts loaded into postgres for dev purposes. Currently, you have to perform a `docker-compose rm db` and re-run to get the .sql files loaded. However, we'll definitely need a more formal process to handle structural db updates.

floresj:
@adelevie All good on my end.

PR #5, opened 2016-12-02
Development sprint 1

bmallred:
This is the pull request concerning the `NODE_ENV` environment variable. On TravisCI for the staging repository we just added it to the TravisCI settings (even though it is not sensitive). We can work out a better place for it, but as long as the environment variable is available to the following command it should be okay:

```
docker-compose run frontend npm run build --silent
```

found in [.travis.yml](https://github.com/18F/e-QIP-prototype/blob/master/.travis.yml#L27)

Additional conversation at [PR #86](https://github.com/truetandem/e-QIP-prototype/pull/86)


bmallred:
This is the pull request concerning the `NODE_ENV` environment variable. On TravisCI for the staging repository we just added it to the TravisCI settings (even though it is not sensitive). We can work out a better place for it, but as long as the environment variable is available to the following command it should be okay:

```
docker-compose run frontend npm run build --silent
```

found in [.travis.yml](https://github.com/18F/e-QIP-prototype/blob/master/.travis.yml#L27)

Additional conversation at [PR #86](https://github.com/truetandem/e-QIP-prototype/pull/86)

adelevie:
> On TravisCI for the staging repository we just added it to the TravisCI settings (even though it is not sensitive).

I'd prefer this get checked in to the repo. The fewer env vars we have to track manually, the better.

bin/deploy.sh
```diff
@@ -4,33 +4,26 @@ API="https://api.fr.cloud.gov"
...
   FRONTEND_NAME="eqip-prototype-staging"
   FRONTEND_MANIFEST="manifest-frontend-staging.yml"
-
-  cf zero-downtime-push $API_NAME -f $API_MANIFEST
-  cf zero-downtime-push $FRONTEND_NAME -f $FRONTEND_MANIFEST
 else
   echo "Unknown space: $SPACE"
   exit
 fi
+
```
adelevie: nice refactor ... did you confirm this works on staging?

src/config/environment.js
```diff
@@ -0,0 +1,16 @@
+class Env {
+  ApiBaseURL () { return 'http://localhost:3000' }
```
adelevie: Eventually turn this into an env var/VCAP_SERVICES value.

src/config/environment.production.js
```diff
@@ -0,0 +1,16 @@
+class Env {
+  ApiBaseURL () { return 'https://eqip-prototype-api.fr.cloud.gov' }
```
adelevie: Thoughts on turning this into config?

src/config/environment.staging.js
```diff
@@ -0,0 +1,16 @@
+class Env {
+  ApiBaseURL () { return 'https://eqip-prototype-api-staging.fr.cloud.gov' }
```
adelevie: Same here.

src/services/api.js
```diff
@@ -1,24 +1,25 @@
 import axios from 'axios'
+import { env } from '../config'
 
 class Api {
   constructor () {
     this.proxy = axios.create({
-      baseURL: 'http://localhost:3000',
+      baseURL: env.ApiBaseURL(),
```
adelevie: Very nice.

src/services/api.js
```diff
@@ -51,38 +52,38 @@ class Api {
...
   }
 
   validateState (state) {
-    return this.proxySecured.get(`/validate/address/${state}`)
+    return this.proxySecured.get(env.EndpointValidateState(state))
   }
 
   validateAddress (address) {
-    return this.proxySecured.post('/validate/address', address)
+    return this.proxySecured.post(env.EndpointValidateAddress(), address)
```
adelevie: Good use of the `env` module (or class?)

src/config/environment.production.js
bmallred: The Javascript is compiled and not sure we have access to environment variables in a static `.js` file.

Are we running this on a NodeJS instance? May have access to it then.

bin/deploy.sh
bmallred: Yeah looks good according to build [230](https://travis-ci.org/truetandem/e-QIP-prototype/builds/180746513)

src/services/api.js
bmallred: `env` is the exported instantiation (ah words) of the `Env` class over at `src/config/environment.js`.

.travis.yml
```diff
@@ -10,6 +10,7 @@ before_install:
   - sudo mv docker-compose /usr/local/bin
 
 script:
+  - sed -i "s/NODE_ENV=/NODE_ENV=$(echo $NODE_ENV)/" .env
```
adelevie: Sorry to be late to this --- what does this command do? I'm not experienced with `sed`.

bmallred: Replaces the line in `NODE_ENV=` in `.env` with whatever the TravisCI environment variable is set to so docker builds the for the right environment.

The `s/` command is substitution... so `s/<what we are looking to replace>/<what we want to replace it with>/`

PR #16, opened 2017-02-09
Release #1 2/9/2017

adelevie:


src/components/Form/Collection/Collection.jsx
```diff
@@ -242,6 +242,22 @@ export default class Collection extends ValidationElement {
...
 
+    // Create the 'byline' which is used to separate items as well as the bottom toggle action
+    const bylineBottom = (item, index) => {
+      if (this.state.items.length < 2) {
+        return ''
+      }
+
+      return (
+        <div className="byline bottom">
+          <a href="javascript:;;" className="toggle" onClick={this.toggle.bind(this, index)}>
```
mgwalker: Is this `href` to satisfy accessibility scanners?  I ask because I've run into the issue before where HTML Code Sniffer complains that a `href` is empty or points to something that doesn't exist (e.g., dynamically routed stuff).

bmallred: It's a good question.

I believe the `href` attribute is optional according to written specifications, but in HTML 4.x it is considered a **hyperlink** if the attribute is present or an **anchor** if the attribute is not present. In HTML 5+ it is always considered a **hyperlink** but if the attribute is missing then it becomes just a placeholder for one.

In some scanners/linters/etc. the rules get confusing and I think the inconsistency between the two major versions is part of it. The inconsistency is also found in browsers as well. For example in some browsers if the `href` is omitted then the mouse cursor or screen reader does not recognize it as a hyperlink.

In reality we could possible not include it and it would work as intended on most newer browsers. If anyone has any input then we definitely would love to hear them!

References:
 [1]: https://www.w3.org/MarkUp/html-spec/html-spec_7.html
 [2]: http://www.w3schools.com/tags/tag_a.asp

mgwalker: 100% on board with leaving it as-is.  Making all the linters and scanners happy has always been an adventure when it comes to anchor tags, doubly so with a lot of the apps now that do routing dynamically.  Then you have something like `<a href="#next-page">` and the scanners get made that `#next-page` doesn't exist.  😱 

I didn't know all that context about when it's considered a link versus an anchor, though - that's cool.  But in any case, I was just curious if that was like the current "right way" - and regardless, I think it's something I'll adopt.  😛

bmallred: I personally use to (and still do depending on the product) use

```html
<a href="#">...</a>
```

However using hashtag routing in single page applications doesn't always play nice with that. :smile:

bmallred:
@mgwalker we went ahead and broke out the different issues and put them on the backlog. The code coverage took a drastic hit this sprint due to the heavy prototyping of the **Your history** section. **No excuses though and we will address this!**

The SVGs, from what I was told, needed to be inline in order to manipulate them with CSS 3 transitions, adjusting colors, stroke widths, etc. We have already talked about refactoring these out to a special component to hopefully make it easier to maintain because that is a concern. While we do that we'll go through and validate whether it is necessary for all of them to be done in this way.

Most of the React warnings are due to the refactoring we have been doing for the **Your history** section within the [Collection)(https://github.com/truetandem/e-QIP-prototype/blob/master/src/components/Form/Collection/Collection.jsx) component.

Below are the issues for tracking:

 - Issue [#441](https://github.com/truetandem/e-QIP-prototype/issues/441): Addressing code coverage
 - Issue [#442](https://github.com/truetandem/e-QIP-prototype/issues/442): Refactor inline SVGs to component or make reference to file path if manipulation is not needed
 - Issue [#443](https://github.com/truetandem/e-QIP-prototype/issues/443): `onValidate` references

src/components/Form/EyeColor/EyeColor.jsx
```diff
@@ -139,7 +139,7 @@ export default class EyeColor extends ValidationElement {
                  onBlur={this.props.onBlur}
                  onFocus={this.props.onFocus}
                  >
-            <div className="eye-icon">
+            <div className="eye-icon hazel">
```
bmallred: From the [comment](https://github.com/18F/e-QIP-prototype/pull/16#issuecomment-278813101) below:

> The SVGs, from what I was told, needed to be inline in order to manipulate them with CSS 3 transitions, adjusting colors, stroke widths, etc.

When we do a bit of refactoring to hopefully centralize these use cases I am optimistic some will be found to not need to be referenced in this manner.

src/components/Form/EyeColor/EyeColor.jsx
```diff
@@ -212,17 +212,17 @@ export default class EyeColor extends ValidationElement {
...
-						                                    c0.1,0.08,0.24,0.12,0.38,0.12c0.18,0,0.38-0.08,0.5-0.24c0.94-1.21,1.35-1.57,1.73-1.85c0.34-0.24,1-0.48,1.73-0.48
-						                                    c1.29,0,2.47,0.82,2.47,1.71c0,1.04-0.54,1.57-1.77,2.13c-1.43,0.64-3.38,2.31-3.38,4.26v0.72c0,0.36,0.28,0.64,0.64,0.64h3.86
-						                                    c0.36,0,0.64-0.28,0.64-0.64c0-0.46,0.58-1.45,1.53-1.99c1.53-0.87,3.62-2.03,3.62-5.08C23.15,11.28,19.29,8.53,15.84,8.53z
-						                                    M18.01,24.61c0-0.36-0.28-0.64-0.64-0.64H13.5c-0.36,0-0.64,0.28-0.64,0.64v3.86c0,0.36,0.28,0.64,0.64,0.64h3.86
-						                                    c0.36,0,0.64-0.28,0.64-0.64V24.61z"/>
-				      </svg>
-			      </div>
+            <div className="eye-icon unknown">
+              <svg viewBox="0 0 30.86 36.84">
+                <path id="eye-color-unknown" d="M15.43,34.25C6.91,34.25,0,27.34,0,18.82C0,10.3,6.91,3.39,15.43,3.39s15.43,6.91,15.43,15.43
```
bmallred: It's definitely a concern for maintainability. Also, huge eye sore making it harder to read through the bits you need. We were thinking of doing something like:

```jsx
<Svg src="img/eye.svg" className="eye-icon blue" dust="fairy" />
```

This way we can
 - always reference the file by path name
 - pass classes
 - and use some property (in this case `dust`) to signify whether it is necessary to use it inline for the CSS wizards

mgwalker:
I'm completely content with your responses.  The SVG thing makes sense - I feel like I've run into that before, but I could just be imagining it.  And refactoring, I get it.  lol.  It can cause all kinds of little quirks in the weirdest ways.

I think we should go ahead and merge this one so we can keep on moving.  👍

PR #23, opened 2017-02-23
Daily - 2/23/2017

mgwalker:


src/components/Section/Identification/ContactInformation/ContactInformation.test.jsx
```diff
@@ -58,11 +58,15 @@ describe('The ContactInformation component', () => {
...
     const component = mount(<ContactInformation {...expected} />)
-    expect(component.find('.table-cell strong').length).toEqual(5)
-    expect(component.find('.table-cell strong').at(0).text()).toEqual('(202) 867-5309 x1234')
-    expect(component.find('.table-cell strong').at(1).text()).toEqual('(202) 867-5309')
-    expect(component.find('.table-cell strong').at(2).text()).toEqual('867-5309')
-    expect(component.find('.table-cell strong').at(3).text()).toEqual('+001 1234567890 x1234')
-    expect(component.find('.table-cell strong').at(4).text()).toEqual('+001 1234567890')
+    // NOTE: This was commented out due to prototyping for next usability testing.
+    // Please removed or uncomment based on results.
+    // expect(component.find('.table-cell strong').length).toEqual(5)
```
mgwalker: Just curious, is there list of places where code was changes specifically for user testing?  If not, keeping one somewhere might make it easy to roll the changes back if the user testing suggests that's the right move.  Just a thought.

bmallred: Not really. This usability testing is our first deviation where a lot prototyping from the design comps is being done.

What kind of way do you mean by easier rolling back? We've talked internally (if this happens frequently) to do one of these:

- separate branches deployed
- system flags to allow for live quick switching

mgwalker: I was just thinking if the decision from user testing was to roll these things back, knowing where in the code the changes were would be useful. But a text search for "usability" will probably get you there as fast or faster than a list. ¯\\\_(ツ)_/¯

PR #27, opened 2017-03-03
Release #1 3/3/2017

adelevie:


src/components/Form/DateControl/DateControl.jsx
```diff
@@ -112,12 +113,12 @@ export default class DateControl extends ValidationElement {
         event.target.date = d
         super.handleChange(event)
 
-        // Always make sure the day is re-validated
-        if (['month', 'year', 'estimated'].includes(event.target.name)) {
-          this.refs.day.refs.input.focus()
-          this.refs.day.refs.input.blur()
-          event.target.focus()
-        }
+        // // Always make sure the day is re-validated
```
adelevie: might want to delete these if they're commented out.

src/components/Form/DateControl/DateControl.jsx
```diff
@@ -260,32 +261,39 @@ export default class DateControl extends ValidationElement {
   }
 
   render () {
+    console.log('month:', this.state.month)
```
adelevie: does this need to be logged in production, or is this just for debugging?

src/components/Form/Dropdown/Dropdown.jsx
```diff
@@ -104,15 +93,18 @@ export default class Dropdown extends ValidationElement {
...
-        valid = true
-        value = x.name
-      }
-    })
 
-    if (valid === false) {
-      errorCodes = 'notfound'
+    if (valid !== null) {
+      this.state.options.forEach(x => {
+        if (x.name.toLowerCase() === value.toLowerCase() || x.value.toLowerCase() === value.toLowerCase()) {
```
adelevie: might want to name the variables and/or sweep into a named function

adelevie:
LGTM

src/components/Form/DateControl/DateControl.jsx
bmallred: Yeah removed this in a recent commit on the `military-history` branch. Definitely was not meant to stick around.

src/components/Form/Dropdown/Dropdown.jsx
bmallred: Which variable do you mean? The `x` and/or `value`?

adelevie: Yeah, I can't really figure out the meaning behind this check. What's it really checking for?

bmallred: Cool... so like each `<option />` element has a `name` or `value` and it is going through comparing against the current `value` supplied by the input. If there is a match it places the value with the appropriate casing and such.

```
enter `az` get `Arizona`
enter `ariZoNa` get `Arizona`
```

PR #33, opened 2017-03-08
Release #1 3/8/2017

adelevie:


src/components/Section/History/Employment/ReasonOptions.jsx
```diff
@@ -0,0 +1,141 @@
...
+  updateText (values) {
+    this.onUpdate('Text', values)
+  }
+
+  updateDate (values) {
+    this.onUpdate('Date', values)
+  }
+
+  labelText () {
+    switch (this.state.Reason) {
```
adelevie: Why not use an object to match keys to values?

bmallred: hmm.... do you mean something like:

```js
return i18n.t(`history.employment.default.left.${this.state.Reason.toLowerCase()}.text`)
```

or

```js
const translations = {
  'Fired': i18n.t('history.employment.default.left.fired.text'),
  // etc....
}
return translations[this.state.Reason]
```

PR #36, opened 2017-03-14
Release #1 3/14/17

adelevie:


src/components/Form/Address/Address.test.jsx
```diff
@@ -99,4 +101,216 @@ describe('The Address component', () => {
...
+        }
+      }
+    ]
+
+    tests.forEach(test => {
+      expect(handleGeocodeResponse(test.response)).toEqual(test.expected)
+    })
+  })
+
+  it('can throttle', (cb) => {
```
adelevie: how does this work?

src/components/Form/Address/AddressSuggestion.test.jsx
```diff
@@ -0,0 +1,60 @@
...
+          zipcode: '22201'
+        },
+        expected: 4
+      }
+    ]
+
+    tests.forEach(test => {
+      const component = mount(<AddressSuggestion suggestion={test.suggestion} current={test.current} />)
+      expect(component.find('.highlight').length).toBe(test.expected)
+    })
```
adelevie: Should we also test for equality of values? Or do you think it's sufficient enough to check for just suggestion counts?

bmallred: @floresj

PR #39, opened 2017-03-20
Release #1 3/20/2017

adelevie:


src/components/Form/DateRange/DateRange.jsx
```diff
@@ -107,13 +93,9 @@ export default class DateRange extends ValidationElement {
...
    */
   handleValidation (event, status, error) {
-    if (!event) {
-      return
-    }
-
-    if (status !== false) {
-      if (this.state.from && this.state.to) {
-        if (this.state.from > this.state.to) {
+    if (event && status !== false && this.state.from && this.state.to) {
```
adelevie: might want to sweep this set of nested conditionals into a function

src/components/Form/Number/Number.jsx
```diff
@@ -94,23 +91,23 @@ export default class Number extends ValidationElement {
...
+        status = parseInt(this.state.value) <= parseInt(this.state.max)
         if (status === false) {
           errorCode = 'max'
         }
         hits++
       }
 
       if (status && this.props.maxlength && this.props.maxlength > 0) {
-        status = status && ('' + this.state.value).length <= parseInt(this.props.maxlength)
+        status = ('' + this.state.value).length <= parseInt(this.props.maxlength)
```
adelevie: what does prepending a string with an empty string do in javascript? do I even want to know?

bmallred: Oh it is true madness. It forces the value to be a string.

In this case `value` may be either a `string` or `number` and we are trying to impose a maximum length. So we force it to be cast as a string then take the length of that for our logic.

If we knew it would be a consistent type we may approach it differently but this solves both.

PR #45, opened 2017-04-06
Release #1 4/6/2017

adelevie:


src/views/Form/Form.scss
```diff
@@ -22,6 +22,10 @@
 
     li {
       margin: 0 4rem;
+
+      p {
+        padding: 0;
+      }
     }
 
     @media screen and (max-width: 1380px) {
```
adelevie: syntax issue?

bmallred: It is definitely [valid syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Syntax). Will have to research again why the Github syntax highlighter doesn't like some of these styles.

PR #52, opened 2017-04-18
Release #2 4/18/2017

adelevie:


src/components/Form/Field/Field.scss
```diff
@@ -72,19 +76,21 @@
 
         > div {
           width: 100%;
+          max-width: 64rem;
         }
 
-        &.shrink {
-          width: initial;
+        // &.shrink {
```
adelevie: delete these comments when you can

bmallred: Will do! Waiting to see if they are going to want them back in there since its in flux.

PR #88, opened 2017-08-14
Release 08/14/2017

adelevie:


api/main.go
```diff
@@ -25,52 +24,31 @@ func main() {
...
 	o.HandleFunc("/{service}/callback", handlers.AuthCallbackHandler)
 
-	// Validation
-	v := r.PathPrefix("/validate").Subrouter()
-	v.HandleFunc("/ssn/{ssn}", handlers.ValidateSSN)
-	v.HandleFunc("/email", handlers.ValidateEmail)
+	// Account specific actions
+	a := r.PathPrefix("/me").Subrouter().Inject(handlers.JwtTokenValidatorHandler)
+	a.HandleFunc("/validate", handlers.Validate).Methods("POST")
+	a.HandleFunc("/save", handlers.Save).Methods("POST")
```
adelevie: should this be PUT? does it matter?

api/model/form/location.go
```diff
@@ -0,0 +1,524 @@
...
+	case LayoutBirthPlaceWithoutCounty:
+		if domestic {
+			stack = validateFields(entity, "city", "state")
+		} else {
+			stack = validateFields(entity, "city", "county")
+		}
+	case LayoutCountry:
+		stack = validateFields(entity, "country")
+	case LayoutUSCityStateInternationalCity:
+		if domestic {
```
adelevie: this if/else pattern and the `validateFields` lines get repeated a lot. maybe a good candidate to refactor.

api/model/form/location.go
```diff
@@ -0,0 +1,524 @@
...
+
+	return !stack.HasErrors(), stack
+}
+
+func validateFields(entity *Location, props ...string) model.ErrorStack {
+	var stack model.ErrorStack
+
+	for _, prop := range props {
+		switch strings.ToLower(prop) {
+		case "street":
```
adelevie: these might be refactorable.

api/model/form/military.go
```diff
@@ -0,0 +1,226 @@
...
+	Explanation        Payload
+}
+
+// Unmarshal bytes in to the entity properties.
+func (entity *MilitarySelective) Unmarshal(raw []byte) error {
+	return json.Unmarshal(raw, entity)
+}
+
+// Valid checks the value(s) against an battery of tests.
+func (entity *MilitarySelective) Valid() (bool, error) {
```
adelevie: the logic here gets a little gnarly ... another refactor candidate?

api/main.go
bmallred: Good question. We should actually listen on both (for at least `/save`). This should be compliant  to standards of HTTP 1.1 and 2.0.

Funny thing about 2.0 is they are talking about doing away with PUT and DELETE verbs.

api/model/form/military.go
bmallred: Could do some early exiting in place of the `if-else` branches.

api/model/form/location.go
bmallred: We'll take a look at it. Pretty much copied the logic/patterns applied on the JavaScript side.

This one may be ugly just by the nature of how many different combinations there are, too.

src/schema/form/textarea.json
```diff
@@ -0,0 +1,7 @@
+export default const textarea = {
```
adelevie: should the file extension be JS and not JSON?

bmallred: Absolutely! This one is being ingested yet but that is a great find

PR #89, opened 2017-08-22
Release 08/22/2017

adelevie:


src/components/Section/Citizenship/Multiple/CitizenshipItem.jsx
```diff
@@ -72,35 +72,46 @@ export default class CitizenshipItem extends ValidationElement {
   }
 
   render () {
+    const d = this.props.Dates || {}
```
adelevie: `d` could probably be renamed to something clearer.

PR #90, opened 2017-08-29
Release 08/29/2017

adelevie:


api/handlers/root_test.go
```diff
@@ -20,12 +20,13 @@ func TestRoot(t *testing.T) {
 		t.Errorf("handler returned wrong status code: got %v want %v", status, http.StatusOK)
 	}
 
-	expected := `{}`
-	if rec.Body.String() != expected {
-		t.Errorf("handler returned unexpected body: got %v want %v", rec.Body.String(), expected)
-	}
+	// TODO: Fix!!!
```
adelevie: this should probably be removed

api/main.go
```diff
@@ -44,11 +44,14 @@ func main() {
 	// Account specific actions
 	a := r.PathPrefix("/me").Subrouter().Inject(handlers.JwtTokenValidatorHandler)
 	a.HandleFunc("/validate", handlers.Validate).Methods("POST")
-	a.HandleFunc("/save", handlers.Save).Methods("POST")
+	a.HandleFunc("/save", handlers.Save).Methods("POST", "PUT")
```
adelevie: does this mean both HTTP methods do the same thing?

api/model/account.go
```diff
@@ -15,7 +15,7 @@ var (
 	JwtSecret         = []byte("more secrets!")
 	Issuer            = "eqip"
 	BasicAuthAudience = "Basic"
-	Expiration        = time.Hour * 24
+	Expiration        = time.Hour * 1
```
adelevie: What does it mean that the token expires ever hour instead of every day?

api/handlers/root_test.go
bmallred: I think this has been removed in most recent  updates.

api/main.go
bmallred: Correct. We have one endpoint which handles creation and updating of the data.

api/model/account.go
bmallred: We had set it to 24 hours when there were some issues and we needed it to last for the duration of usability testing. In reality this will most likely be changed to something less than one hour in the future depending on production requirements.

PR #100, opened 2017-10-03
Release 10-3-2017

ctro:
I figured I'd keep these coming since I know there is a _big_ back-end PR in the works :)


src/components/Section/History/History.jsx
```diff
@@ -38,6 +38,28 @@ export const sort = (a, b) => {
...
 }
 
+export const totalYears = (birthdate) => {
+  let total = 10
+  if (!birthdate) {
+    return total
+  }
+
+  const eighteen = daysAgo(birthdate, 365 * -18)
+  total = Math.ceil(daysBetween(eighteen, today) / 365)
```
mgwalker: This reminds me of something I noticed when I was doing code reviews for Alan while he was out a few months ago.  There's some code in `src/components/Section/History/dateranges.js` that's doing some date math.  Date math always makes me nervous because it's so quirky.  The tests in `dateranges.test.js` aren't comprehensive and the mathy parts don't seem to be tested at all.  There are assumptions in here about the number of days in a year, which might be fine depending on how it's being used, though I'm not sure about that (that's really a question for y'all, since you know how the dates are being used)

For example, here, if the supplied `birthdate` is precisely 18 years ago, `daysAgo` will return a date that is at least 4 days short of 18 years (maybe up to 5? brain is being fuzzy about this right now...).  Then `daysBetween` will be a few days short of actual.  Okay, and then you do a `ceil` which takes care of that.  Okay, so here is fine.

So I guess three notes on this topic:

1. some comments here explaining the math that's being done would be helpful so future devs don't have to spend a lot of time reasoning through the code to understand it like I did up there ☝️ 
2. maybe rename these variables to `eighteenthBirthdate` (so it's clear what it represents and that it's a date) and `yearsSince18th` - `total` can be a separate variable
3. consider whether the date math in `dateranges.js` chould be replaced with a library; something like [timediff](https://www.npmjs.com/package/timediff) (the big downside to `timediff` is that it depends on [moment](https://npm.im/moment), and that adds a couple of MB of dependencies)
4. ...or if keeping `dateranges.js` makes sense, add more comprehensive tests, especially for the mathy functions (e.g., `daysBetween`, `daysAgo`, `julian`, etc.)

src/components/Section/Print/Print.test.jsx
```diff
@@ -0,0 +1,24 @@
...
+  // Setup
+  window.token = 'fake-token'
+  const middlewares = [ thunk ]
+  const mockStore = configureMockStore(middlewares)
+
+  it('visible when authenticated', () => {
+    const store = mockStore({ authentication: { authenticated: true, twofactor: true, application: applicationState } })
+    const component = mount(<Provider store={store}><Print /></Provider>)
+    expect(component.find('.section-print-container').length).toBe(10)
+  })
```
mgwalker: Is it possible to mock `window.print` here, simulate a print button click, and verify that `window.print()` was called?  I'm not sure if that's something you can do with enzyme but if so, that'd be a nice addition.

bmallred: We should be able to and great idea! Not sure if it'll be directly using enzyme but we've done similar things in the past which we can leverage.

src/components/Section/History/History.jsx
bmallred: Yeah I remember the conversation last time discussing bringing in `moment` which I thought we had a task/issue in the backlog for but I couldn't find it. We decided not to implement that change at the moment mainly due to size (but not opposed to doing so!).

We can definitely comment and make changes to the names to help with readability! It was also discussed to move all of the date functions in to one file (at one moment they were pretty dispersed but I think it is better now) so when we do decide to transition to a library it isn't that bad.

I'll come back with a link to this via an issue to track it and possible slate it for next iteration? Maybe a nice "welcome back" for @ctro 😈

bmallred: Found it! https://github.com/truetandem/e-QIP-prototype/issues/526

Will update w/ this conversation

mgwalker: Nice!  Yep, I like the idea of saving all of that for @ctro.  😉

PR #112, opened 2018-04-10
Release 03/07/2018

ryanhofdotgov:


ryanhofdotgov:
I proceeded with the merge despite the failing CircleCI build; it should be resolved with the PR for release-04-04-2018 that will be completed later this morning. The go-pg library wasn't set to a specific build and an update to the library broke the build. 

The code has already been updated to work with the new version of the library in the later Sprint 35 (04-04-2018) tag. We need to re-visit how we track dependencies as eApp heads towards a stable, deployed release.

PR #385, opened 2018-06-06
Fix modal scroll issue

line47:
There is a lot of conflicting CSS for the modal, so it took me a bit to track down.
I cleaned up some of the unneeded styles and as a bonus also set the background when the modal is active to something darker which is more typically seen.  There is still a lot of CSS cleanup work to do 

### Modal with darker background 
screen shot 2018-06-06 at 3 51 54 pm

### Modal scroll in IE11
modal-scroll

### DIFF
--- a/src/components/Form/Consent/Consent.scss
+++ b/src/components/Form/Consent/Consent.scss
@@ -1,8 +1,6 @@
 @import 'eqip-colors';
 
 .consent-modal .modal {
-  background-color: rgba(255, 255, 255, 0.95) !important;
-  z-index: 9999;
 
   .modal-wrap {
     // Originally this is a `table-cell` but the consent form is not embed within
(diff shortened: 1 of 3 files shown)


afeld:
Nice work! I like when fixes involve more deleting than adding.

bpdesigns:
Hi @line47 I like the modal with darker background for the questionnaire but it brings up a problem elsewhere. 

It changes the background color on the loading page which is not ideal. If possible to keep the dark bg from the questionnaire but change the loading back to white that would look better.

screen shot 2018-06-11 at 10 57 15 am

line47:
Hey @bpdesigns can you add this as a new issue to track?

bpdesigns:
@line47 see https://github.com/18F/e-QIP-prototype/issues/400

PR #387, opened 2018-06-07
[WORK IN PROGRESS] improve frontend development workflow

afeld:
Builds on #386.

## TODOs

* [ ] Clear out "watch" code that's no longer used


afeld:
Closing in favor of a more incremental approach in #393. I think switching out the various build patterns in this repository to leverage create-react-app isn't a bad idea, but ended up being a slightly bigger project than I expected. Might revisit later.

PR #389, opened 2018-06-07
use hosted version of DAP

afeld:
Just another thing that simplifies the setup.


ryanhofdotgov:
Sorry – to be clear, this is a good change 👍– intent was just to document on the ticket why they might have done it originally a few months back and why it makes sense to revert.

PR #393, opened 2018-06-09
improve JavaScript build speed/workflow

afeld:
This is an alternative to #387, taking a more incremental approach of tweaking the existing tooling rather than replacing it. What's changed:

* Watching for asset builds has been incorporated into `make run` - separate top-level `build`/`watch` process no longer needed
* Using Webpack (rather than Gulp) for watching JS changes.
    * This means faster build times, because it's smart enough to only rebuild what's necessary, rather than the entire frontend
    * Split JS and CSS building into their own containers, since each now has its own process
* Clarified/expanded development documentation

**Time for frontend changes to be reflected are down from 60 seconds to ~14,** and there this could be further improved without too much effort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,18 +51,20 @@ GitHub commits can be traced back to their corresponding tasks through commit co
 
 To view the items completed during each development sprint and to view the burndown charts for each respective sprint, please visit the [Sprint Backlogs][26] page.
 
-## Getting to know the code
+## Development
 
-### Dependencies
+### Initial setup
+
+#### Dependencies
 
  - [git](https://git-scm.com)
  - [docker][21]
  - [docker-compose][20]
  - [make](https://www.gnu.org/software/make/)
 
-For more information on licenses and third-party source code please refer to the [dependencies](DEPENDENCIES.md) documentation.
+For more information on licenses and third-party source code please refer to the [dependencies](docs/DEPENDENCIES.md) documentation.
 
-### Clone all things
+#### Clone all things
 
 Clone the repository and `cd` into it:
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -115,27 +124,6 @@ make test
...
-Then direct your browser at [http://localhost:8080](http://localhost:8080). The access the site in development use the username `test01` and password `password01`.
-
-## Docker containers
-
-| Container | Image               |
-| --------  | ------------------- |
-| api       | [Dockerfile.api](Dockerfile.api) |
-| db        | postgres:9.6.5      |
-| web       | nginx:alpine        |
-| frontend  | node:8.5.0          |
```
afeld: Not much use in restating what's in the Dockerfiles.

afeld:
Ok, ready for review!

line47:
It looks like the JS building is working but the CSS is not. 
I updated the fields.scss file and did not see changes reflected after it live rebuilt also after the css rebuilt the images in the header for the USA flag and the eApp logo were throwing 404s

screen shot 2018-06-11 at 5 14 15 pm

line47:
Also noticed a console error 
```
Refused to apply style from 'http://localhost:8080/css/eqip.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.```

related fix?
https://stackoverflow.com/questions/48248832/stylesheet-not-loaded-because-of-mime-type

**Comment by afeld** [comment]
How long did you wait? I'm now seeing the CSS build take a long time, which I didn't notice before.

```
css_1   | [16:32:29] Finished 'sass' after 50 s
```

Can deal with that separately - I wouldn't expect that build time to change with this pull request.

PR #398, opened 2018-06-11
remove unnecessary shell scripts

afeld:
None of these are used/needed.

### DIFF
--- a/Dockerfile.frontend
+++ b/Dockerfile.frontend
@@ -9,4 +9,4 @@ RUN yarn install
 
 COPY . .
 
-CMD ./bin/watch
+CMD yarn watch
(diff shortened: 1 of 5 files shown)


ryanhofdotgov:
Do you have an idea why the file permissions were being reset initially? It was something added fairly recently in https://github.com/18F/e-QIP-prototype/commit/4710c15480f3f67bec4d2ab2d108fda223dd612e#diff-bee0b57440e2e3ad9c189377b68e2528 – the other changes files on the commit, the timing of the change, as well as subsequent commits suggest to me that it might have to do with the AWS GovCloud containers we generate and push up. If I were to guess, Bryan was working with Eagle/Kx team at NBIS to get that environment working, and this was a solution to a problem they saw up there.

afeld:
Yeah, there can be some weirdness with permissions of files coming into/out of containers, so left `bin/permissions` in for now. Doesn't seem to be needed for local development, and may investigate production later.

PR #414, opened 2018-06-13
fix date range in people test

### LINKED ISSUE #379: `Person validator › should validate people` test is failing
Fix the `Person validator › should validate people` test so it no longer fails. Looks like it has been failing unnoticed for a while due to #377

```
FAIL  src/validators/people.test.js (418 MB heap size)
  ● Person validator › should validate people

    expect(received).toBe(expected)
    
    Expected value to be (using ===):
      true
    Received:
      false
      
      at src/validators/people.test.js:346:83
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (src/validators/people.test.js:345:11)
          at <anonymous>
```

afeld:
Fixes #379.

The validator checks to make sure that the user knows someone through the current year, but the test was hard-coded to end in 2017. In the last commit, changed it to go through the current year. In other words, this test has presumably been failing since 1/1/2018 😛

Also included some refactoring - see commits.


afeld:
Build fixed!

src/validators/people.test.js
```diff
@@ -174,8 +54,8 @@ describe('Person validator', function () {
                     to: {
                       month: '1',
                       day: '1',
-                      year: '2017',
-                      date: new Date('1/1/2017')
+                      year: currentYear,
+                      date: new Date(`1/1/${currentYear}`)
```
line47: 👍

PR #417, opened 2018-06-13
improve Jest speed

afeld:
See commits. Takes the tests from [10+ minutes](https://circleci.com/gh/18F/e-QIP-prototype/172) to [seven](https://circleci.com/gh/18F/e-QIP-prototype/174).


afeld:
Hmm, this test failure is legitimate:

```
 FAIL  src/services/api.test.js (588 MB heap size)
  ● The API › can parse query parameters from url

    TypeError: Cannot redefine property: search
        at Function.defineProperty (<anonymous>)

      76 |   it('can parse query parameters from url', () => {
      77 |     const previousLocation = window.location
    > 78 |     Object.defineProperty(window.location, 'search', {
         |            ^
      79 |       writable: true,
      80 |       value: '?foo=bar&test=1'
      81 |     })

      at Object.<anonymous> (src/services/api.test.js:78:12)
```

Will need to look into the best way to fix.

PR #418, opened 2018-06-13
improve JS build speed

afeld:
This pull request does a number of things to improve the JavaScript build speed:

* Upgrades Webpack and Babel
* Removes plugins that were unused or are now included by default
* Caches the intermediate compiled JS files for faster rebuilds
* Switches to a different kind of source map
* Reduces Webpack's module search space

As a result, **this takes initial build time from 116s to 44s, and subsequent build time from 116s to 19s** 🚀 Bonus: the Webpack config is now cleaner, too ✨


package.json
```diff
@@ -17,18 +17,14 @@
...
-    "babel-plugin-react-html-attrs": "^2.0.0",
-    "babel-plugin-transform-async-to-generator": "^6.16.0",
-    "babel-plugin-transform-class-properties": "^6.18.0",
-    "babel-plugin-transform-decorators-legacy": "^1.3.4",
+    "babel-loader": "^7.1.4",
     "babel-polyfill": "^6.26.0",
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-react": "^6.16.0",
-    "babel-preset-stage-0": "^6.16.0",
+    "babel-preset-stage-2": "^6.24.1",
```
afeld: Not sure why it was loading `stage-0` before, since the [`.babelrc`](https://github.com/18F/e-QIP-prototype/blob/master/.babelrc) has `stage-2` specified.

line47:
I'm getting an error when running `make run`

```
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
js_1    |  - configuration has an unknown property 'mode'. These properties are valid:
js_1    |    object { amd?, bail?, cache?, context?, dependencies?, devServer?, devtool?, entry, externals?, loader?, module?, name?, node?, output?, parallelism?, performance?, plugins?, profile?, recordsInputPath?, recordsOutputPath?, recordsPath?, resolve?, resolveLoader?, stats?, target?, watch?, watchOptions? }
js_1    |    For typos: please correct them.
js_1    |    For loader options: webpack 2 no longer allows custom properties in configuration.
js_1    |      Loaders should be updated to allow passing options via loader options in module.rules.
js_1    |      Until loaders are updated one can use the LoaderOptionsPlugin to pass these options to the loader:
js_1    |      plugins: [
js_1    |        new webpack.LoaderOptionsPlugin({
js_1    |          // test: /\.xxx$/, // may apply this only for some modules
js_1    |          options: {
js_1    |            mode: ...
js_1    |          }
js_1    |        })
js_1    |      ]
js_1    | error Command failed with exit code 1.
js_1    | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

afeld:
Try a `make setup-containers` first.

afeld:
Hmm, I was getting the same issue - seems to be a weird Docker caching thing. Try

```sh
docker-compose down && make run
```

PR #423, opened 2018-06-13
improve build speed of frontend Docker container

afeld:
Don't upload the `node_modules` to the Docker daemon, and don't bother copying the repository files into the image. Also added some documentation.

### DIFF
--- a/Dockerfile.frontend
+++ b/Dockerfile.frontend
@@ -7,6 +7,4 @@ COPY package.json package.json
 COPY yarn.lock yarn.lock
 RUN yarn install
 
-COPY . .
-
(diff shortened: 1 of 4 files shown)


Dockerfile.frontend
```diff
@@ -7,6 +7,4 @@ COPY package.json package.json
 COPY yarn.lock yarn.lock
 RUN yarn install
 
-COPY . .
```
afeld: Tried this as an attempt to fix https://github.com/18F/e-QIP-prototype/pull/418#issuecomment-397022500 - turned out to be useful regardless.

PR #425, opened 2018-06-13
move Dockerfiles and Cloud Foundry manifests out of top-level directory

afeld:
Just a tiny cleanup.


ryanhofdotgov:
Looks like this change ended up breaking the deploy – see: https://circleci.com/gh/18F/e-QIP-prototype/230

```
error: Error processing app files: lstat /app/deploy/api: no such file or directory
```
CircleCI wasn't allowing me to SSH in after re-running the job (connection timed out), so wasn't able to diagnose there or otherwise determine the cause – so I've reverted the merge.

It might be related to the cf cli tools being many versions behind – 6.26 vs 6.37 – over a year out of date. Build process uses this unmaintained Docker image: [adelevie/cf-cli](https://hub.docker.com/r/adelevie/cf-cli/)

afeld:
Ah, whoops, I didn't fix the paths in the manifest files. Ah well.

PR #438, opened 2018-06-19
linting improvements

afeld:
See commits.


afeld:
The linting is run in CircleCI, so sufficient to check there that it acts appropriately.

PR #443, opened 2018-06-19
move Dockerfiles and Cloud Foundry manifests out of top-level directory

afeld:
#425, take two. Since this change isn't actually tested by CI, this _might_ break the deploy. If that's the case, I'll submit another pull request, rather than reverting this.


ryanhofdotgov:
Success: https://circleci.com/workflow-run/eaca4d2d-4932-44ca-9beb-88a71f3cae4d

PR #458, opened 2018-06-20
support SAML for local development

### LINKED ISSUE #424: proposal: make authentication more consistent
In the current setup, the application acts differently when using an external SAML provider, versus faking one in development mode. What if we used an "external" SAML provider locally, such as [this Dockerized one](https://hub.docker.com/r/kristophjunge/test-saml-idp/)? Some benefits:

* We get to tear out a bunch of otherwise dead code
* Local development acts more like production, in that it's using an external SAML provider
* ...which also makes local debugging of SAML stuff easier

@ryanhofdotgov Thoughts?

afeld:
Closes #424.


docker-compose.identity.yml
```diff
@@ -0,0 +1,24 @@
+# extends docker-compose.yml
+# https://docs.docker.com/compose/extends/#multiple-compose-files
+
+version: '2'
+
+services:
+  identity:
+    image: wso2is:5.5.0
```
afeld: Note that this version doesn't match what's in use elsewhere, but seems to be ok...?

afeld:
OK, this is ready for review!

@line47 As someone with less context, would you be willing to try this out to make sure the instructions and everything are clear? Note that your `.env` file will need to be updated from the `.env.example`.

docker-compose.identity.yml
afeld: Noted in the docs.

docs/saml.md
```diff
@@ -0,0 +1,63 @@
...
+    1. Save that in `api/wso2.crt`.
+
+        ```
+        -----BEGIN CERTIFICATE-----
+        <contents>
+        -----END CERTIFICATE-----
+        ```
+
+1. Enable SAML on the "client" side.
+    1. In your `.env`, set `BASIC_ENABLED=` and `SAML_ENABLED=1`.
```
ryanhofdotgov: Maybe some clarification that all the `SAML_` environment variables need to be set, just in case they didn't start off with `.env.example`?

PR #463, opened 2018-06-21
don't proxy requests to the API

afeld:
Broken out from #458.

Previously, requests to the API were going through NGINX, but only in development. Now, NGINX is bypassed for those requests. Also, it now ensures that AJAX requests can be made to the API host by dynamically computing and injecting that host+port into the Content Security Policy.


afeld:
> It's doesn't feel super satisfactory to have to change files under /etc at runtime

Yeah, I was trying to do it at build time (in a Dockerfile), but ran into issues. For example, there isn't an easy way for the `.env` file to be read at build time.

> cloud.gov was going to eventually re-deploy containers that had files changed post-deploy?

Like it would watch GitHub or something? Not sure I follow. Also, seems to me that we aren't using the nginx config on cloud.gov, right?

ryanhofdotgov:
I can't find the reference, but I thought I read in the future cloud.gov was going to use file integrity monitoring to detect if files are changed after they were deployed (maliciously or otherwise) and automatically re-deploy the app from the original image it had.

And, I just realized that our cloud.gov instances are using a completely different setup for NGINX. Doh!

afeld:
Ah, yes, you're right - using [Sysdig Falco](https://sysdig.com/opensource/falco/). Hopefully that monitoring doesn't kick in until after startup, but we'll see.

PR #464, opened 2018-06-22
safer starting of the server

afeld:
See commits.


afeld:
More about the issue here: https://github.com/docker/compose/issues/4337 A bit unclear what the current recommended setup for this would be.

Issue #275, opened 2018-05-24
Logout link in upper right corner does not logout for SAML-based sign-ons


ryanhofdotgov:
Steps to reproduce:
- Sign-in via SAML (i.e., WSO2)
- Click Logout in upper right corner

Expected behavior:
Return to eApp login screen

Actual behavior:
Application data gets re-loaded, user stays logged in

Replicated  in `nbis-test.com` environment.

afeld:
Is there a good way to test this locally?

ryanhofdotgov:
I can replicate the issue with local WSO2 and eApp instances. I'll forward the instructions for set up that were passed to me, in case you'd like to work on this issue and get WSO2 incorporated into our project infrastructure.

afeld:
Wasn't able to reproduce using a local WSO2 5.5.0 (see #458), but will take a look at the logic in there and see if there's anything that looks flaky.

PR #465, opened 2018-06-22
display logout errors

afeld:
...in a query parameter and in the browser console. Not sure the former is needed, really, but I left it in. This isn't a fix for #275, but will help troubleshoot. Mind trying this branch locally to see what you get?

Also included a small refactor.

### DIFF
--- a/src/actions/AuthActions.js
+++ b/src/actions/AuthActions.js
@@ -33,14 +33,19 @@ export function login (username, password) {
 /**
  * Logs out a user
  */
-export function logout (error = '') {
+export function logout () {
   return function (dispatch, getState) {
-    const clear = () => {
+    const clear = (path) => {
       api.setToken('')
       dispatch({ type: AuthConstants.LOGOUT })
-      dispatch(push(`/login${error ? '?error=' : ''}${error || ''}`))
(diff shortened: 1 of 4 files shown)


ryanhofdotgov:
Using the `logout-fix` branch I get this error in the JavaScript console when I click Logout:
```
warning.js:33 Warning: Login is changing an uncontrolled input of type hidden to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
```
There aren't any errors in the network log nor the api log – `logout` is returning 200. Not related to anything on this branch, but I did notice that if I kept on clicking logout over and over, about the 3rd or 4th time the UI showed I was really logged out. 

I think to make headway on the related defect, you might have to do a deep-dive on how the JWT token is handled, when it is cleared, etc., how caching effects it (hopefully not at all), TTL on token renewal, verify that the UI faithfully shows logout status, etc.

afeld:
Going to pair to troubleshoot this.

afeld:
Will open a new pull request for this.

PR #469, opened 2018-06-25
[WORK IN PROGRESS] use React Router for navigation

afeld:

PR #470, opened 2018-06-25
[WORK IN PROGRESS] dramatically simplify use of AuthenticatedView

afeld:
Only needed on the top-level view components, which are the `Form` and `Help`. Also removed some unused `import`s.

### DIFF
--- a/src/components/Form/Introduction/Introduction.jsx
+++ b/src/components/Form/Introduction/Introduction.jsx
@@ -3,7 +3,6 @@ import { connect } from 'react-redux'
 import { i18n } from '../../../config'
 import { updateApplication } from '../../../actions/ApplicationActions'
 import { logout } from '../../../actions/AuthActions'
-import AuthenticatedView from '../../../views/AuthenticatedView'
 import Branch from '../Branch'
 import Modal from '../Modal'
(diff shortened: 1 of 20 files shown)


ryanhofdotgov:
Failing tests aside, just skimming the code, and I'm wondering if the AuthenticatedView does serve a real purpose on all the components. 

In general, if the client is no longer authenticated but clicks on a component element, you'd always want that to re-direct to login, right? Otherwise, for example, your UI actions might make a bunch of API calls that might fail silently to the user?

afeld:
Going to hold off on this until #469 is in, as I think we can also simplify the authentication logic by following [this blog post](https://tylermcginnis.com/react-router-protected-routes-authentication/).

afeld:
@ryanhofdotgov Since React is re-rendering the full tree on every update, the top-level "views" wrapped with `AuthenticationView` handle that case.

afeld:
Sorry, I never really explained the motivation here: Components wrapped in an `AuthenticationView` have added complexity and hidden magic added by that wrapper, and are more cumbersome to test and debug.

This pull request was way too many updates at once, so splitting into smaller ones with test updates to match.

PR #471, opened 2018-06-26
update ESLint

afeld:
* Upgraded ESLint
* Expand the ESLint config to use most recommended rules
* Fix ESLint not previously looking at JSX files (🤦‍♂️)

Doing this because I think it would have caught some subtle bugs.


ryanhofdotgov:
Have you seen this error before?
```
package golang.org/x/lint/golint: unrecognized import path "golang.org/x/lint/golint" (parse https://golang.org/x/lint/golint?go-get=1: no go-import meta tags ())
Note: checking out 'tags/v0.13.1'.
```
This is in the setup phase, which is causing `golint` to be missing in test phase.

afeld:
Restarted the build - seemed to be a fluke. Look ok otherwise?

Issue #475, opened 2018-06-27
remove custom routing logic


afeld:
The frontend code is working around React Router, doing its own history management rather than leveraging the library in an idiomatic way. The good news is that improving this should mean we get to delete a good bit of confusing code.

PR #476, opened 2018-06-27
use React Router for section navigation

afeld:
Part of #475. Builds on #473 and #474.

Previously, `click` events on the section navigation links were being captured, and then history was being updated manually. React Router [`<Link>`s](https://reacttraining.com/react-router/web/api/Link) already handle this, so there's no need to duplicate that functionality. See the deleted `onClick`s for examples.

Also moved all the middleware that depended on those route changes to the `Form` component, which is able to "listen" for route changes in a similar way, but is more scoped and [idiomatic](https://stackoverflow.com/a/44410281/358804).


src/views/Form/Form.jsx
```diff
@@ -12,23 +18,101 @@ import { env } from '../../config'
...
   }
 
-  componentDidUpdate () {
+  componentDidUpdate(prevProps) {
     this.defaultRedirect()
+
+    // https://stackoverflow.com/a/44410281/358804
+    if (this.props.location !== prevProps.location) {
+      this.onRouteChanged(prevProps.location);
+    }
```
afeld: This is where all the middleware moved to.

PR #478, opened 2018-06-27
[WAIT TO MERGE] use React Router instead of push() for navigation

afeld:
Part of #475. Builds on #476.

**Plain-ish language version:** Use React Router as the source of truth for navigation state.

**Nitty gritty:** The `dispatch(push())` calls are using Redux actions to trigger [`history`](https://reacttraining.com/react-router/web/api/history) changes, meaning that Redux is the source of truth for navigation. This is contrary to how React Router is meant to be used. Now, navigation is triggered in these cases by updating the `history` directly.

### DIFF
--- a/src/middleware/history.js
+++ b/src/middleware/history.js
@@ -1,5 +1,4 @@
 import axios from 'axios'
-import { env } from '../config'
 import { updateApplication } from '../actions/ApplicationActions'
 import { sectionData } from '../components/Section/sectionData'
 import schema from '../schema'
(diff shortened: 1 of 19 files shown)


src/middleware/history.js
```diff
@@ -18,32 +17,6 @@ export const findPosition = (el) => {
...
-}
-
-/**
- * historyMiddleware is a custom middleware function for redux that allows history actions to be
- * dispatched in order to change router paths.
- */
-export const historyMiddleware = store => next => action => {
-  // If we get a PUSH_STATE type, modify hisory
-  if (action.type === PUSH_STATE) {
-    env.History().push(action.to)
```
afeld: This is previously where the Redux action would turn into a `history` update.

Issue #486, opened 2018-06-29
improve styling workflow


afeld:
SASS changes can currently take 40+ seconds to rebuild. Let's get this faster.

afeld:
Trying out [Browsersync](https://browsersync.io/) and it's promising.

```shell
cd dist
browser-sync start --proxy web.eapp.local.test:8080 --files "css/*"
```

PR #487, opened 2018-06-29
improve CSS build time

afeld:
Part of #486.

Takes rebuild time from 20-40 seconds down to < 6, using node-sass-chokidar for watching files (more directly) rather than going through Gulp.


afeld:
Build fixed!

PR #495, opened 2018-07-02
add documentation for multi-host setup in development

afeld:
Steps used to support #489.

### DIFF
--- a/docs/multi-host.md
+++ b/docs/multi-host.md
@@ -1,0 +1,14 @@
+# Multiple hosts in local development
+
+To mimic production in having the various services served from different hostnames, do the following:
+
+1. Add the subdomains.
+
+    ```sh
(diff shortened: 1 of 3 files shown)


docs/multi-host.md
```diff
@@ -0,0 +1,14 @@
...
+
+To mimic production in having the various services served from different hostnames, do the following:
+
+1. Add the subdomains.
+
+    ```sh
+    sudo sh -c 'echo "127.0.0.1\tapi.eapp.local.test web.eapp.local.test" >> /etc/hosts'
+    ```
+
+1. If using [SAML](saml.md), modify the Service Provider to use a `Assertion Consumer URL` of `http://api.eapp.local.test:3000/auth/saml/callback`.
```
ryanhofdotgov: * `COOKIE_DOMAIN` is only relevant if you are doing SAML, so you could nest it there
* I don't think it is strictly required, but you probably want to update `API_REDIRECT`, `API_BASE_URL` otherwise things might get confusing.

PR #549, opened 2018-07-13
fail build if there are any golint errors

afeld:
Wasn't actually failing before - whoops! @ryanhofdotgov Would you be willing to fix the linting errors, since they're around areas of the code you've worked on?

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -652,14 +652,10 @@ func PurgeAccountStorage(context DatabaseService, account int) {
 	}
 }
 
-func FormatShaSum(hash [sha256.Size]byte) string {
-	return hex.EncodeToString(hash[:])
-}
-
 // Hash returns the SHA256 hash of the application state in hexadecimal
 func Hash(context DatabaseService, account int) string {
 	hash := sha256.Sum256(Application(context, account, true))
-	return FormatShaSum(hash)
+	return hex.EncodeToString(hash[:])
 }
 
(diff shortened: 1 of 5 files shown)


api/application.go
```diff
@@ -652,14 +652,10 @@ func PurgeAccountStorage(context DatabaseService, account int) {
 	}
 }
 
-func FormatShaSum(hash [sha256.Size]byte) string {
-	return hex.EncodeToString(hash[:])
-}
```
afeld: I like this function, actually. Not as obvious what's happening without it.

ryanhofdotgov: Well, I liked it too at some point. :) 

But I went to write the comment for it to resolve the lint error, and realized that the comment for that function was going to be almost exactly the same comment as the function 2 lines down from it, modulo 2 words.

And the only reason `FormatShaSum` existed at all was to call it during unit testing – usually that's not encouraged. And in the test code, the actual value of the string doesn't really matter. 

Finally, it doesn't make the `Hash()` _that_ much easier to understand (since it itself is only 2 lines, with a nice comment at the top), so I gave it the heave-ho and reduced the number of functions publicly visible in the package.

ryanhofdotgov: Oh – should mention that I only added Scott as reviewer since GitHub wouldn't let me add you (because you opened the PR), but if you're willing to buy off on my reasoning, feel free to merge in.

PR #561, opened 2018-07-16
Release 07/16/2018

ryanhofdotgov:


ryanhofdotgov:
Holding off merge until Snyk reports are reviewed/triaged https://github.com/18F/e-QIP-prototype/issues/563 – Snyk is currently only evaluating the deltas on a commit, and I think this is the first time we've seen a more comprehensive report.

afeld:
Worth noting that those findings aren't a regression, i.e. the last tagged release will likely have as many or more of the same issues, as the dependencies haven't been updated in a while.

ryanhofdotgov:
Proceeding with merge – nothing in Snyk report affects the app while deployed, just modules that aren't used, or used during linting/testing/etc.

Issue #585, opened 2018-07-18
ABC Issue


michaelccata:
**What did you do?**

You know what you did. 

**What did you expect to see?**

*Describe the desired behavior and what would deem this issue, bug, or feature
complete.*

**What did you see instead?**

*Include a high-level description of the feature or error here*

**Additional Information**

*Include any images, steps to recreate, notes, emojis, or anything else that
seems pertinent.*

 - What browser and version are you using?
 - What operating system?

afeld:
https://web.nitsl.mitre.org/jira/browse/EAPP-1001

PR #589, opened 2018-07-19
[WORK IN PROGRESS] switch from localStorage to sessionStorage

afeld:
The latter has additional security properties, like only being accessible to the one tab, and getting cleared when the tab is closed.

cc #585


ryanhofdotgov:
https://gsa-tts.slack.com/archives/C13HJ0ZBN/p1532025014000175

PR #593, opened 2018-07-23
simplify token handling

afeld:
See commits.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "gulp": "^4.0.0",
     "history": "^4.7.2",
     "jest": "^23.1.0",
+    "js-cookie": "^2.2.0",
     "jsdoc": "^3.5.5",
     "node-sass-chokidar": "^1.3.0",
     "prop-types": "^15.6.0",
--- a/src/services/api.js
+++ b/src/services/api.js
@@ -27,8 +27,10 @@ const getSplitValue = (key, raw, delim1, delim2) => {
 
 export const deleteCookie = (name) => {
   const domain = process.env.COOKIE_DOMAIN || window.location.hostname
-  // TODO complain if cookie not present
   Cookies.remove(name, { domain })
+  if (Cookies.get(name)) {
+    console.warn(`${name} cookie couldn't be removed - check that domain matches, etc.`)
+  }
 }
 
 class Api {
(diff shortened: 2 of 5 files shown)


src/services/api.js
```diff
@@ -1,37 +1,13 @@
...
-}
-
 export const deleteCookie = (name) => {
   const domain = process.env.COOKIE_DOMAIN || window.location.hostname
-  // TODO complain if cookie not present
-  document.cookie = `${name}=; domain=${domain}; path=/; expires=Thu, 01 Jan 1970 00:00:01 GMT;`
+  Cookies.remove(name, { domain })
+  if (Cookies.get(name)) {
+    console.warn(`${name} cookie couldn't be removed - check that domain matches, etc.`)
+  }
```
afeld: This is just a safeguard.

src/views/Login/Login.jsx
```diff
@@ -39,25 +40,13 @@ export class Login extends React.Component {
...
-    const err = getQueryValue(window.location.search, 'error')
-    if (err) {
-      switch (err) {
-      case 'token':
-        this.props.history.push('/token')
-        return
-      case 'access_denied':
-        this.props.history.push('/accessdenied')
-        return
-      }
```
afeld: Follow-up to 06d39ce8d28bb412c7c6aba9c38c86215251102e, since the query parameter doesn't seem to be getting set anywhere. We should make sure that the "access denied" flow with SAML is checked in manual QA, though.

package.json
```diff
@@ -34,6 +34,7 @@
     "gulp": "^4.0.0",
     "history": "^4.7.2",
     "jest": "^23.1.0",
+    "js-cookie": "^2.2.0",
```
afeld: This should handle any cross-browser compatibility weirdness, on top of allowing us to delete custom cookie-handling code.

src/views/Login/Login.jsx
ryanhofdotgov: Don't we still need some of the above logic to handle the "access denied" case? https://github.com/18F/e-QIP-prototype/blob/1f0f9a6d4af5aa424963415aad45b3f4c7e89cf0/api/http/saml.go#L133

afeld: Ahhh, you are right - missed that, sorry! Walked back those changes.

afeld:
Let's come back to it later - my bad for expanding the scope of this pull request to cover that error handling.

PR #599, opened 2018-07-24
Fix ws enabled doc

ryanhofdotgov:
Remembered to do this after merging in #598

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -324,7 +324,7 @@ Determines whether to enabled the submission to the eqip webservice
 
 **Target** - Back-end (api)<br>
 **Default** - True<br>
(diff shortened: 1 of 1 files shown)


docs/CONFIGURATION.md
```diff
@@ -324,7 +324,7 @@ Determines whether to enabled the submission to the eqip webservice
 
 **Target** - Back-end (api)<br>
 **Default** - True<br>
```
afeld: When we say "default" here...default if there's no `.env`?

ryanhofdotgov: That's correct – it described what the configuration value will be if the software doesn't detect it in the OS environment (for runtime variables) or if it's not in `.env` for the build time ones.

Issue #441, opened 2018-06-19
fix specs


afeld:
**What did you do?**

Ran `make specs`.

**What did you expect to see?**

Something...?

**What did you see instead?**

No output, and the process ended almost immediately, seemingly without running the specs.

**Additional Information**

Integration with CircleCI or NBIS self-hosted CI is out of scope for this ticket and will be a follow-on.

ryanhofdotgov:
I've not dived into this area on the project yet, but essentially:
* there is a very small number of Selenium tests, no other integration testing
* these get kicked off via the process described [here](https://github.com/ryanhofdotgov/e-QIP-prototype-truetandem/blob/master/SPECTEST.MD) on very slim chance you didn't see it already
* the previous team did not feel they were appropriate to run in conjunction with the other automated tests that run CircleCI (primarily due to length of time)
* right before the previous team went off contract, we tried to re-focus efforts on getting a good integration test suite in place, but had to [downscope that effort quite a bit](https://github.com/ryanhofdotgov/e-QIP-prototype-truetandem/issues/2877) due to other end-of-contract concerns.

afeld:
> the process described here

Heh, remembered that doc right after creating this issue. Thanks!

> the previous team did not feel they were appropriate to run in conjunction with the other automated tests that run CircleCI

What if we parallelized them, or ran nightly or something?

ryanhofdotgov:
Right - running nightly seemed appropriate when the team discussed it last.

afeld:
Repurposing this issue for getting the specs passing, not just running.

line47:
Have all the PRs in for getting the specs working as they were written.
There is still more spec work to be done as a few of the specs had commented out sections.

In addition, there is some redundant code that can be refactored for example navigating to sections.

ryanhofdotgov:
@line47 Just looking at it now... I get this error when I try to run `make specs`:
```
[e-QIP-prototype]# make specs
Running integration test suite
Starting eqipprototype_web_1 ... 
Starting eqipprototype_hub_1 ... done
Starting eqipprototype_chrome_1 ... done
Building nightwatch
ERROR: Forbidden path outside the build context: ../conf/docker/Dockerfile.frontend ()
make: *** [specs] Error 1
```
Does that work on your machine?

ryanhofdotgov:
[...]

PR #600, opened 2018-07-25
Fixing auth spec

line47:
Removing the pauses that are not needed and fixing the authentication spec.

screen shot 2018-07-25 at 9 42 13 am

This begins to chip away at #441

### DIFF
--- a/specs/features/step_definitions/authentication.js
+++ b/specs/features/step_definitions/authentication.js
@@ -86,6 +86,7 @@ defineSupportCode(({Given, Then, When}) => {
     return client
       .isVisible('a.logout', (result) => {
         client.click('a.logout')
+        .waitForElementVisible('.consent-acceptance', 5000)
       })
       .end()
   }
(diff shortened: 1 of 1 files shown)


ryanhofdotgov:
Is there a guiding principle that determines when a pause is required and when it's not? e.g., when it's making network call?

line47:
After inquiring with some dev/eng folsk it seems like `.pause()` isn't good to use and seems to be bad practice unless debugging the console. Seems like if it must be used then`.waitForVisibleElement()` would be a better choice. 

I'm wondering if this was around from previous work and was just never cleaned up?

ryanhofdotgov:
Great – just wanted to make sure there was some kind of approach to it as you remove pauses, extend them, etc. around network calls – also taking into account the CI system may have different performance characteristics. Thanks!

afeld:
Right - you need to wait any time the page is doing re-rendering, which often happen as a result of (after waiting for) network requests. `pause()`s are the brittle way to do it - agreed that waiting for elements to appear/disappear is better.

specs/features/step_definitions/authentication.js
```diff
@@ -86,7 +85,7 @@ defineSupportCode(({Given, Then, When}) => {
   const logout = () => {
     return client
       .isVisible('a.logout', (result) => {
-        client.click('a.logout').pause(1000)
+        client.click('a.logout')
```
afeld: We should `.waitForElementVisible('.consent-acceptance', 5000)` here so that subsequent steps don't fail.

line47: I updated in the code but it's not reflecting here 🤷‍♂️

PR #601, opened 2018-07-25
Update citizenship spec

line47:
Update nav item selections

- During testing the first time the test ran for citizenship everything passed. 
- The second time (because answered items were not cleared) some of the test items failed
- We need a way to flush the test data after every test. 

This continues to chip away at #441

### DIFF
--- a/specs/features/step_definitions/citizenship.js
+++ b/specs/features/step_definitions/citizenship.js
@@ -119,7 +119,7 @@ const completeCitizenshipForeignPassports = (promise) => {
 }
 
 const navigateToSection = (section) => {
-  const selector = '.usa-sidenav-list a[aria-controls="/form/' + 'citizenship' + '"]'
+  const selector = '.usa-sidenav-list a[aria-controls="/form/' + section + '"]'
   return client
     .assert.visible(selector)
     .click(selector)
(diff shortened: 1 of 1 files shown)


specs/features/step_definitions/citizenship.js
```diff
@@ -119,7 +119,7 @@ const completeCitizenshipForeignPassports = (promise) => {
 }
 
 const navigateToSection = (section) => {
-  const selector = '.section a[title="' + section + '"]'
+  const selector = '.usa-sidenav-list a[aria-controls="/form/' + 'citizenship' + '"]'
```
afeld: Can we change `'citizenship'` back to using the `section` variable?

specs/features/step_definitions/citizenship.js
```diff
@@ -119,7 +119,7 @@ const completeCitizenshipForeignPassports = (promise) => {
 }
 
 const navigateToSection = (section) => {
-  const selector = '.section a[title="' + section + '"]'
+  const selector = '.usa-sidenav-list a[aria-controls="/form/' + section + '"]'
```
line47: @afeld when I use `section` here is places "Citizenship" and not "citizenship" lowercase. I'm unsure why? Above section is set to lower case on line 28. Any ideas why? 

I'm getting this error ```ERROR: Unable to locate element: ".usa-sidenav-list a[aria-controls="/form/Citizenship"]" using: css selector```

afeld: Did you look where `navigateToSection` is called from?

PR #602, opened 2018-07-25
add import checks to eslint

afeld:
Was running into some import-related problems on another branch, and figure this may help.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,8 @@
 {
   "extends": [
     "eslint:recommended",
-    "plugin:react/recommended"
+    "plugin:react/recommended",
+    "plugin:import/errors"
   ],
   "env": {
     "browser": true,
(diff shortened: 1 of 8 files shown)


.eslintrc
```diff
@@ -20,6 +22,7 @@
     "react"
   ],
   "rules": {
+    "import/no-duplicates": "off",
```
afeld: Like the others below, we probably want to enable this at some point - just disabling to avoid changing a zillion files right now.

afeld: Noted in #607.

src/components/Section/Legal/Police/index.js
```diff
@@ -1,2 +0,0 @@
-import Police from './Police'
```
afeld: This component no longer exists. Ditto for `RelationshipStatus`.

PR #603, opened 2018-07-25
Update to use section for lowercase "citizenship"

line47:


specs/features/step_definitions/citizenship.js
```diff
@@ -27,7 +27,7 @@ defineSupportCode(({Given, Then, When}) => {
     subcontext = subsection
     const section = 'citizenship'
     const sectionTitle = 'Citizenship'
```
afeld: Let's remove this (now) unused variable.

PR #604, opened 2018-07-25
[WORK IN PROGRESS] check for circular dependencies in webpack

afeld:
Builds on #602.

In #605 I was running into problems with circular dependencies. Babel/Webpack seem to handle them _sometimes_, but not others, and I keep getting unlucky 😁 

* https://stackoverflow.com/questions/49116860/webpack-es6-modules-circular-dependencies-when-using-index-files
* https://stackoverflow.com/questions/38841469/how-to-fix-this-es6-module-circular-dependency

This pull request adds an automatic check for circular `import`s. Turns out we already have some!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,9 @@ jobs:
           name: Lint
           command: make lint
       - run:
+          name: Build
+          command: make build
+      - run:
           name: Test
           command: make test
       - run:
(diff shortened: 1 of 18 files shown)


afeld:
More broadly: the use of the `index.js` files to aggregate `exports` are a frequent culprit of these circular imports, so medium-term, I would suggest getting rid of them.

el-mapache:
@ryanhofdotgov I think we should close this PR, unless someone is planning on picking this work up this sprint

ryanhofdotgov:
Good idea to clean this up – since I think there is some value to revisit, I've opened up https://github.com/18F/e-QIP-prototype/issues/1145

PR #605, opened 2018-07-26
create the rest of the Identification subsections dynamically

afeld:
Following up on #595, this pull request takes the remaining instantiations of the subsection components for the Identification section and generates them dynamically from the navigation config. There are still improvements I want to make, tests I want to add, etc., but I don't want this pull request to get even bigger.

I recommend ignoring whitespace when viewing the diff.


afeld:
Ready for review!

afeld:
Part of #614.

PR #610, opened 2018-07-26
Get financial specs working 

line47:
🎉 

screen shot 2018-07-26 at 11 59 15 am


specs/features/step_definitions/financial.js
```diff
@@ -181,7 +180,7 @@ const navigateToSection = (section) => {
 }
 
 const navigateToSubsection = (section, subsection) => {
-  const selector = '.subsection a[href="/form/' + section + '/' + subsection + '"]'
+  const selector = '.usa-sidenav-sub_list a[href="/form/' + section + '/' + subsection + '"]'
   return client
     .assert.visible(selector)
     .click(selector)
```
afeld: Seems like these functions are almost identical between the different spec files... we should try extracting them as a shared file as a follow-up.

PR #620, opened 2018-07-30
Return error if template key missing

ryanhofdotgov:
Default template behavior is to insert `<no value>` into the template
when key is missing; instead stop processing and return error.

### DIFF
--- a/api/eqip/elements.go
+++ b/api/eqip/elements.go
@@ -20,9 +20,13 @@ type ImportRequest struct {
 	Content       Base64Content
 }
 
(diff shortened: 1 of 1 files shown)


api/eqip/elements.go
```diff
@@ -20,9 +20,13 @@ type ImportRequest struct {
 	Content       Base64Content
 }
 
+// When processing templates, indicates that execution should stop immediately.
+const missingKeyIsError = "missingkey=error"
```
afeld: That's a weird way for Go to take options...

PR #624, opened 2018-07-31
Improve eQIP API error handling

ryanhofdotgov:
Add higher-level error messages and differentiate between protocol and application-level errors.

### DIFF
--- a/api/eqip/client.go
+++ b/api/eqip/client.go
@@ -38,8 +38,7 @@ func (c *Client) Send(action RequestBody, response Body) error {
 	// Generates XML based of text/templates from the golang lib.
 	xmlContent, err := NewSOAPEnvelopeTemplate(action, unsignedToken, string(signedToken))
 	if err != nil {
-		fmt.Println("Error: ", err)
-		return err
+		return fmt.Errorf("Could not generate SOAP message: %s", err.Error())
 	}
 	buf := new(bytes.Buffer)
 	buf.ReadFrom(xmlContent)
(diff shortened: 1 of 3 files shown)


api/eqip/client.go
```diff
@@ -52,25 +51,31 @@ func (c *Client) Send(action RequestBody, response Body) error {
...
 	err = xml.Unmarshal(buf.Bytes(), &temp)
 	if err != nil {
-		fmt.Println("Error unmarshalling xml: ", err)
-		return err
+		return fmt.Errorf("SOAP request is malformed: %s", err.Error())
 	}
+
 	resp, err := http.Post(c.endpointURL, "text/xml; charset=utf-8", bytes.NewReader(temp.XML))
-	// Regardless of what happens, we always save the response body. In the event of network http error,
-	// we set the response body to the error message
```
afeld: Just confirming: this should no longer be the case?

api/eqip/client.go
```diff
@@ -52,25 +51,31 @@ func (c *Client) Send(action RequestBody, response Body) error {
...
-		return err
+		return fmt.Errorf("Cannot read SOAP response: %s", err.Error())
 	}
+
+	err = xml.Unmarshal(bodyBytes, response)
+	if err != nil {
+		return fmt.Errorf("SOAP response is malformed: %s: %s", err.Error(), string(bodyBytes))
+	}
+
 	response.SetResponseBody(bodyBytes)
```
afeld: This seems circular...isn't the line above taking the data out of the `bodyBites` and saving into the `response`?

api/eqip/client.go
ryanhofdotgov: Yeah, that wasn't a design choice I agreed with and makes the semantics of what gets persisted in the database muddy. Anything other than a SOAP response will get logged like other errors.

PR #628, opened 2018-08-01
Fix military spec

line47:
Part of #441 

screen shot 2018-08-01 at 10 12 08 am

### DIFF
--- a/specs/features/step_definitions/authentication.js
+++ b/specs/features/step_definitions/authentication.js
@@ -50,7 +50,7 @@ defineSupportCode(({Given, Then, When}) => {
 
   const goToLoginPage = () => {
     return client
-      .url(client.launch_url + '')
+      .url(client.launch_url)
       .waitForElementVisible('body', 1000)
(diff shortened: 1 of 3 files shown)


specs/features/step_definitions/authentication.js
```diff
@@ -50,7 +50,7 @@ defineSupportCode(({Given, Then, When}) => {
 
   const goToLoginPage = () => {
     return client
-      .url(client.launch_url + '/#/login')
+      .url(client.launch_url + '')
```
ryanhofdotgov: Can we just ditch the `+ ''`?

specs/features/step_definitions/authentication.js
```diff
@@ -60,7 +60,7 @@ defineSupportCode(({Given, Then, When}) => {
 
   const signIn = (username, password) => {
     return client
-      .assert.urlContains('/login')
+      .assert.urlContains('')
```
ryanhofdotgov: This assert is a no-op, right? – we should remove.

Issue #614, opened 2018-07-28
dynamically generate the sections


afeld:
https://trello.com/c/mdSZqL7p/98-deduplicate-section-specification

Rather than have the list of subsections per section listed multiple times in the code, let's have a single source of truth.

* [x] Identification
* [x] Financial
* [ ] Relationships
* [x] Citizenship
* [ ] Military
* [ ] History
* [ ] Foreign
* [ ] Legal
* [ ] Psychological
* [ ] SubstanceUse

PR #666, opened 2018-08-10
fix Identification snapshot test(s)

afeld:
...and generate them for every subsection. Previously, it was rendering empty, which isn't useful at _all_. Totally missed that before.

[Snapshot testing](https://jestjs.io/docs/en/snapshot-testing) is what I'm using to help prevent regressions while doing #614.

Recommend viewing diff with whitespace ignored.

### DIFF
--- a/src/components/Section/Identification/Identification.test.jsx
+++ b/src/components/Section/Identification/Identification.test.jsx
@@ -5,6 +5,7 @@ import thunk from 'redux-thunk'
 import { Provider } from 'react-redux'
 import Identification, { IdentificationSections } from './Identification'
 import { mount } from 'enzyme'
+import navigation from './navigation'
 
 const applicationState = {
   Identification: {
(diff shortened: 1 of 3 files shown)


src/components/Section/Identification/Identification.test.jsx
```diff
@@ -26,51 +27,80 @@ jest.mock('../../Form/ValidationElement/helpers', () =>
...
+    it(`renders the Identification component for the ${
+      subsection.url
+    } subsection`, () => {
+      const store = mockStore({
+        authentication: { authenticated: true },
+        application: applicationState
+      })
+      const component = renderer.create(
+        <Provider store={store}>
+          <Identification subsection={subsection.url} />
```
afeld: This was the key change.

ryanhofdotgov:
How'd you catch it out of curiosity? Can you visualize the snapshot files?

afeld:
I just open up the snapshot file and look for [things like this](https://github.com/18F/e-QIP-prototype/blob/8dc9cde62c07f72e75bdbee30c9931e983d4c89f/src/components/Section/Identification/__snapshots__/Identification.test.jsx.snap#L4-L6)...which is usually in response to a test not failing when it should. There is likely a smarter way to catch this.

afeld:
Identification is just the section I started with, yeah. Thanks!

Issue #396, opened 2018-06-10
Codecov report only evaluates Javascript code


ryanhofdotgov:
Backend api test coverage is not reflected in [Codecov reports](https://codecov.io/gh/18F/e-QIP-prototype/pull/395?src=pr&el=tree) attached to issues – just the frontend; e.g. https://github.com/18F/e-QIP-prototype/pull/395

afeld:
I'm starting to think that Codecov isn't set up to detect branches properly, either...

PR #669, opened 2018-08-12
Send test coverage for backend to codecov.io

ryanhofdotgov:


ryanhofdotgov:
This ended up being more complicated than I thought:
* `dep` doesn't support pulling down non-Go dependencies:
https://github.com/golang/dep/issues/1306#issuecomment-374789539
(would have preferred to pull `codecov` down that way, similar to how we pull for JS container)
* `codecov` bash takes different options that `codecov` Node, can only specify one coverage file per run, if not using its search function. The `codecov` paradigm not really designed for multi-language project layout we have.
* codecov.io doesn't work well (% totals, file links to GitHub, etc.) if you are submitting reports for sub-directories of a project, it really wants the top-level view of the project).

Code coverage statistics are abysmal at 6% for the backend. There also does not seem to be a Go mechanism for reporting test coverage on the conditional blocks in the XML templates. Frontend remains at 90%.

ryanhofdotgov:
Partially addresses #396

PR #670, opened 2018-08-13
fix navigation to "Review and submit"

afeld:
The `Attachments` and `Print` subcomponents were subclassing the `SectionElement` base class when they didn't need to be. This was causing a problem where it giving a warning for certain [`PropTypes`](https://reactjs.org/docs/typechecking-with-proptypes.html) not being set in those components, even though they weren't necessary.

Not exactly sure why this wasn't being caught in the linting/tests...


afeld:
Given that it's mostly complaining about lines whose formatting was changed, going to call that a "won't fix" (right now).

Issue #553, opened 2018-07-13
When Review section has errors, and it is the selected page, should have red "flair"


ryanhofdotgov:
**What did you do?**

Go to Review section with errors, make sure it is the selected page.

**What did you expect to see?**

Red "flair" to the left of the Review menu item in the navigation menu.

**What did you see instead?**

Blue "flair" to the left of the Review menu item.

<img width="697" alt="screen shot 2018-07-13 at 12 41 29 pm" src="https://user-images.githubusercontent.com/37100189/42718098-db56da7c-86ba-11e8-9a08-62faf6f81d61.png">

**Additional Information**

Consequence of nav menu rebuild in #482.

Using latest Chrome (67.0.3396.99) on macOS 10.12.

PR #672, opened 2018-08-13
Show accurate validity of review page in nav

hursey013:
Checks the validity of the entire section for the `Review` link and adds invalid styling if there are issues, and the default `usa-current` class if it is valid. Fixes https://github.com/18F/e-QIP-prototype/issues/553

**Before:**
screen shot 2018-08-13 at 3 43 09 pm

**After:**
screen shot 2018-08-13 at 3 42 06 pm

### DIFF
--- a/src/components/Navigation/SectionLink.jsx
+++ b/src/components/Navigation/SectionLink.jsx
@@ -43,14 +43,10 @@ class SectionLink extends React.Component {
     return className
   }
 
-  getActiveClassName () {
-    return this.hasErrors() ? '' : 'usa-current'
-  }
-
   render () {
     return (
       <li>
-        <NavLink to={this.href()} activeClassName={this.getActiveClassName()} className={this.getClassName()}>
+        <NavLink to={this.href()} activeClassName="usa-current" className={this.getClassName()}>
           <span className="section-name">
             {this.props.section.name}
           </span>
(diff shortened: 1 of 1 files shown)


src/components/Navigation/SectionLink.jsx
```diff
@@ -40,10 +43,14 @@ class SectionLink extends React.Component {
     return className
   }
 
+  getActiveClassName () {
+    return this.hasErrors() ? '' : 'usa-current'
```
afeld: Hmm, this is manipulating the classes to accommodate the CSS, no? I think we want it to have `usa-current` when a user is looking at that subsection, regardless of whether there are errors or not.

afeld:
Nice work on the fix otherwise, though!

src/components/Navigation/SectionLink.jsx
hursey013: Just want to make sure I'm following this. What you are describing above is how it currently functions, when a user goes to `Review`, `usa-current` is added which applies the blue line to the left of the link. Are you saying we still want `usa-current` added, but for it not to appear as the blue line?

hursey013:
@afeld pushed a new commit with my understanding of your request. Now if you land on the Review page and there are errors the Review link in the side navigation will have both the `usa-current` class along with `has-errors` (so proper styles are applied).

screen shot 2018-08-14 at 12 58 59 pm

src/components/Navigation/SectionLink.jsx
afeld: Correct!

Issue #130, opened 2018-05-23
Relationships - marital - location of marriage not saving - Data Persistence


ryanhofdotgov:
From e-QIP-prototype-truetandem created by [Jay-Neighbours-I72](https://github.com/Jay-Neighbours-I72) : ryanhofdotgov/e-QIP-prototype-truetandem#2916  

- [ ] Marriage location not saving

![screen shot 2018-04-27 at 10 20 10 am](https://user-images.githubusercontent.com/19935974/39367514-f02c60ea-4a04-11e8-8e43-56598e5f364d.jpg)

bpdesigns:
The location of Marriage seems to be persisting now, but "Provide this person's country(ies) of citizenship" is not persisting

ryanhofdotgov:
When I've read about these types of issues in the project's history it is usually because of a variable typo somewhere; e.g., part of the form calls it 'foobar' and some other layer calls it 'foo'

ryanhofdotgov:
What you've been doing so far has been great – reviewing issues of all kinds and verifying they are still reproducible. Persistence errors in particular are important (if there are still others), since they could prevent successful completion of form.

afeld:
We can/should also make automated tests for this: fill out a section, refresh, and check that the data displayed is the same. Shouldn't be _tooo_ hard.

ryanhofdotgov:
Yup, totally agree – here's an outstanding related ticket https://github.com/18F/e-QIP-prototype/issues/144 I asked TrueTandem to work on this before they got off contract, but they ran out of time. Once we finish the current iteration of getting the spec tests working, we should revisit.

hursey013:
It appears that `Provide the location` of the marriage and `Provide this person's country(ies) of citizenship` are persisting when testing the app locally - can anyone else reproduce this persistence issue?

ryanhofdotgov:
Yup, still happens for me unfortunately. Steps to reproduce, using Chrome's incognito window:

1) Sign in, expand Relationships, expand Marital & relationship status, click on Marital.
2) Click on "Currently in a civil marriage"
3) Scroll to Provide location. Click on Yes, in United States. Fill out City, State, County. 
4) Scroll down to bottom of page, click on Save, even if it says it was already just Saved.
5) Log out, close window.
6) In new incognito window, sign in again, go to "Provide location" – no data appears in fields.

PR #675, opened 2018-08-14
Add civil union location to api and db

hursey013:
Added database and API references for missing `Location` field. Fixes https://github.com/18F/e-QIP-prototype/issues/130


ryanhofdotgov:
@hursey013 One thing I missed in my review is that the migration script datestamp had a typo in the filename, it should have been today's date. But since it was monotonically increasing from the last migration script, that had no impact.

hursey013:
@ryanhofdotgov I don't know whats going on with me but that's like the third time this month I've written the date as July instead of August. I did try to update the timestamp but I accidentally made it `0714` instead of `0814`, my bad 😬

ryanhofdotgov:
Haha – no worries.

PR #676, opened 2018-08-14
Small UI tweak to side navigation

hursey013:
Anchors caret to the last word of section name in side navigation to avoid the icon breaking onto it's own line.

Before:
screen shot 2018-08-14 at 4 25 19 pm
screen shot 2018-08-14 at 4 26 41 pm

After:
screen shot 2018-08-14 at 4 23 40 pm
screen shot 2018-08-14 at 4 24 23 pm

### DIFF
--- a/src/components/Navigation/ToggleItem.jsx
+++ b/src/components/Navigation/ToggleItem.jsx
@@ -40,19 +40,12 @@ class ToggleItem extends React.Component {
   render() {
     const url = this.url()
     const active = this.isActive()
-    const sectionName = this.props.section.name.split(' ')
-    const lastWord = sectionName.pop()
 
     return (
       <li className="toggle-item">
         <a className={this.getClassName()} aria-controls={url} aria-expanded={active} role="button">
           <span className="section-name">
-            {sectionName.join(' ')}{' '}
-            <span className="no-wrap">
-              {lastWord}
-              <i className="fa fa-angle-up"></i>
-              <i className="fa fa-angle-down"></i>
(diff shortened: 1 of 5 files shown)


src/components/Navigation/ToggleItem.jsx
```diff
@@ -40,14 +40,19 @@ class ToggleItem extends React.Component {
   render() {
     const url = this.url()
     const active = this.isActive()
+    const sectionName = this.props.section.name.split(' ')
+    const lastWord = sectionName.pop()
```
afeld: Mind adding a comment about why this is being done?

hursey013:
In hindsight I realized this could be done with just CSS so removed the additional javascript and as an added bonus, removed some unnecessary HTML markup. The visual output still remains the same as the screenshots above.

afeld:
Is the `App` snapshot up-to-date? It shouldn't have changed that much...

You may need to add the [workaround for the generated field GUIDs](https://github.com/18F/e-QIP-prototype/blob/1ef519bab9d31f5c26b002003bb01d24247e66ab/src/components/Section/Identification/Identification.test.jsx#L19-L25). Trying out a better way to handle that across the board...

hursey013:
@afeld yeah I'm not sure what's up with the `App` snapshot, it doesn't seem relevant to my changes so perhaps it's out of date and was not previously committed - is that a possibility?

afeld:
Ahh, I see the problem. Pull request coming.

PR #689, opened 2018-08-15
use fake GUIDs for fields in tests

afeld:
Rather than having to put mocks in every snapshot of form fields, this ensures that fields will not have GUIDs that change with every test run.

**The catch:** Every field having the same GUID _may_ have implications for the tests, and make things behave in a strange way. I don't understand what they're used for enough to understand the implications.

There's definitely a better way to ensure consistency (like hashing the section, subsection, and field name?), but I'm not sure what it is.

### DIFF
--- a/src/components/Form/ValidationElement/ValidationElement.jsx
+++ b/src/components/Form/ValidationElement/ValidationElement.jsx
@@ -1,8 +1,9 @@
 import React from 'react'
+import env from '../../../config/environment'
 import { newGuid, flattenObject, mergeError, triageErrors } from './helpers'
 
 export default class ValidationElement extends React.Component {
-  constructor (props) {
+  constructor(props) {
     super(props)
 
     this.handleChange = this.handleChange.bind(this)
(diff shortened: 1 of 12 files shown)


hursey013:
@afeld having distinct (but consistent) guids appears to fix the affected tests

src/components/Form/ValidationElement/ValidationElement.jsx
```diff
@@ -67,26 +68,31 @@ export default class ValidationElement extends React.Component {
...
 
-  guid () {
-    return newGuid()
+  guid() {
+    // give a fake GUID so the field IDs don't differ between snapshots
+    if (env.IsTest()) {
+      return newMockGuid()
+    } else {
+      return newGuid()
+    }
```
afeld: Rather than having the conditional, maybe we should consider just using the counter...

hursey013: You mean in lieu of the randomly generated guid? Guess I don't really see why it needs to be completely unique, but I could definitely be missing something.

afeld: Yeah, would definitely want to have a better understanding of the GUID usage before doing that.

afeld:
Hmmmmm, one major downside of this is that running the tests in a different order (or a single test in isolation) changes the counter. It's therefore only deterministic to a point.

afeld:
Maybe we should just revert for now? The mocking isn't terrible...it just adds cruft to the test files.

PR #700, opened 2018-08-17
check that rendered components have a minimum number of elements

afeld:
This is to prevent the snapshots from looking like this accidentally:

```
exports[`The financial section renders the Financial component 1`] = `
<div>
  <div />
</div>
`;
```


src/components/Section/__snapshots__/Section.test.jsx.snap
```diff
@@ -1,7 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`The section component renders the Section component 1`] = `<div />`;
```
afeld: I didn't even know about this one - this pull request helped me catch it!

PR #702, opened 2018-08-17
generate snapshots for every section+subsection

afeld:
Builds on #700. Rather than needing snapshot tests in every section test file, this just does it once and for all. Downside is that it makes that one snapshot file _huge_.

* [ ] Figure out why tests are failing


afeld:
@hursey013 Thoughts on this approach?

afeld:
I'm tempted to put in an

```javascript
if (sectionThatIsntWorking) {
  return
}
```

and make an issue to come back to it, since that will be better than not having snapshots for the other sections.

afeld:
Thoughts about [the workaround to get this mergeable](https://github.com/18F/e-QIP-prototype/pull/702#issuecomment-413863118)?

afeld:
Ok, this should be ready!

PR #703, opened 2018-08-17
add a testSectionSnapshots() helper

afeld:
Alternative to #702, for slightly less boilerplate than what's on `develop` but still having the snapshots in every test file. Thoughts?


afeld:
Going with #702.

Issue #370, opened 2018-05-25
Adjust spacing for question in Investigative and criminal history > Police record > Offenses


bpdesigns:
Spacing for these questions should be closer together because they are grouped together as part of the same question. I would recommend cutting the space in half. 

<img width="527" alt="screen shot 2018-05-25 at 6 40 56 pm" src="https://user-images.githubusercontent.com/15329140/40568899-84215e7a-604b-11e8-950d-2a3137647d53.png">

 - What browser and version are you using?
Chrome
 - What operating system?
Mac

bpdesigns:
tagging this medium because it could confusion.

Issue #367, opened 2018-05-25
Extra line spacing in Military History > Disciplinary procedures 


bpdesigns:
**What did you do?**

viewing Military History > Disciplinary procedures

There is an extra space above the Text box for "Provide the name of the disciplinary procedure
Such as Court Martial, Article 15, Captain’s mast, Article 135 Court of Inquiry, etc."

**Additional Information**

<img width="493" alt="screen shot 2018-05-25 at 6 18 14 pm" src="https://user-images.githubusercontent.com/15329140/40568331-fbefbb26-6047-11e8-91e6-05d7cfd2538c.png">

 - What browser and version are you using?
 - What operating system?

bpdesigns:
Label: visual design, low

PR #718, opened 2018-08-21
Update spacing between elements

line47:
This is part of the work related to UI improvements which include 
Fixes https://github.com/18F/e-QIP-prototype/issues/370
Fixes https://github.com/18F/e-QIP-prototype/issues/367

- Lessening the space between questions
- Updating the width of accordions to flow with the left edge of the page
- Lessening the padding around the accordion contents
- Adding a container around the save button to better control the spacing of the button at different screen resolutions
- Increasing the page width and left justifying the page content instead of centering
- Removing styles that are no longer needed (cleanup)

## Screenshots

screen shot 2018-08-21 at 4 40 34 pm
screen shot 2018-08-21 at 4 40 41 pm
screen shot 2018-08-21 at 4 41 38 pm
screen shot 2018-08-21 at 4 43 48 pm


src/components/SavedIndicator/SavedIndicator.scss
```diff
@@ -121,3 +125,9 @@ $border-size: 5px;
   display: none;
   visibility: hidden;
 }
+
+@media only screen and (min-width: 1250px) {
```
hursey013: Should we make 1250px a global sass variable? Is it used elsewhere?

src/components/Section/SectionView.scss
```diff
@@ -1,22 +1,18 @@
 @import 'eqip-colors';
 
 .section-view {
-  max-width: 70rem;
+  max-width: 80rem;
```
hursey013: Since we are using 80rem as the max width for the content area and saved-indicator, we could probably turn that into a variable as well.

line47: Yeah I'm thinking that would be a larger refactoring effort. From poking around in the styles for a while I get the feel that there is _**a lot**_ we can refactor.

bpdesigns:
LGTM

PR #719, opened 2018-08-22
Add redux logger support

hursey013:
To support state debugging.


hursey013:
@ryanhofdotgov - rebased, this should be good now.

PR #727, opened 2018-08-23
Persist data with multiple employment entries

### LINKED ISSUE #693: Reason for leaving job not persisting
**What did you do?**

Fill out at least 2 multiple employment history entries. Specify value for "Provide the reason for leaving the employment activity" for any but the current employment entry. Save, log out, log back in. 

Values persisted.

**What did you see instead?**

No value in the reason field.

hursey013:
`PhysicalAddress`, `ReasonLeft`, and `Supervisor` fields were being matched solely on their entityID which meant they were not being matched within the context of the larger accountID. This caused the same entityID to be overwritten with data whenever a job was added or updated. Fixes #693.

### DIFF
--- a/api/physicaladdress.go
+++ b/api/physicaladdress.go
@@ -199,7 +199,7 @@ func (entity *PhysicalAddress) SetID(id int) {
 
 // Find the previous entity stored if one is available.
 func (entity *PhysicalAddress) Find(context DatabaseService) error {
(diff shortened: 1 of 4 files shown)


hursey013:
Also removed `name: this.props.name` from supervisor component since it doesn't actually exist.

Issue #574, opened 2018-07-17
Residence duration gaps not flagged 


michaelccata:
**What did you do?**

Checked validation on residence addresses for differing lengths of time and different size gaps.

**What did you expect to see?**

I'd expect all gaps in residences greater than 30 days to flag.

**What did you see instead?**

Gaps between residences larger than 30 days are not flagged if an applicant provides 10 years or more of housing. The gaps between residences larger than 30 days _are_ flagged if the total residences submitted sum to less than 10 years.

**Additional Information**
See photos attached.
We should also check employment timelines. 

 - What browser and version are you using?
Chrome
 - What operating system?
macos

michaelccata:
<img width="759" alt="screen shot 2018-07-17 at 1 02 12 pm" src="https://user-images.githubusercontent.com/4422873/42833554-62103c48-89c2-11e8-8005-843f4d661760.png">
^ is the response that shows gaps if the dates provided sum to less than 10 years of residence history

michaelccata:
<img width="802" alt="screen shot 2018-07-17 at 1 09 44 pm" src="https://user-images.githubusercontent.com/4422873/42833656-b1f68c62-89c2-11e8-94e8-ad785410661f.png">
^ residence gap greater than 30 days is not flagged if the sum total of history submitted is greater than 10 years

michaelccata:
<img width="848" alt="screen shot 2018-07-17 at 1 11 45 pm" src="https://user-images.githubusercontent.com/4422873/42833730-ff40049e-89c2-11e8-8a76-ba8f7f216977.png">
^ residence gap greater than 30 days is flagged if the sum total of history submitted is equal to 10 years

ryanhofdotgov:
Fixed by https://github.com/18F/e-QIP-prototype/pull/734

PR #734, opened 2018-08-27
Fix conditional statement for calculating date ranges

hursey013:
There was some logic introduced recently (https://github.com/ryanhofdotgov/e-QIP-prototype-truetandem/pull/2689) to suppress gaps when the date range of an entry met or exceeded the required date range of the section. I _think_ this was meant mostly to address an issue with the `People who know you well` section, since it was still showing gaps between entries even if one of the people spanned the entire 7 year period.

It looks like there was a mistake in the logic which causes some false positives and suppresses gaps even when they should be appearing as in https://github.com/18F/e-QIP-prototype/issues/574. Rather than using today's date to calculate the end of the desired range, it was using the date for the beginning of the range. So if _any_ entry has a start date that occurs before the required beginning date of the range all gaps would be suppressed. This corrects that to make sure the range begins before the required start date, and ends at today's date (with a 30 day buffer on each end of the range).

### DIFF
--- a/src/components/Section/History/dateranges.js
+++ b/src/components/Section/History/dateranges.js
@@ -2,7 +2,7 @@
  * Take one of the many schemas of a date and always output a valid
  * Date object or null.
  */
-export const extractDate = (dateObj) => {
+export const extractDate = dateObj => {
   if (dateObj instanceof Date) {
     return dateObj
   }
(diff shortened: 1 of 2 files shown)


ryanhofdotgov:
Closes #574

Issue #735, opened 2018-08-27
Test case #3 has bogus? validation errors


ryanhofdotgov:
If you sign into test case #3 on `dev` (use account listed [here](https://docs.google.com/document/d/17wH_EXRzrlC9KksbQlx-9uTMLjSLjSO199E7XAJh0Ho/edit#heading=h.j7v108xspr7u)), and then go to full form review, 2 entries in the Relationship sections are flagged, but when you open them no fields are shown as having an error.

The data was successfully submitted to e-QIP outside the GUI, skipping over this eApp validation.

PR #740, opened 2018-08-28
Do not require aliases for non-immediate family members

hursey013:
When entering relatives the `Has this relative used any additional names?` question is required unless the relative is not considered an immediate family member (in which case the question is hidden). This updates the validation logic to not flag non-immediate family members when the alias field is not answered. Fixes https://github.com/18F/e-QIP-prototype/issues/735

### DIFF
--- a/src/validators/relatives.js
+++ b/src/validators/relatives.js
@@ -159,6 +159,12 @@ export class RelativeValidator {
 
   validAliases() {
     const items = this.aliases.items || []
+    const nonImmediateFamily = ['Fosterparent', 'Father-in-law', 'Mother-in-law', 'Guardian']
+
+    if (nonImmediateFamily.includes(this.relation)) {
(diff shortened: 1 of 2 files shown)


src/validators/relatives.js
```diff
@@ -159,6 +159,12 @@ export class RelativeValidator {
 
   validAliases() {
     const items = this.aliases.items || []
+    const nonImmediateFamily = ['Fosterparent', 'Father-in-law', 'Mother-in-law', 'Guardian']
```
ryanhofdotgov: Looks good – I double-checked this list against the list on page 94 of [Validation Matrix SF 86 2017_08_07.pdf](https://drive.google.com/open?id=1_lIHjuKC2t8hNWOnKJDOAdF7Z7e8ylAI).

PR #765, opened 2018-09-07
Only send digits for selective service number

### LINKED ISSUE #750: Selective service numbers can only be digits
**What did you do?**

Entered in selective service number with dashes (e.g. `99-9999999-9`) using a pattern similar to what www.sss.gov reports my selective service number in. Submitted application to e-QIP.

Successful submission.

**What did you see instead?**

Error indicating that selective service number can only have digits.

**Additional Information**

We're limited to the lowest common denominator of what e-QIP accepts since data may have to be round-trip'ed through it or via its XML format. I have not been able to find an authoritative source that indicates that selective service numbers are uniformly sectioned with dashes in a particular way across the decades, so recommend just allowing digits.

mgwalker:
In the redux state, only store digits for the Selective Service registration number.  Within the UI, however, also allow dashes to match the output from sso.gov, but strip everything else.

Closes #750

### DIFF
--- a/src/components/Section/Military/Selective/Selective.jsx
+++ b/src/components/Section/Military/Selective/Selective.jsx
@@ -121,7 +121,6 @@ export default class Selective extends SubsectionElement {
                     name="RegistrationNumber"
                     className="registration-number"
                     label={i18n.t('military.selective.label.number')}
-                    name={this.props.RegistrationNumber.name}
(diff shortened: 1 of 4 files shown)


mgwalker:
Sets a pattern on the Selective Service number field, adds error text, and updates the validator.

PR #772, opened 2018-09-11
Fix maiden name date range by adding optional to date control validation

### LINKED ISSUE #746: Spouse's maiden name range should be enterable from birth
**What did you do?**

Enter in a spouse, providing a birthday for the spouse of 05/01/1992. Entered in a maiden name for the spouse, providing the birthday as the start of the date range.

No validation error.

**What did you see instead?**

Validation error saying start of range has to be after the birthday.

**Additional Information**

Alter validation logic to permit start of range to be on or after the spouse's birthday.

screen shot 2018-08-28 at 1 19 50 pm

apburnes:
This closes #746 by adding an optional property to allow the `datecontrol.js` to allow `minDate` and `maxDate` validations be `>=` or `<=` if passed the boolean props `minDateEquaTo` or `maxDateEqualTo`.  These are set to `false` by default to not impact another expected functionality.

Added the `minDateEqualTo` prop on the `<CivilUnion />` date range component to enable the maiden name from date to be equal to or greater than the birthdate.

### DIFF
--- a/src/components/Form/DateControl/DateControl.jsx
+++ b/src/components/Form/DateControl/DateControl.jsx
@@ -449,7 +449,9 @@ DateControl.defaultProps = {
   prefix: '',
   noMaxDate: false,
   maxDate: null,
+  maxDateEqualTo: false,
   minDate: null,
+  minDateEqualTo: false,
   relationship: '',
   toggleFocus: (w, changed, el, day, month) => {
     day.focus()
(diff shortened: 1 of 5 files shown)


hursey013:
I don't have a good grasp of all of the places the min/max date logic is being used within the app, but I wonder if there are any scenarios where we actually want it to be `<` or `>` rather than `<=` or `>=`? Maybe we should just default to `<=` and `>=` to avoid similar date comparison issues in other parts of the app?

apburnes:
I was thinking the same thing but wanted to keep this PR as unobtrusive as possible.  It may be worth putting together another ticket to identify all the date range scenarios.

PR #777, opened 2018-09-11
Cap currency values at positive 32-bit signed int

### LINKED ISSUE #749: Improve currency validation
e-QIP does additional validation that eApp does not currently do on currency fields, resulting in eApp validating forms that e-QIP subsequently rejects.

Specifically, currency amounts should not be greater than MAX_INT (2,147,483,647) in Java. Sending a value of 422577456456 for a bankruptcy amount resulted in a submission failure.

Probably good place to start is `src/validators/helpers.js` and any Currency field in a `*.jsx`.
```

### PR DESCRIPTION
**Pull request by mgwalker**
On the `Currency.jsx` component, sets a max value.  Adds an error message.  Updates the currency validator to enforce max value.

screen shot 2018-09-11 at 3 27 07 pm

closes #749

### DIFF
--- a/src/config/locales/en/error.js
+++ b/src/config/locales/en/error.js
@@ -777,6 +777,10 @@ export const error = {
     }
   },
   currency: {
+    max: {
+      title: 'Currency value is too large',
+      message: 'Currency amounts over $2,147,483,647 are not supported.'
+    },
     min: {
       title: 'There is a problem with the losses',
       message: 'The reported losses should have a dollar value.',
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/config/locales/en/error.js [hunk 1]
```diff
@@ -777,6 +777,10 @@ export const error = {
     }
   },
   currency: {
+    max: {
+      title: 'Currency value is too large',
+      message: 'Currency amounts over $2,147,483,647 are not supported.'
```
mgwalker: I guess if you've had an epic bankruptcy, you're just now allowed to have a clearance.  😜

PR #784, opened 2018-09-12
Reconcile phone number type options

### LINKED ISSUE #761: Reconcile telephone online fields with paper form/e-QIP
eApp uses the same telephone form field set for all telephone entry across the form:

screen shot 2018-09-06 at 8 26 53 am

This is sort of consistent with the paradigm for the applicant's contact info on the paper form (and e-QIP):

screen shot 2018-09-06 at 8 25 09 am

However, elsewhere on the paper form (and e-QIP) there is a different paradigm:
screen shot 2018-09-06 at 8 25 20 am

There is also a mismatch on the eApp form with respect to quantity; e.g. you can only have one home phone number for the applicant in e-QIP, not multiple. You can only have one evening phone, not multiple (for other places).

hursey013:
Fixes #761 

In reviewing the incoming xml schema, I was able to determine that there are only three different cases where specifying a phone number type (Home, Work, Mobile, etc) is supported. Those are when the applicant provides their own phone numbers, the phone numbers of people you know well, and when providing a reference for a residence. There were also no occurrences of allowing `Other` as a phone number type.

More specifically:
Your contact information: `[Home, Mobile, Work]`
Relationships: `[Home, Mobile]`
Residency: `[Day, Evening, Mobile]`

This PR removes `Other` as an option, turns off the phone number type picker by default, and adds validation to the applicant's own phone number entries to make sure that only one of each type is provided. 

It seems like the best approach may be to ultimately convert the `Your phone numbers` section to be more in line with the others where a separate field is provided for Home, Mobile, and Work rather than having the ability to add an infinite number of phone numbers. The form only allows a maximum of three, and the types cannot be duplicative (can't have two mobile numbers), so from a usability standpoint it may make sense to move away from this accordion pattern in the future.

Validation error when there are duplicate phone types:
screen shot 2018-09-12 at 9 02 36 pm


ryanhofdotgov:
This is a good approach – thanks! 

It maybe an unrelated, existing issue that I noticed when I pulled down the branch, but if I go to add a number and I don't click day only, night only, or both, I get a validation error at top-level accordion but nothing on the phone number to let me know that's the problem when I open it up.

Just double-checking, but does anything in the JSON serialization change across the form, aside from `Other` no longer being an option for `numberType`?  Will any of the `numberType` instances change from `Work`, etc. to `NA`? Trying to determine where I'll need to update my test fixtures in `api/testdata/complete-scenarios`, etc. so they match the current output of eApp.

I would open up a ticket for the design recommendation you are making and reference in this ticket before we close this one and move on.

hursey013:
@ryanhofdotgov I can't seem to reproduce the `day only, night only, or both` issue - what page are you on where you are seeing that?

ryanhofdotgov:
One thing I should add though, is that I purge my database and re-load from complete test case scenarios, so that's always in the back of my mind that that could be the cause of an issue if the JSON serialization has changed at all.

hursey013:
One limitation of this current approach is that the validation message appears separate from the list of field specific errors found at the top of the Review page. I've opened a new issue to document this and would like to find a cohesive way to combined field-specific and collection-level errors within the same UI in the future: https://github.com/18F/e-QIP-prototype/issues/788

PR #791, opened 2018-09-13
Add zip code to state validation when state address entered

### LINKED ISSUE #631: Require that state and zip code be consistent
e-QIP reject an application submission if it provides an address where the state and zip code don't match; i.e., you can't specify California as the state and put down a zip code for Honolulu.

Since we let applicants ignore the validation error provided by USPS, this results in a situation where eApp can validate a form, but if submitted, it will fail at the e-QIP level.

Modify eApp so that it requires that states/zip codes are consistent.

apburnes:
Closes #631 by adding validation to the `<Address />` form component where the zip code must be from the entered US state.

- Statically add state zip code json files
- Adds zip code in state method to the validator in `location.js`
- Updates the `<Address />` form
- Passes down a `status` prop to the `<Generic />` form used to construct the `<Zipcode />` to handle errors
- Adds error message for `address.zipcode.status` in locales `error.js`


ryanhofdotgov:
@apburnes I think there is a way to get the validation error to show on the accordion when it is rolled up. Can you figure out how to do that? Currently it looks like this:

screen shot 2018-09-13 at 1 50 13 pm

Here is an example of what I mean:

screen shot 2018-09-13 at 1 52 17 pm

hursey013:
@ryanhofdotgov longer term question, from a maintenance perspective what's the expectation on keeping the static zip code data up to date?

ryanhofdotgov:
The expectation is that that someone will have to maintain this static list of states/zip codes. The current e-QIP takes a similar approach, so in that respect if this implementation sticks around then it isn't adding any more burden than the existing system. 

We could just offload this to USPS too (i.e., Andrew's original plan), but:
* Right now, that's more moving parts than I'd like to change this month (e.g., getting Andrew up to speed on backend, adding another backend API call that technically changes security footprint triggering more verification work, etc.)
* I'm not even sure that would be ideal, because this also solves the problem of submit-ability when USPS service is not available; i.e. if USPS goes down or the server IP gets temporarily banned then this would help prevent users from creating a bunch of applications that e-QIP won't accept.

Something half way between is to push the static list to the backend, but that still has the first problem.

apburnes:
The static zip codes are +550KB and sometimes it feels a bit sluggish on the UI.  Based on the attached map, would writing a state zip code approximation algorithm be worth it?


zip_code_zones

ryanhofdotgov:
@apburnes so much for my 5-10KB compressed 😂

apburnes:
Added additional validation before running the zip code in state validation.  I think for now it will do and  looking at writing an algorithm for the future should be an easier update in the `location` validator.

PR #792, opened 2018-09-13
Make tests, XML consistent with new form changes

### LINKED ISSUE #789: Tweak complete scenario test case JSON files in backend
Reference #784

ryanhofdotgov:
Fixes #789


api/telephone.go
```diff
@@ -10,6 +10,7 @@ var (
 	formatTelephoneDomestic      = regexp.MustCompile("\\d{3}-?\\d{3}-?\\d{4}")
 	formatTelephoneInternational = regexp.MustCompile("\\d{3}-?\\d{4}-?\\d{4}")
 	formatTelephoneDSN           = regexp.MustCompile("\\d{3}-?\\d{4}")
+	formatNumberType             = regexp.MustCompile("(Home|Work|Cell|NA|^$)")
```
hursey013: `NA` is no longer an option since choosing `Home|Work|Cell` is required, but probably doesn't hurt to keep it here.

Issue #781, opened 2018-09-12
Only collect 2 email addresses, home or work


ryanhofdotgov:
For the applicant, eApp can collect unbounded number of emails, unspecified by type. Only collect a home email address and a work email address, just like the SF-86.

Will require changes to XML generation.

PR #797, opened 2018-09-14
Use separate fields from applicant's home and work email addresses

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/781

Currently we allow for an applicant to add an infinite number of personal email addresses even though the form only allows for a home and work email address. This refactors the email address section into a two separate fields instead of the accordion based pattern.

**Before:**
screen shot 2018-09-14 at 4 35 01 pm

**After:**
screen shot 2018-09-14 at 4 30 01 pm

_*Sorry for the unrelated formatting changes in the json files, it looks like I previously missed those when I ran `prettier` on the code base._

### DIFF
--- a/api/identification.go
+++ b/api/identification.go
@@ -602,16 +602,19 @@ func (entity *IdentificationSSN) Find(context DatabaseService) error {
 
 // IdentificationContacts represents the payload for the identification contact information section.
 type IdentificationContacts struct {
-	PayloadEmails       Payload `json:"Emails" sql:"-"`
+	PayloadHomeEmail       Payload `json:"HomeEmail" sql:"-"`
+	PayloadWorkEmail       Payload `json:"WorkEmail" sql:"-"`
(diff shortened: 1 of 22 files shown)


apburnes:
Everything works and feels as expected when running through the forms. The `IdentificationContactInformationValidator` test not passing makes me wonder how the section is running the validator.

hursey013:
@ryanhofdotgov it looks like its the XML test that is failing, I took a guess at what the new syntax should look like but I must have missed something: https://github.com/18F/e-QIP-prototype/pull/797/files#diff-c494a8f31682dacf2b1a4efa95233f32R3

ryanhofdotgov:
@hursey013 I'll take a look at those XML test cases

ryanhofdotgov:
@hursey013 Would you review the last 2 commits I just made and if it looks good, merge in this PR in conjunction with Andrew's approval with the rest of the changes?

PR #798, opened 2018-09-17
Update zip code search in state search to first 3 digits

apburnes:
Using the first three digits is the minimum needed to determine if zip code is in the state. Removes a lot of static data and cleans up the validation process.

[Zip code reference](https://en.wikipedia.org/wiki/List_of_ZIP_code_prefixes)


ryanhofdotgov:
@apburnes If you are only doing the 1st 3 numbers, is it possible for an applicant to put in an unallocated zip code (i.e., can a user still put in a zip code that e-QIP will fail?).

hursey013:
@ryanhofdotgov is it possible to use the e-QIP response to provide feedback to the frontend (in the event that it fails even after passing the three number validation)?

ryanhofdotgov:
@hursey013 Unfortunately, that's not a easy to pull off:
* e-QIP doesn't give you the specific address that has the bad zip code, it just cites the SF-86 section, it's not always a straightforward mapping.
* eApp isn't going to be talking directly to e-QIP any more. Another component is going to sit in-between and asynchronously communicate the SF-86 XML and attachments to e-QIP. So a failed submit at that level is going to be much like a kick-back from a reviewer.

apburnes:
@ryanhofdotgov not sure if a passing 3 digit zip with invalid last two digits would be rejected my e-Qip. It may be worth testing or noting to revisit down the road.

ryanhofdotgov:
Just to clarify, a bad zip code at the eApp level would look something like:
* eApp submit to the upstream component in between us and e-QIP (success)
* That component submits to e-QIP (failure)
* Someone has to notice that there was an error and inspect the logs
* They contact us, we identify problem, contact them back
* Application gets kicked back to applicant to fix zip code (notified via email, manually typed note explaining what went wrong, how to fix it)
* Applicant signs into eApp again, submits again

apburnes:
@ryanhofdotgov we definitely want to avoid that scenario. I guess this should probably down the road for future fixing.

ryanhofdotgov:
Tested e-QIP and it is only doing 3 digit zip code validation. So lets go ahead and use this approach. 
 I'll note that e-QIP does not account for the dozen zip codes that are [shared across 2 states](https://gis.stackexchange.com/questions/53918/determining-which-us-zipcodes-map-to-more-than-one-state-or-more-than-one-city). I follow up on that with NBIB.

PR #809, opened 2018-09-17
Ask for National Guard service state

### LINKED ISSUE #752: Prompt and require state of service if national guard selected
If you are filling out military history and specify ArmyNationalGuard or AirNationalGuard, the UI must prompt applicant to specify the State of service (e.g., CA, OR, WA, etc.). e-QIP XML should be updated as well to process this value.

See validation matrix, (15.2.2) State of service, and Applicant Responses XSD.

hursey013:
Fixes #752

Adds a `State of Service` field when Army or Air National Guard are selected in military history. Also created a new location layout for `State` only.

screen shot 2018-09-18 at 11 24 22 am

### DIFF
--- a/api/location.go
+++ b/api/location.go
@@ -14,6 +14,7 @@ const (
 	LayoutCountry                             = "Country"
 	LayoutUSCityStateInternationalCity        = "US City, State, International city"
 	LayoutUSCityStateInternationalCityCountry = "US City, State, International city country"
+	LayoutState                               = "State"
 	LayoutCityState                           = "City, State"
 	LayoutStreetCityCountry                   = "Street, City, Country"
 	LayoutCityCountry                         = "City, Country"
@@ -135,6 +136,8 @@ func (entity *Location) Valid() (bool, error) {
 		} else {
 			stack = validateFields(entity, "city", "country")
 		}
+	case LayoutState:
+		stack = validateFields(entity, "state")
 	case LayoutCityState:
 		stack = validateFields(entity, "city", "state")
 	case LayoutStreetCityCountry:
--- a/api/templates/military-history.xml
+++ b/api/templates/military-history.xml
@@ -7,8 +7,8 @@
     <Branch>{{radio $Item.Service}}</Branch>
     <Comment></Comment>
     {{monthYearDaterange $Item.Dates}}
-    {{if or (radio $Item.Service | eq "ArmyNationalGuard") (radio $Item.Service | eq "ArmyNationalGuard")}}
-    <GuardLocation>{{location $Item.GuardLocation}}</GuardLocation>
+    {{if or (radio $Item.Service | eq "AirNationalGuard") (radio $Item.Service | eq "ArmyNationalGuard")}}
+    <GuardLocation><State>{{$Item.GuardLocation.props.state}}</State></GuardLocation>
     {{end}}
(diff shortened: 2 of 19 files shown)


hursey013:
@ryanhofdotgov I'm guessing this might need some adjustment: https://github.com/18F/e-QIP-prototype/pull/809/files#diff-ec4f5f728ad6a448022e39963d457dafR11 - I noticed the use of `<Place></Place` elsewhere, not sure if that is needed here as well?

api/templates/military-history.xml
```diff
@@ -7,9 +7,9 @@
     <Branch>{{radio $Item.Service}}</Branch>
     <Comment></Comment>
     {{monthYearDaterange $Item.Dates}}
-    <!-- <GuardLocation> -->
-    <!--   <State>DQ</State> -->
-    <!-- </GuardLocation> -->
+    {{if or (radio $Item.Service | eq "ArmyNationalGuard") (radio $Item.Service | eq "ArmyNationalGuard")}}
+    <GuardLocation>{{location $Item.GuardLocation}}</GuardLocation>
```
ryanhofdotgov: Yeah, we need to do something different here per the XSD:
```
<GuardLocation><State>{{$Item.GuardLocation.props.state}}</State></GuardLocation>
```

api/location.go
```diff
@@ -14,6 +14,7 @@ const (
 	LayoutCountry                             = "Country"
 	LayoutUSCityStateInternationalCity        = "US City, State, International city"
 	LayoutUSCityStateInternationalCityCountry = "US City, State, International city country"
+	LayoutState                               = "State"
```
ryanhofdotgov: I don't see a `.go` file that addresses how this new field gets persisted. Does that work auto-magically? Have you done a smoke test where you set the state, log out, open new incognito browser, log in, and see data?

api/templates/military-history.xml
ryanhofdotgov: Also there's a copy and paste error – Army National Guard listed twice.

api/location.go
hursey013: Yeah I've tested it and it persists, I followed the same format as the other location layouts so I suspect it's inheriting some of the existing functionality.

api/templates/military-history.xml
hursey013: Thanks, xml updated

ryanhofdotgov:
@hursey013 I think we will need to tag team these types issues for the next few weeks, as fully completing this type of changes requires:
1) Access to NITSL (so you can submit XML to e-QIP)
2) Access to e-QIP AV

Also requires a little bit of knowledge on how to clear database, load up test cases into system from command-line, etc. that I've developed over past few weeks, but haven't documented yet – Doh!

Finally, I think for any new field we are adding or changing, we need to make sure it is represented in one of the 5 full test cases (or a new one) where we can easily visually inspect the output and submit it to e-QIP. The other tiny, self-contained test case snippets only exercise so much.

After you review my commit contribution, would you merge in?

hursey013:
@ryanhofdotgov yeah I do feel like I'm flying blind on some of these changes without a way to fully test them myself, but my main hope is to minimize the amount of work you have to do once I pass it over. If there are aspects of these types of changes that I'm missing or if they're parts that make more sense for you to do just let me know. 

Changes look good, merging in!

PR #825, opened 2018-09-18
Encode reason for quitting in correction location

ryanhofdotgov:
Resolves row 75 of NBIB defects spreadsheet.


api/templates/history-employment.xml
```diff
@@ -104,9 +104,9 @@
         {{- if eq (branch $Dismissal.Has) "Yes"}}
         <Dismissal ID="{{inc $lindex}}">
           <Date Type="{{dateEstimated $Dismissal.Date}}">{{monthYear $Dismissal.Date}}</Date>
-          <Reason>{{textarea $Item.ReasonLeft.props.ReasonDescription}}</Reason>
```
hursey013: Is `ReasonLeft.props.ReasonDescription` no longer used anywhere?

ryanhofdotgov: Yeah, just in a different part of this section – history-employment.xml" line 160.

Issue #373, opened 2018-05-29
Generate and store application data archival PDF on backend


ryanhofdotgov:
As a user, I want to be able to download my application data in archival format (i.e., PDF) even if my browser or OS does not support save to PDF.

As system owner, I want an application data archival PDF to be persisted like our other archival PDFs (certifications, medical releases, etc) as an official record.

ryanhofdotgov:
Cryptographic hash of application data was stored in database in #372. Deferring remaining functionality.

ryanhofdotgov:
Long term plan is for this to be delegated to a system external to eApp, but we need a short term solution.

One idea I had was something like this:
https://medium.com/compass-true-north/go-service-to-convert-web-pages-to-pdf-using-headless-chrome-5fd9ffbae1af

Any requirements beyond the Go runtime need to be coordinated with NBIS infrastructure team.

macrael:
Is it desirable for the data to be in the same format as the original form? We are using some go based pdf generation code in another project where we have a pdf of a paper form and insert images of text into the appropriate boxes and then render the whole thing to a pdf in the end.

ryanhofdotgov:
The NBIS product owner wanted to do the least amount of effort to get something functional, since it is probably all going to get thrown away in 2-4 months, so that probably drives the technology/design choice more than anything. They were hoping to skip the effort around reviewing a new PDF design at this time as well as minimize development time.

Historically, the archival PDF in e-QIP served at least a couple of different purposes:
* Allow applicant to store their data themselves, offline
* Allow them to easily re-enter it in e-QIP if e-QIP for some reason didn't preserve their data on a subsequent background check

What Go PDF library were you thinking of? Rendering text as images is likely a non-starter anyways as these PDFs really need their form values searchable and copy/paste-able. 

When I implemented the signature PDFs in eApp at what I thought was at "the last hour", I couldn't find anything appropriate other than https://unidoc.io, which would require a commercial license for this project. At that time we couldn't stray from Go-based solutions, due to constraints imposed by our interim authority to test (IATT).

PR #829, opened 2018-09-19
Update CSS for selection of button when printing

### LINKED ISSUE #819: USPHS Commissioned Corps is truncated on button label
This is in employment history section:

screen shot 2018-09-18 at 8 39 02 am

### LINKED ISSUE #820: Truncated text in alcohol abuse/illegal drugs section
See attached hand-generated eApp archival PDF, pages 123 and 137. Use PDF to identify section pages that need to be updated. Screenshots also below.

[test2a-archival-5546.pdf](https://github.com/18F/e-QIP-prototype/files/2393486/test2a-archival-5546.pdf)
screen shot 2018-09-18 at 8 52 08 am
screen shot 2018-09-18 at 8 51 54 am

line47:
This fixes #819 
This also fixes #820 

**Screenshots**
screen shot 2018-09-19 at 10 55 40 am
screen shot 2018-09-19 at 10 56 06 am

### DIFF
--- a/src/components/Form/Field/Field.scss
+++ b/src/components/Form/Field/Field.scss
@@ -142,7 +142,6 @@
         max-width: 64rem;
 
         > div {
-          width: 100%;
           max-width: 64rem;
         }
 
--- a/src/sass/print.scss
+++ b/src/sass/print.scss
@@ -256,11 +256,12 @@
 
   /* Set selected checkboxes with thick border */
   .block label.checked {
-    border: 3px solid black;
+    border: 1px solid transparent;
     background-color: #fff;
     color: #000;
     font-weight: 600;
     opacity: 1;
+    outline: 3px solid #000;
   }
 
   /* Get rid of left blue border within accordion */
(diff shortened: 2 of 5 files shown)


src/components/Form/Field/Field.scss
```diff
@@ -142,7 +142,6 @@
         max-width: 64rem;
 
         > div {
-          width: 100%;
```
line47: Was in the neighborhood and removed this (divs are 100% by default 😸)

src/sass/print.scss
```diff
@@ -256,11 +256,12 @@
 
   /* Set selected checkboxes with thick border */
   .block label.checked {
-    border: 3px solid black;
+    border: 1px solid transparent;
     background-color: #fff;
     color: #000;
     font-weight: 600;
     opacity: 1;
+    outline: 3px solid #000;
```
line47: Switched to outline instead of border. Border restricts the inside space and outline does not 👍

ryanhofdotgov:
Looks good – this solves the particular itch we needed to scratch. There is a clipping problem with the outside border on some of the buttons when the button is selected and up against the margin, but we can live with that, as I'm not certain how much life this print to PDF feature has. We have an [issue to replace it](https://github.com/18F/e-QIP-prototype/issues/373) (or it may get implemented in a different system altogether)

screen shot 2018-09-19 at 2 47 35 pm

PR #839, opened 2018-09-20
Remove passport card option

### LINKED ISSUE #818: Remove passport card option
Don't allow user to select Passport Card vs Passport. Just prompt for the passport number. You can move guidance about starting with "A" for passport vs "C" for Card in the help text.

hursey013:
Fixes #818

This removes the `Passport Card` option from the `Foreign associations > U.S. passport information` UI, also updates the validation and i18n files to reflect the changes. The database field for Passport Card still remains and will be documented in a [separate issue](https://github.com/18F/e-QIP-prototype/issues/840) should it need to be reenabled in the future.

**Before:**
screen shot 2018-09-20 at 2 28 15 pm

**After:**
screen shot 2018-09-20 at 2 24 38 pm

### DIFF
--- a/src/components/Section/Foreign/Passport/Passport.jsx
+++ b/src/components/Section/Foreign/Passport/Passport.jsx
@@ -296,7 +296,7 @@ Passport.defaultProps = {
   Comments: {},
   HasPassports: {},
   suggestedNames: [],
-  reBook: '^[a-zA-Z0-9]{6,9}$',
+  reBook: '^[a-zA-Z0-9]{9}$',
   onUpdate: queue => {},
   onError: (value, arr) => {
     return arr
--- a/src/components/Section/Foreign/Passport/Passport.jsx
+++ b/src/components/Section/Foreign/Passport/Passport.jsx
@@ -219,7 +219,7 @@ export default class Passport extends SubsectionElement {
                     `foreign.passport.placeholder.bookNumber`
                   )}
                   pattern={this.props.reBook}
-                  maxlength="10"
+                  maxlength="9"
                   className="number passport-number"
                   ref="number"
                   prefix="passport"
@@ -282,7 +282,7 @@ Passport.defaultProps = {
   Comments: {},
   HasPassports: {},
   suggestedNames: [],
-  reBook: '^[a-zA-Z]{1}[0-9]{6,9}$',
+  reBook: '^[a-zA-Z0-9]{6,9}$',
   onUpdate: queue => {},
   onError: (value, arr) => {
     return arr
--- a/src/components/Section/Foreign/Passport/Passport.jsx
+++ b/src/components/Section/Foreign/Passport/Passport.jsx
@@ -265,9 +265,6 @@ export default class Passport extends SubsectionElement {
                   label={i18n.t(
                     `foreign.passport.label.bookNumber`
                   )}
-                  placeholder={i18n.t(
-                    `foreign.passport.placeholder.bookNumber`
-                  )}
                   pattern={numberRegEx}
                   maxlength={numberLength}
                   className="number passport-number"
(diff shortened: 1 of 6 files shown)


src/components/Section/Foreign/Passport/Passport.jsx
```diff
@@ -311,7 +283,6 @@ Passport.defaultProps = {
   HasPassports: {},
   suggestedNames: [],
   reBook: '^[a-zA-Z]{1}[0-9]{6,9}$',
```
apburnes: I happened to have my passport near by and it has a type letter and then a 9 digit passport number.  When I entered that schema into the form, I cannot input all 9 digits into the form. Looks like the regex only allows 9 places including the letter. Should this be updated to `7-10` to allow the initial letter space?

apburnes: Looks like the regex is fine. Just need to make the max length of `10`

src/components/Section/Foreign/Passport/Passport.jsx
```diff
@@ -215,38 +209,16 @@ export default class Passport extends SubsectionElement {
...
-                    `foreign.passport.label.${passportType.toLowerCase()}Number`
+                    `foreign.passport.label.bookNumber`
                   )}
                   placeholder={i18n.t(
-                    `foreign.passport.placeholder.${passportType.toLowerCase()}Number`
+                    `foreign.passport.placeholder.bookNumber`
                   )}
-                  pattern={re}
+                  pattern={this.props.reBook}
                   maxlength="9"
```
apburnes: Increase to `10`

src/components/Section/Foreign/Passport/Passport.jsx
hursey013: The help text is also a little misleading as well, the passport number in my book doesn't technically begin with a letter... I guess that is referring to the `Type` letter?

apburnes: Should we turn it into two inputs? One is a letter type and the other is a 6-9 digit number?

hursey013: This seems inaccurate as well `Passports will start with the letter A.` https://github.com/18F/e-QIP-prototype/pull/839/files#diff-f3e54c23a039ce2474f26a85429c3bc8R79, @ryanhofdotgov do you know the source of that help text? The paper SF-86 makes no mention of needing a letter at all.

ryanhofdotgov:
@apburnes @hursey013  Yeah, I don't know where that help text came from. We should remove it.

Any time we are changing the fields, we should look at the validation matrix to double-check. On page 6 it says:
* If issue date >= 1/1/1990, verify it is a exactly nine characters, alphanumeric
* if less than 1/1/1990, any ASCII string

Here is another .gov reference: 
https://www.e-verify.gov/about-e-verify/e-verify-data/e-verify-enhancements/june-2011

It seems to contradict the validation matrix. Let me do some tests against e-QIP before we call this "done".

ryanhofdotgov:
I've checked e-QIP, it does what the validation matrix says. Let's implement that. To be clear, we are not soliciting the passport type code, even though that might be implied from the previous help text.

Also, on second read, that e-verify.gov document doesn't really contradict what we see in the validation matrix, just isn't as precise as to the condition a less than nine character passport number might come into play.

hursey013:
Alright, updated with more accurate help text and adjusted the regex to accommodate alphanumeric strings between 6 and 9 characters, loosening it a bit to account for 9 digit number only strings (like mine).

hursey013:
Ready for another look - takes passport issue date into consideration in choosing the proper number validation (before 1990, alphanumeric string, no length limit. after 1990, alphanumeric string, 6-9 chars).

Issue #834, opened 2018-09-19
Clean up/document load/flush tools 


ryanhofdotgov:
For NBIB testing, I've extended the existing `load` tool and have a process of purging/loading up test cases.

Clean up the code and document the process.

PR #842, opened 2018-09-20
Tools/docs to purge/export/load test case scenarios

ryanhofdotgov:
Fixes https://github.com/18F/e-QIP-prototype/issues/834


ryanhofdotgov:
@hursey013 Fleshed out the related tooling and documentation.

PR #843, opened 2018-09-20
Remove allowNotApplicable from reference phone number

### LINKED ISSUE #654: Require phone number for someone who can verify unemployment
**What did you do?**

Specified a period of unemployment on form. Specified a person who could verify and their address. No phone number was given. Submit to e-QIP.

No errors on submission.

**What did you see instead?**

Error requiring that phone number be provided for the person verifying period of unemployment.

**Additional Information**

Specifically was testing unemployment branch, but issue probably affects all types of employment as there is a "I don't know" option available for phone number.

Pages 49, 51 in [Validation Matrix](https://drive.google.com/open?id=1_lIHjuKC2t8hNWOnKJDOAdF7Z7e8ylAI) confirms requirement; i.e., "I don't know" should not be an option.

screen shot 2018-08-06 at 9 50 34 am

apburnes:
This updates the reference phone number for all employment activities to be required (not just unemployment).

Closes #654 

**Before**
screen shot 2018-09-20 at 3 17 24 pm

**After**
screen shot 2018-09-20 at 3 03 42 pm

### DIFF
--- a/src/validators/employment.js
+++ b/src/validators/employment.js
@@ -164,7 +164,7 @@ export class EmploymentValidator {
   validReference() {
     return (
       new NameValidator(this.referenceName).isValid() &&
-      validPhoneNumber(this.referencePhone) &&
+      validPhoneNumber(this.referencePhone, { numberType: false }) &&
       new LocationValidator(this.referenceAddress).isValid()
     )
   }
--- a/src/validators/employment.js
+++ b/src/validators/employment.js
@@ -164,7 +164,7 @@ export class EmploymentValidator {
   validReference() {
     return (
       new NameValidator(this.referenceName).isValid() &&
-      validPhoneNumber(this.referencePhone, { numberType: false }) &&
+      validPhoneNumber(this.referencePhone) &&
       new LocationValidator(this.referenceAddress).isValid()
     )
   }
--- a/src/components/Section/History/Employment/EmploymentItem.jsx
+++ b/src/components/Section/History/Employment/EmploymentItem.jsx
@@ -492,6 +492,8 @@ export default class EmploymentItem extends ValidationElement {
                   name="ReferencePhone"
                   className="reference-phone"
                   {...this.props.ReferencePhone}
+                  allowNotApplicable={false}
+                  showNumberType={true}
                   onUpdate={this.updateReferencePhone}
                   onError={this.props.onError}
                   required={this.props.required}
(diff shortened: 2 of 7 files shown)


ryanhofdotgov:
Unfortunately, it looks like you'll need to remove that 2nd line you added that shows phone types. Sorry – all this defect tracking in spreadsheets, issue tracking in GitHub, branches, reviews, for one line of code.

To explain in detail, telephone numbers are collected in a varied way across the form. For this one, we don't want the phone number type because the SF-86 doesn't have it and we can't deviate without violating policy. Here is a screenshot from the SF-86.

screen shot 2018-09-20 at 3 42 57 pm


You can also verify via the [https://drive.google.com/open?id=1R7ZG4zs6gj9WF1Tb1jmzyzvBHrBXoMDK](SF86_2017-07_ApplicantResponses.xsd) (haven't received sign-off to publish that to our repo. Sometimes the validation matrix gives insight too.

apburnes:
Sounds good... and I thought I was going to get away with a two line commit.

apburnes:
Are number types for office number and supervisor number used on the form?  If yes, they are currently not recorded and if no, they will need to be updated to allow a null number type.

ryanhofdotgov:
Nah, not those either:

screen shot 2018-09-21 at 7 04 13 am

These screenshots are from the hard-copy SF-86 PDF, available here:
https://www.opm.gov/forms/standard-forms/

Brian also did an analysis over the from recently:
https://github.com/18F/e-QIP-prototype/issues/761

src/validators/employment.js
```diff
@@ -164,7 +164,7 @@ export class EmploymentValidator {
   validReference() {
     return (
       new NameValidator(this.referenceName).isValid() &&
-      validPhoneNumber(this.referencePhone) &&
+      validPhoneNumber(this.referencePhone, { numberType: false }) &&
```
ryanhofdotgov: @apburnes @hursey013 Is this something we need to do all over the application, since most don't have a number Type? Or should this be done in the inverse manner (i.e., explicitly turn it on for the 1? 2? places where number type is used)

apburnes: I like that better. What should the default validation be? With `numberType` or without `numberType`?

hursey013: Without, there is only one spot in the entire application that uses `numberType` (when applicant enters their own contact phone numbers).

apburnes:
Ended up staying with using an option to require `numberType` validation and switched the default setting.  Now only the `identificationcontact` requires phone number type.

PR #849, opened 2018-09-24
Show accordion contents when printing

### LINKED ISSUE #701: Archival PDF doesn't print all the sections
**What did you do?**

Save archival PDF after submitting.

All the sections, fully expanded.

**What did you see instead?**

Just the very briefest summary for people the applicant knows, their relatives, employment, school, etc.

**Additional Information**

A painful, quasi-workaround is to go into each section with expandable content, open each entry for relative, workplace, school, etc. and print the section page or review page.

line47:
This fixes #701

### DIFF
--- a/src/sass/print.scss
+++ b/src/sass/print.scss
@@ -248,7 +248,7 @@
     background-color: #fff;
     // Open accordion contents on print
     .details.close.static {
-      max-height: 0;
+      max-height: 100%;
       visibility: initial;
     }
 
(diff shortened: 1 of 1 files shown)


ryanhofdotgov:
@line47 I pulled this branch, down but it doesn't look like things are getting expanded for me. Can you take a look at page 9 of this archive I produce from test case 5? Thanks!

[test5-auto-archival.pdf](https://github.com/18F/e-QIP-prototype/files/2412476/test5-auto-archival.pdf)

ryanhofdotgov:
That seems to have been the trick - thanks. The printing the PDF this way for `test5.json` is almost usable. It's about 100 pages longer than if I stitch together 10 section review PDFs – I think mostly due to some odd margins applied. But that will be a different ticket.

PR #852, opened 2018-09-24
Revert "Allow opting in to the redux logger, off by default"

ryanhofdotgov:
Reverts 18F/e-QIP-prototype#741


ryanhofdotgov:
PR #741 breaks IE11:
```
SCRIPT5009: 'URLSearchParams' is undefined
```
Just get a blank screen when navigating to site URL.

Issue #824, opened 2018-09-18
Test cases have phantom validation errors


ryanhofdotgov:
**What did you do?**

Load up `api/testdata/complete-scenarios/test1.json` into an flushed/empty account with `load-scenario` command-line tool. Go to overall review section and validate. 

Same for test cases 2 and 3.

**What did you expect to see?**

No validation errors.

**What did you see instead?**

Validation errors in multiple sections. When you drill down to find the errors, there is nothing specifically flagged.

**Additional Information**

e-QIP does not flag any validation errors when submitted, so likely a false positive in eApp logic.

Completed ticket corrects and verifies that the following test cases do not have any validation errors when loaded:
- [x] test1.json
- [x] test2.json
- [x] test3.json
- [x] test4.json
- [x] test5.json
- [x] test6.json

ryanhofdotgov:
@apburnes Just as a heads up, most of the errors seem to be the same phantom table errors we've had for a few weeks at least, but I'm also noticing false zip code errors now (e.g., zip code flagged as bad when it is definitely good). I'm going to publish some docs on my load tools today so everyone can easily take a look for themselves.

hursey013:
Using https://github.com/18F/e-QIP-prototype/commit/a326d1a1c0b824de50bc89b5ca9f6970dc156055 and wiping the db before each test case, here are the results I've gotten:

- Test 1: Successful review
- Test 2: Phantom error, `Financial record > Delinquent payments`
- Test 3: Successful review
- Test 4: Successful review
- Test 5: Phantom error, `Relationships > Marital, Cohabitants, Relatives` and `Investigative and criminal history > Offenses`
- Test 6: Successful review

PR #855, opened 2018-09-24
Provide complete date for delinquent payment

hursey013:
One of the date entries for `Delinquent payments` was missing the day portion of the date field. In this particular case, only the month and year are asked for, but the UI will send a hidden value of `1` for the day portion along with the month/year, but for whatever reason it doesn't look like that made it into the test json. This relates to https://github.com/18F/e-QIP-prototype/issues/824

### DIFF
--- a/api/testdata/complete-scenarios/test2.json
+++ b/api/testdata/complete-scenarios/test2.json
@@ -898,7 +898,7 @@
                     "type": "datecontrol",
                     "props": {
                       "month": "01",
-                      "day": "",
+                      "day": "1",
                       "year": "1999",
                       "estimated": false
                     }


ryanhofdotgov:
@hursey013 Can you confirm that if you fill out that form section in a blank account that eApp, as-is, will generate the JSON as you expect? i.e., confirm that there isn't a code-level issue here?

hursey013:
@ryanhofdotgov yes I can confirm that by using the UI it sets `day` to `1` on date fields that have the `hideday` property set (so basically where it just needs month/year), which is actually quite prevalent, it's being used on 21 different date fields. My guess is since the day is not represented in the final XML its just being collected to form a complete date for validation purposes and then just getting dropped off in the end.

We could perhaps modify the date validation logic to take into account when it we only have a month/year, but it looks like the current pattern is to just collect the day in a hidden field.

ryanhofdotgov:
FYI, I couldn't find any other examples of this odd data scenario

PR #857, opened 2018-09-24
Frontend fixes for test case 5

hursey013:
Assortment of frontend validation bugs that were leading to phantom errors when loading test case 5. Related to https://github.com/18F/e-QIP-prototype/issues/824

### DIFF
--- a/api/testdata/complete-scenarios/test5.json
+++ b/api/testdata/complete-scenarios/test5.json
@@ -16654,7 +16654,7 @@
                   "DivorceLocation": {
                     "type": "location",
                     "props": {
-                      "layout": "US City, State, International city",
+                      "layout": "US City, State, Zipcode International city",
                       "city": "Monterey",
                       "state": "CA",
                       "zipcode": "93940",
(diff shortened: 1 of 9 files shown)


api/testdata/complete-scenarios/test5.json
```diff
@@ -16654,7 +16654,7 @@
                   "DivorceLocation": {
                     "type": "location",
                     "props": {
-                      "layout": "US City, State, International city",
+                      "layout": "US City, State, Zipcode International city",
```
ryanhofdotgov: This test data change originated back with https://github.com/18F/e-QIP-prototype/commit/c19617bebff7d29e63f9481c1227babef1a08683 and https://github.com/18F/e-QIP-prototype/pull/833

In that PR some other test data sets were created with the same layout (`"US City, State, International city"`) – do they need to be updated as well with the `Zipcode` variant?

hursey013: Yeah, good catch. Updated those references as well.

PR #862, opened 2018-09-25
Add first, last, method, frequency to all non-citizen relatives contact

### LINKED ISSUE #848: Fails to ask all required questions for relatives
eApp fails to ask all the required questions for relatives that are not a U.S. Citizen, but have a U.S. address and are not deceased. See highlighted screenshot and father-in-law in `test7.json`.

screen shot 2018-09-23 at 10 11 27 pm

eApp does not ask:
* first contact
* last contact
* methods of contact
* frequency of contact

apburnes:
Closes #848

The SF86 form asks for first contact date, last contact date, contact frequency, and contact method for all non US Citizen relatives that live both in the country and out of the country.

Previously the form only asked for these inputs for relatives living outside the country.  Now all non US Citizen relatives will have these inputs shown.


apburnes:
On form appearances alone, they read like first, last, frequency, and method are both required to me. The only difference is the US documentation allowing them to live within the US.  The lack of the oxford comma makes it confusing though.

__US Address__
screen shot 2018-09-25 at 8 17 05 am

__Foreign Address__
screen shot 2018-09-25 at 8 17 29 am

ryanhofdotgov:
Yeah, for each section, the sub-conditions listed in the grey at the top should be considered logically AND'ed sub-conditions – do this if X AND Y AND Z. 

But as Andrew points out, sometimes there is an artificial split of conditions (i.e., they didn't optimize the size and quantity of boolean expressions :P) due to the physical design of the paper form.

ryanhofdotgov:
Using `test7.json` and updatig the FIL entry, I ran into a few issues:
* Methods of contact is a `check all that apply` in the PDF, but can only choose one option
* Selecting Frequency of contact --> Other, typing in explanation required, then saving, logging out, signing back in, shows that the explanation is not persisting to database (or not making the round-trip somehow)
* Selecting Frequency of contact --> Other, but leaving the required explanation blank does not result in expected validation error 

(Not sure if these problems are isolated to this particular instance, or these are latent issues where in other places where these questions are asked)

ryanhofdotgov:
Also, when changing frequency of contact to an option that doesn't require an explanation, the explanation text area does not get cleared and hidden automatically, like the method of contact input.

apburnes:
_Some additional notes_

When reviewing the `test7` errors.  It looks like the tests data cannot be update do to `duplicate key` errors. 

When creating another relative along side the test data, it updates the `method` and `frequency` types but will keep the `Other` type's text if it is not manually deleted from the input. Also, the `Other` type's text box will function as expected the first time inputing the text, but if it is retrieving from the database, the `Other` type's text input sticks after choosing another option and will continue to retrieve the `Other` text from the database even if it is saved as another type ie `Telephone`.

ryanhofdotgov:
@apburnes Can you clarify your comment about the duplicate key errors? Is that from a clean/empty database when loading in the scenario? I will admit that since duplicate key errors are generated on purpose in eApp wherever it wants to do an "upsert", I basically ignore them. It my opinion it is a bad design. 

Is this a situation where the duplicate key errors are legitimate and not noise?

apburnes:
Whenever working on updating the three preloaded relatives, the server spit out a ton of the duplicate key and foreign constraint errors.  But when adding additional relatives through the app, they were updating properly.  It's really hard to read exactly what's going on but it seems like the preloaded data may have some sort of a schema bug.

Issue #730, opened 2018-08-24
Evaluate archival PDFs design


ryanhofdotgov:
Archival PDF is just a print-out of the UI, usually in greyscale. This makes review difficult in some cases. 

Here is one flagged by a test case reviewer, regarding whether Present was actually selected or not for a date range. Also the place holder text "##" was misinterpreted as user-supplied data.

Existing teams of course currently very familiar with existing archival PDF format, which is designed for printing and has a more concise, compact layout.

<img width="300" alt="present-not-selected" src="https://user-images.githubusercontent.com/37100189/44596830-8001f880-a782-11e8-8218-b144d4f7d622.png">
<img width="300" alt="present-selected" src="https://user-images.githubusercontent.com/37100189/44596831-81332580-a782-11e8-97f5-da84685ac829.png">

Related issues:
https://github.com/18F/e-QIP-prototype/issues/701
https://github.com/18F/e-QIP-prototype/issues/373

PR #885, opened 2018-09-26
Fix margins on archival PDF, button sizes

### LINKED ISSUE #874: Fix margins on archival PDF, button sizes
**What did you do?**

Load up `test5.json` (or another test case), submit, print archival PDF. 

Margins similar to when manually printing each section review page, with consistent button sizes and button labels that aren't truncated/spaced oddly within the button.

**What did you see instead?**

Weird margins, resulting in PDF that is 100 pages longer than manual concatenation of section review print-to-PDFs. Buttons of odd size. Button labels that don't fit within the buttons. I suspect that changing margin will fix the buttons, but maybe not.

**Additional Information**

In [Archival PDF from Windows 8.1 with Chrome.pdf](https://github.com/18F/e-QIP-prototype/files/2418104/Archival.PDF.from.Windows.8.1.with.Chrome.pdf) see following examples:
page 27 (button sizes)
page 117 (clipped button label)
page 126 (giant buttons)

For comparison:
[test5b-archival.pdf](https://github.com/18F/e-QIP-prototype/files/2418106/test5b-archival.pdf)

line47:
This Fixes #874 
- The margins have been updated 
- Button sizes look good
- Fixed side margin 
- Removed green checkboxes from fields
- Fixing the spacing of accordions to allow for a wider content area

The button sizing is off because there are specific sizes that are used for certain sections for example in `EducationItem.scss` the following is used...

```CSS
.history {
  .type,
  .diploma {
    label {
      width: 25rem !important;
    }
  }
```

This width is not applied on the final print because there is no `.history` class wrapping that on the final print page like their is earlier in the application. 😢

### DIFF
--- a/src/components/Section/History/History.jsx
+++ b/src/components/Section/History/History.jsx
@@ -762,7 +762,7 @@ export class HistorySections extends React.Component {
       return false
     }
     return (
-      <div>
+      <div className="history">
         <Residence
           {...this.props.Residence}
           defaultState={false}
--- a/src/sass/print.scss
+++ b/src/sass/print.scss
@@ -1,5 +1,8 @@
 // Override the default styling of TABLE applied from
 // USWDS.
+.pre-print-view {
+  padding: 0 1rem;
+}
 .print-table {
   display: block;
   margin: 0;
@@ -269,6 +272,36 @@
     opacity: 1;
     outline: 3px solid #000;
   }
+
+/* Hiding options that were not selected and only showing the selection */
+// .option-list {
+//   .block label {
+//     display: none;
+//   }
+//   .block label.checked {
+//     display: block;
+//   }
+// }
+
+input[type='text'].usa-input-success,
+input[type='number'].usa-input-success,
+input[type='email'].usa-input-success,
+textarea.usa-input-success, select.usa-input-success {
+  background-image: none;
+}
+
+/* Adjusting accoridon spacing for print */
+.accordion {
+  padding: 0;
+  .summary .left {
+    padding-left: 0;
+    width: 100%;
+  }
+  .item .details {
+    padding: 0 1rem;
+  }
+}
+
 
   /* Get rid of left blue border within accordion */
   .accordion .summary .left.open,
--- a/src/sass/print.scss
+++ b/src/sass/print.scss
@@ -3,6 +3,7 @@
 .pre-print-view {
   padding: 0 1rem;
 }
+
 .print-table {
   display: block;
   margin: 0;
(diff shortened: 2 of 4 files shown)


line47:
Seems like I need to add a way to place the class `history` around the corresponding section on the `print.jsx` page 
https://github.com/18F/e-QIP-prototype/blob/develop/src/components/Section/Package/Print.jsx#L142
```
case 'history':
          sectionComponent = <HistorySections {...this.props.History} />
          break
```

src/components/Section/History/History.jsx
```diff
@@ -762,7 +762,7 @@ export class HistorySections extends React.Component {
       return false
     }
     return (
-      <div>
+      <div className="history">
```
line47: Added to make the buttons in that section the proper width

src/sass/print.scss
```diff
@@ -270,6 +274,35 @@
...
+  //     display: block;
+  //   }
+  // }
+
+  input[type='text'].usa-input-success,
+  input[type='number'].usa-input-success,
+  input[type='email'].usa-input-success,
+  textarea.usa-input-success, select.usa-input-success {
+    background-image: none;
+  }
```
line47: hiding the green checks ✅

src/sass/print.scss
```diff
@@ -270,6 +274,35 @@
...
+    padding: 0;
+    .summary .left {
+      padding-left: 0;
+      width: 100%;
+    }
+    .item .details {
+      padding: 0 1rem;
+    }
+  }
+
```
line47: Fixing the spacing of accordions to allow for a wider content area

src/sass/print.scss
```diff
@@ -270,6 +274,35 @@
...
 
+  /* Hiding options that were not selected and only showing the selection */
+  // .option-list {
+  //   .block label {
+  //     display: none;
+  //   }
+  //   .block label.checked {
+  //     display: block;
+  //   }
+  // }
```
line47: The above is commented out but this is how we can begin to hide unselected items from the print view 🤖

line47:
**Screenshots of buttons that are no longer issues** 
screen shot 2018-09-26 at 11 50 55 am
screen shot 2018-09-28 at 9 26 45 am
screen shot 2018-09-28 at 9 27 28 am
screen shot 2018-09-28 at 9 31 15 am
screen shot 2018-09-28 at 9 38 15 am
screen shot 2018-09-28 at 9 38 55 am

line47:
@apburnes Hiding the green checks is already happening, @ryanhofdotgov wanted to hold off on hiding the unchecked items.  I could take a look at https://github.com/18F/e-QIP-prototype/issues/730 and tackle that next though.

ryanhofdotgov:
@line47 Comparing the archival PDF on the review page with the manually created archival linked in the original ticket description, the page count is still almost 100 pages larger – 650 pages vs 565 pages for test5b-archival.pdf.

The margins in the auto-generated archival PDF seems way larger still, which I'm guessing is the primary contributing factor at this point. Can you address this? Looking for an end result where this PDF is at least as good/bad as manual approach.

Issue #594, opened 2018-07-23
ensure assets are compiled with cross-browser support


afeld:
> IE11 and recent versions of Edge, Chrome, Firefox, and Safari on desktop. For mobile, mobile Safari and Chrome for Android

https://gsa-tts.slack.com/archives/C13HJ0ZBN/p1532369774000316?thread_ts=1532369398.000002&cid=C13HJ0ZBN

Relevant documentation on polyfills etc:

* https://babeljs.io/docs/en/babel-polyfill/
* https://babeljs.io/docs/en/babel-preset-env/
* https://postcss.org/

PR #890, opened 2018-09-26
Upgrade polyfills and breakout script in index

apburnes:
Closes #890 

The app was already using polyfill support so upgraded babel polyfill and split out the script in the `index.html`.  Splitting out the polyfill script tag will allow us to conditionally pull it in if the browser does not support the features we need (_Will need to review the features we are using to take advantage of the optional require_).


ryanhofdotgov:
Closes https://github.com/18F/e-QIP-prototype/issues/594

PR #891, opened 2018-09-26
Replace OPM logo with NBIS logo

### LINKED ISSUE #887: Update logo
Use this new logo instead of the OPM one on the home page and the banner. Downscale to appropriate resolution.

https://drive.google.com/open?id=1mcPnbuFlZoOBB_2YYQXZCwSFYlnGknQ8

ryanhofdotgov:
Fixes #887


hursey013:
Putting my designer hat on really quickly, do you know if they happen to have an svg version rather than the PNG? Benefit of the svg being that it'll always render crisply on higher resolution screens (like retina displays, etc).

ryanhofdotgov:
No SVG right now. And unfortunately SVG we currently have from OPM was ~500KB and didn't work properly on IE11 (rendered black in some cases). 

And the PNG image I just used looks horrible in IE11 when scaled due to this problem:
https://github.com/sukhoi1/ie-bicubic-img-interpolation-plugin/wiki

screen shot 2018-09-26 at 1 53 29 pm

Issue #897, opened 2018-09-27
Re-check test cases 1-6 for phantom zip code errors


ryanhofdotgov:
I'm getting phantom zip code errors for test cases 2 and 3 when reloading from file, under `develop` branch. Need to recheck all of them again, and fix problem code.

When testing a file, start by going to the overall validation page. Then if that is clear, go to each section review page. I've noticed at least once (see #898) that the overall validation page is *not* a superset of the section review page validations.

- [x] test1.json
- [x] test2.json
- [x] test3.json
- [x] test4.json
- [x] test5.json
- [x] test6.json

ryanhofdotgov:
The way I get past the phantom validation errors is to complete remove text in the zip code field, tab out of field, then put in the exact same zip code back in, tab out of field.

hursey013:
Just went through everything again, including manually going to each sections `Review`. All of the errors relate to zip codes and don't appear to be phantom errors, they all have corresponding validation messages. I also observed the same behavior - if you reselect the State, click in and out of the Zip code field the error is cleared.

- Test 1: ok
- Test 2: Final review: `Substance use > Mandatory counseling or treatment`, Substance use review: `Counseling 1: Employer, Court official01/1990 - 06/1990`
- Test 3: Final review: `Relationships`, `Foreign associations`
- Test 4: Final review: `Relationships > People who know you well`
- Test 5: ok
- Test 6: ok

I'm sort of wondering if this has something to do with the test data, because if there is an issue with the zip code validation I would expect it to be more widespread, but its not happening in Test 5 for example.

hursey013:
Yeah I just saw that PR, for some reason it's not applying to lowercase values that are coming from the db.

PR #903, opened 2018-09-27
Uppercase all state abbreviations

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/897


hursey013:
This will be unneeded after https://github.com/18F/e-QIP-prototype/pull/904, but still might be worth merging to normalize the data?

PR #906, opened 2018-09-27
Set lastRefresh on initial form load

### LINKED ISSUE #895: Session timeout terminates session early
**What did you do?**

Login, accept terms, don't on anything except maybe: navigation menu expand/close, instructions. Don't click on Next or click onto a sub-section page. Wait 1-2 minutes.

No logout until the 15 minute session timeout.

**What did you see instead?**

Logout.

**Additional Information**

`tick()` in TimeoutWarning.jsx does a check against props.lastRefresh the first time it is called, which if you don't change routes by navigation to a page, will be null (evaluated to 0),  due to `defaultProps` of `TimeoutWarning` – set on initial page load. This incorrectly makes the diff >= to timeout.

To correct, I believe lastRefresh needs to be set to the current time after the first successful login.

hursey013:
Fixes #895 

The form looks for route changes to update the `lastRefresh` value (by way of `onRouteChanged()` > `refreshToken()` > `updateApplication()`: https://github.com/18F/e-QIP-prototype/blob/develop/src/views/Form/Form.jsx#L116-L127) however on initial load of the app this normal handling is skipped and instead simply redirects users to the intro page: https://github.com/18F/e-QIP-prototype/blob/develop/src/views/Form/Form.jsx#L51, never setting a value for `lastRefresh`. This uses React's `componentDidMount()` to run the `refreshToken()` method on initial render, setting an initial value for `lastRefresh`.

### DIFF
--- a/src/views/Form/Form.jsx
+++ b/src/views/Form/Form.jsx
@@ -41,6 +41,10 @@ class Form extends React.Component {
     }
   }
 
+  componentDidMount() {
+    this.refreshToken()
+  }
+
   getParams() {
     return this.props.params || this.props.match.params
   }


ryanhofdotgov:
@hursey013 Does `componentDidMount()` get called before or after the user successfully authenticates? For example, if I just open the front page and don't sign in for 20 minutes, what will the value of lastRefresh be?

hursey013:
@ryanhofdotgov it gets called after the `form` component is loaded which doesn't happen until after a successful login. From what I can see none of the timeout logic is activated on the login page itself.

Issue #653, opened 2018-08-06
Require entries for mother-in-law and father-in-law if married/civil union/etc.


ryanhofdotgov:
**What did you do?**

Indicate that I was married on form. Don't specify data for mother-in-law or father-in-law. Submit to e-QIP.

**What did you expect to see?**

No errors on submission.

**What did you see instead?**

e-QIP level error requiring mother-in-law and father-in law information.

**Additional Information**

Also correct/test civil union case, etc.

PR #757, opened 2018-09-05
Add mother and father validation

hursey013:
Fixes #630

Issue #918, opened 2018-10-02
Collapse civil marriage, civil union, etc. into one button to choose from 


ryanhofdotgov:
In eApp, when choosing marital status, civil marriage is separated from civil union, etc. as a separate option. 

In e-QIP, these choices are combined (see screenshot below). NBIB policy team has ruled that eApp needs to conform to how e-QIP/SF-86 does it. 

Be sure to inspect/address how this change affects the underlying JSON/SF-86 XML generation.

![unnamed-1](https://user-images.githubusercontent.com/37100189/46356681-134b0b00-c618-11e8-853a-93beda835381.jpg)

PR #921, opened 2018-10-02
Make mother and father-in-law required when married or separated

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/653

Integrates marriage status into relatives validator. If applicant indicates that they are `Married` or `Separated` they will be required to add a mother-in-law and father-in-law as relatives.

**After:**
screen shot 2018-10-02 at 4 05 47 pm

As with https://github.com/18F/e-QIP-prototype/pull/757 I think we will need to revisit how these collection based errors are handled so that they are all done in a consistent way.

### DIFF
--- a/src/validators/relatives.js
+++ b/src/validators/relatives.js
@@ -60,7 +60,7 @@ export default class RelativesValidator {
     const requiredRelations = ['Father', 'Mother']
 
     if ((this.list.branch || {}).value === 'No') {
-      if (this.list.items || {}.length > 0) {
+      if ((this.list.items || {}).length > 0) {
         let relations = []
         for (const item of this.list.items) {
           if (!item || !item.Item || !item.Item.Relation) {
--- a/src/validators/relatives.js
+++ b/src/validators/relatives.js
@@ -39,7 +39,7 @@ export default class RelativesValidator {
     }
 
     if ((this.list.branch || {}).value === 'No') {
-      if (this.list.items || {}.length > 0) {
+      if ((this.list.items || {}).length > 0) {
         let relations = []
         for (const item of this.list.items) {
           if (!item || !item.Item || !item.Item.Relation) {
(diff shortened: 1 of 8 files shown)


ryanhofdotgov:
@hursey013 Can you write up an issue that describes the re-factor you have in mind to handle these collection based errors better and reference these 2 related tickets? Thanks.

src/validators/relatives.js
```diff
@@ -18,11 +27,40 @@ export default class RelativesValidator {
...
+    }
+    return true
+  }
+
   validMinimumRelations() {
     const requiredRelations = ['Father', 'Mother']
 
     if ((this.list.branch || {}).value === 'No') {
-      if (this.list.items && this.list.items.length > 0) {
+      if (this.list.items || {}.length > 0) {
```
ryanhofdotgov: This should get enclosing parens before calling length?

hursey013: Yeah, updated, thanks.

src/validators/relatives.js
```diff
@@ -18,11 +27,40 @@ export default class RelativesValidator {
     })
   }
 
+  validMaritalRelations(context = null) {
+    const maritalStatuses = ['Married', 'Separated']
```
ryanhofdotgov: Just noting that while this is consistent with validation matrix, because we have been treating Married as a different thing than InCivilUnion, this would be a defect. However now that we are coalescing Married/InCivilUnion in #918, it is moot.

src/validators/relatives.js
```diff
@@ -18,11 +27,40 @@ export default class RelativesValidator {
...
+    const maritalStatus = context || getContext()
+
+    console.log(maritalStatus)
+
+    if (!maritalStatus || !maritalStatuses.includes(maritalStatus)) {
+      return true
+    }
+
+    if ((this.list.branch || {}).value === 'No') {
+      if (this.list.items || {}.length > 0) {
```
ryanhofdotgov: Another place where we need parens

src/validators/relatives.js
hursey013: Resolved with https://github.com/18F/e-QIP-prototype/pull/927

PR #927, opened 2018-10-03
Combine civil marriage and civil union into one button

### LINKED ISSUE #918: Collapse civil marriage, civil union, etc. into one button to choose from

hursey013:
Fixes #918 

Combined options and updated language to reflect what is in the current SF-86 PDF.

**Before:**
screen shot 2018-10-03 at 12 33 54 pm

**After:**
screen shot 2018-10-03 at 12 43 19 pm

### DIFF
--- a/api/relationships.go
+++ b/api/relationships.go
@@ -68,7 +68,7 @@ func (entity *RelationshipsMarital) Valid() (bool, error) {
 
 	sv := entity.Status.Value
 	switch {
-	case sv == "Separated":
+	case sv == "Married" || sv == "Separated":
 		// Check if the civil union information is valid
 		if ok, err := entity.CivilUnion.Valid(); !ok {
 			stack.Append("CitizenshipMarital", err)
--- a/api/relationships.go
+++ b/api/relationships.go
@@ -68,7 +68,7 @@ func (entity *RelationshipsMarital) Valid() (bool, error) {
 
 	sv := entity.Status.Value
 	switch {
-	case sv == "InCivilUnion" || sv == "Separated":
+	case sv == "Separated":
 		// Check if the civil union information is valid
 		if ok, err := entity.CivilUnion.Valid(); !ok {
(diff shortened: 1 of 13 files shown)


api/xml/xml.go
```diff
@@ -495,7 +495,6 @@ func maritalStatus(status string) string {
 	alias := map[string]string{
 		"NeverMarried": "NeverMarried",
 		"Married":      "Married",
-		"InCivilUnion": "Married",
```
ryanhofdotgov: Since we've removed InCivilUnion, this method can be removed – both where it is referenced in this file and in the places where it is referenced in the templates.

ryanhofdotgov: This is because the eApp vernacular (left side) now matches the e-QIP vernacular (right side) in all the cases.

hursey013: Makes sense, removed the function and references to it.

api/relationships.go
```diff
@@ -68,7 +68,7 @@ func (entity *RelationshipsMarital) Valid() (bool, error) {
 
 	sv := entity.Status.Value
 	switch {
-	case sv == "InCivilUnion" || sv == "Separated":
+	case sv == "Separated":
```
ryanhofdotgov: It seems like there is a bug in the existing code – what do you think? Should be doing something for "Married" value at first glance.

hursey013: Yeah on second look I think you're right, I'll add `Married` in since I don't see it being evaluated anywhere else in that switch statement.

PR #929, opened 2018-10-03
Date validation audit 

### LINKED ISSUE #683: Alias date range can't start at your birthday
**What did you do?**

Indicate that applicant was known by another name beginning with their birthday. 

No error.

**What did you see instead?**

Error that says alias date range can't be the same as your birthdate.

**Additional Information**

Seems like this would happen frequently, born under one name,  then it got changed at a later date?

### LINKED ISSUE #836: In dual citizenship date range, from date should be able to be birthday
**What did you do?**

Tried to add a new citizenship entry in dual citizenship section. Said start date was birthdate.

No errors.

**What did you see instead?**

Error saying date has to be after birth date.

**Additional Information**

Should be greater-than-or-equal to test, not greater-than.

### LINKED ISSUE #913: Audit all places where dates are compared as part of validation
Audit all places where dates are compared as part of validation and ensure that the correct comparison operation is being used.

For example, it has been a pervasive issue that this type of check is done:
date-XYZ > birthdate
when it should  be:
date-XYZ >= birthdate

Use the validation matrix PDF as a guide. You can search for "greater than" or ">=" or ">" or "less than" within the PDF to help find all the date validation.

We recently added an option to the date field to support greater than. Refer to https://github.com/18F/e-QIP-prototype/pull/772 to see how to make similar changes on other parts of the form.

### LINKED ISSUE #976: Residence item errors when the "from" date is equal to applicant's birthday
**What did you do?**

1. Add birthday
2. Add an address where you have lives
3. Set start of residence to the same date as the birthday

No Errors

**What did you see instead?**

Error that the `from` date should be after your date of birth.

**Additional Information**

`from` date should be able to be equal to the user's birthday.
I suspect the solution should be to go into `ResidenceItem.jsx` and find the corresponding `<DateRange />` component and add a `minDateEqualTo` prop to make it true.

line47:
Updating date validation in most cases to allow for a `>=` 

This fixes #913 
This fixes #683
This fixes #836 
This fixes #976


line47:
There is more work to be done for the date validation that is more complex and needs to compare more than 2 dates. I will open a separate issue for that work as to not continue to add to this PR. 👍

src/components/Form/DateControl/DateControl.jsx
```diff
@@ -269,10 +269,30 @@ export default class DateControl extends ValidationElement {
       let props = null
       if (date && !existingErr) {
         // Prepare some properties for the error testing
+        const buildValidators = (props) => {
```
el-mapache: Since its going to require more work than we originally thought to handle validation against multiple dates, should we revert this section to what we currently have in the develop branch?

src/components/Form/DateControl/DateControl.jsx
```diff
@@ -269,10 +269,30 @@ export default class DateControl extends ValidationElement {
...
+          }
+          
+          return validators
+        };
+
         props = {
           ...this.props,
           ...this.state,
-          validator: new DateControlValidator({ ...this.props, ...this.state })
+          validator: buildValidators(this.props)
```
el-mapache: see above!

src/components/Section/Citizenship/Multiple/CitizenshipItem.jsx
```diff
@@ -116,7 +116,9 @@ export default class CitizenshipItem extends ValidationElement {
             name="Dates"
             {...this.props.Dates}
             className="citizenship-dates"
-            minDateEqualTo
+            maxDate={(this.props.minDate || {}).date}
+            maxDateEqualTo
+            minDateEqualTo={true}
```
el-mapache: I think for consistency's sake we should either omit the `={true}` from the places it's used or use it everywhere

src/components/Section/Foreign/Activities/BenefitActivity/FutureBenefit.jsx
```diff
@@ -114,6 +114,8 @@ export default class FutureBenefit extends ValidationElement {
             className="begin"
             {...this.props.Begin}
             label={i18n.t('foreign.activities.benefit.future.label.begin')}
+            minDate={(Date.now || {}).date}
```
el-mapache: I don't think this is going to work correctly. `Date.now` is a function and won't have a `date` property. Additionally, the output of `Date.now()` is a value in milliseconds (representing time elapsed since Jan 1 1970).

There is probably an existing helper function in the code that will construct a date object that the app understands from `new Date()`

src/components/Section/Relationships/RelationshipStatus/CivilUnion.jsx
```diff
@@ -355,6 +355,8 @@ export default class CivilUnion extends ValidationElement {
             <DateControl
               name="enteredCivilUnion"
               className="entered"
+              // minDate={[this.props.Birthdate, this.props.ApplicantBirthdate]}
```
el-mapache: We should just remove this commented out code

src/components/Section/Citizenship/Multiple/TravelItem.jsx
```diff
@@ -56,6 +56,10 @@ export default class TravelItem extends ValidationElement {
             name="Dates"
             className="travel-item-dates"
             {...this.props.Dates}
+            minDate={(this.props.Birthdate || {}).date}
+            minDateEqualTo={true}
+            maxDate={(this.props.Birthdate || {}).date}
+            maxDateEqualTo
```
line47: To review

src/components/Section/Citizenship/Multiple/CitizenshipItem.jsx
line47: Yep, agreed!

ryanhofdotgov:
Per review with @line47 earlier today, we're going to take a few actions:
* I'm going to cherry pick low-risk changes that address specific problems (i.e., changes that use the existing `minDateEqualTo` property, etc.) into a different branch and merge that
* The allowEqualTo property does not seem necessary – changing comparisions operations to use <=, >= in the validator and component as you always want to do an equal-based comparison (that we can tell).
* There is some redundant logic in the component and validator for DateRange. Follow on work will figure out how to have it one place.
* Follow on work will address situations where you need to set min date or max date to the min/max of multiple dates.
* Take another pass through and verify newly added constraints make sense

src/components/Section/Foreign/Activities/BenefitActivity/FutureBenefit.jsx
```diff
@@ -114,6 +114,8 @@ export default class FutureBenefit extends ValidationElement {
             className="begin"
             {...this.props.Begin}
             label={i18n.t('foreign.activities.benefit.future.label.begin')}
+            minDate={new Date()}
```
line47: This works for allowing futures dates but the minDateEqual to property is not working as expected here. 
I can set to 1 day in the future but can't set to today's date. @ryanhofdotgov

ryanhofdotgov: Yeah, if it's not working as expected with minDateEqual we should pull it out for now and do it separately.

ryanhofdotgov:
@line47 I'm seeing places in this PR where:
* it's setting minDateEqualTo without minDate set. Does that make sense? I thought minDateEqualTo required minDate to do anything.
* `minDate={applicantBirthdate}` or `minDate={this.props.ApplicantBirthdate}` was removed from a form component  – are those intentional?

Also, have you found any instances in the form where the comparisions operated should not use equals?

line47:
Thanks for taking another look @ryanhofdotgov 😄 
- I have found that setting `minDateEqualTo` compares to applicant DOB by default unless otherwise specified with using `minDate` explicitly like `minDate={(this.props.Issued || {}).date}` so I don't think a minDate is needed unless overriding that default behavior.
- Removed `minDate={applicantBirthdate}` or `minDate={this.props.ApplicantBirthdate}` as I found that those were not needed. See reasoning in first bullet point.
- I have not found any places in the form where comparisons would not use equals. Most dates are `>= ` and applicant date or some other date based on context.

Issue #915, opened 2018-10-02
ExistingConditions entries must not ask "Identify the diagnosis or health condition"


ryanhofdotgov:
**What did you do?**

Said "Yes" to existing conditions in Section 10. Filled out a specific treatment entry.

**What did you expect to see?**

No question asking "Identify the diagnosis or health condition", since it is not on the SF-86 nor is there a place for it in e-QIP XSD.

**What did you see instead?**

The question "Identify the diagnosis or health condition", which gets persisted in `TreatmentList.props.items.Item.Condition`

**Additional Information**

This was probably done by accidental oversight, rather than new design, believing that this part of the form was consistent with the other places where we solicit treatment information.

Update `test8.json` and any other JSON test cases to reflect change when completed.

PR #930, opened 2018-10-03
Remove question from existing conditions

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/915

Removes the question `Identify the diagnosis or health condition` asked within Existing Conditions, as it does not exist on the SF-86.

**Before:**
screen shot 2018-10-03 at 4 46 35 pm

**After:**
screen shot 2018-10-03 at 4 43 22 pm

### DIFF
--- a/src/components/Section/Psychological/Diagnoses/Diagnosis.jsx
+++ b/src/components/Section/Psychological/Diagnoses/Diagnosis.jsx
@@ -79,20 +79,10 @@ export default class Diagnosis extends ValidationElement {
     const applicantBirthdate = getContext().applicantBirthdate
     return (
       <div className="diagnosis">
-        <Field
-          title={i18n.t(`psychological.${prefix}.heading.condition`)}
-          scrollIntoView={this.props.scrollIntoView}>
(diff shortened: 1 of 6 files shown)


src/components/Section/Psychological/Diagnoses/Diagnosis.jsx
```diff
@@ -79,20 +79,10 @@ export default class Diagnosis extends ValidationElement {
...
-              name="Condition"
-              className="diagnosis-condition"
-              {...this.props.Condition}
-              onUpdate={this.updateCondition}
-              onError={this.props.onError}
-              required={this.props.required}
-            />
-          </Show>
-          <Show when={this.props.prefix === 'diagnosis'}>
+        <Show when={this.props.prefix !== 'existingConditions.diagnosis'}>
```
ryanhofdotgov: This is a pervasive pattern across the app; i.e. conditional behavior within a re-usable component based on location of component instance, but I think we will need to revisit as we start implementing different forms. It should probably be another non-path related property that gets set on instantiation.

PR #931, opened 2018-10-03
Update State component to save abbreviation rather typed input

### LINKED ISSUE #716: Normalize case of state abbreviations
UI will accept and validate state abbreviations of mixed case; i.e., 'Or' instead of 'OR', 'Ca' instead of 'CA'. Accept mixed case input, but store uppercase version.

christopherhuii:
Fixes #716 

In this PR:

- Extracted out the state children from the render and concatenate the internal states and the children states.
- When an applicant types in the full state name, the component swaps it out and sends the state abbreviation in the payload.
- When an applicant types an invalid state name, an error will show up describing the error.

I'm not 100% sure if the behavior of this component is satisfactory. It might help to have a second pair of eyes on it.


ryanhofdotgov:
@christopherhuii Can you confirm that this actually persists the uppercase version of the state abbreviation? Displaying in uppercase was OK (so I think we want to keep the CSS), but in the serialized JSON that was saved to DB (and in-memory representation) the abbreviation was stored in mixed-case.

You can see what is being stored by inspecting the save network call using the chrome developer tools and looking through the payload.

christopherhuii:
@ryanhofdotgov Got it, I misunderstood the issue. Thanks for the clarification

ryanhofdotgov:
@christopherhuii Is this ready for review again (didn't jump in earlier since it was still labelled WIP)

christopherhuii:
@ryanhofdotgov Yeah, this might be a good place to review it again and get feedback. I'm not 100% sure if the behavior is acceptable, so it might be worth another set of eyes to review.

ryanhofdotgov:
screen shot 2018-10-09 at 9 47 41 pm

src/components/Form/State/State.jsx
```diff
@@ -21,6 +119,17 @@ export default class State extends ValidationElement {
     return this.props.onError(value, arr)
   }
 
+  handleUpdate(stateObj) {
+    const value = stateObj.value.length > 2
+      ? this.getStateAbbreviation(stateObj.value)
```
el-mapache: I think here I would probably just do a `find` directly on `this.states`. 
You should be able to write something like this, pretty much the same code you already have:

`this.states.find((state) => state.name.toLowerCase() === stateName.toLowerCase())`

src/components/Form/State/State.jsx
```diff
@@ -5,7 +5,105 @@ import Dropdown from '../Dropdown'
 export default class State extends ValidationElement {
   constructor(props) {
     super(props)
+    this.getStateAbbreviation = this.getStateAbbreviation.bind(this)
+    this.getStates = this.getStates.bind(this)
     this.handleError = this.handleError.bind(this)
+    this.handleUpdate = this.handleUpdate.bind(this)
+
+    this.states = [
```
el-mapache: This could probably be a constant that lives outside this component, depending how many other places we use an array of states. We definitely define a couple of ways to interact with states and their abbreviations in `src/validators/location.js`

christopherhuii: Oh interesting file. Yeah it looks like there are a couple of different ways states are defined.

I would prefer not to duplicate the list of states/territories again in `src/validators/location.js`. So it looks like the options are to alter the `location.js` file in a way where it supports all the current use cases, or leave it as it.

In `States.jsx`, it does not have `U.S Armed Forces - Americas`, `U.S. Forces - Europe`, `U.S. Armed Forces - Pacific`. The ONLY situation where  the `States.jsx` is used, is in `MilitaryStates.jsx`, where it adds the 3 Armed Forces as `this.props.children`. It's a little strange..

src/components/Form/State/State.jsx
christopherhuii: The problem I found doing a `find` directly on `this.states` is that it doesn't take into account the additional options that can be passed through `this.props.children`.

The `MilitaryStates.jsx` component leverages the State component (`State.jsx`) and adds the Armed Forces as children.

src/components/Form/State/State.jsx
el-mapache: I'm not sure how you feel about it, but I'd also be fine just adding those military states...I get that they wan't to make it flexible by defining a way for `MilitaryState` to decorate it, but as you noted, the only state is `MilitaryState`. Since we only care about the concept of 'states'  within the US (as far as this form is concerned), it seems unnecessary to have that separation.

I'm also not trying to add a ton of work for you! Just thinking about how we can improve the app so we (hopefully) don't need to have these discussions ad nauseam.

ryanhofdotgov: I can't tell if we are all on the same page after re-reading both your comments, but just chiming in to say that by policy the SF-86 has to indicate one of the two lists of US "state" abbreviations, depending on the specific form question. 

Some of the form questions permit the APO/FPO military states (AA, AP, AE) plus all the non-military "states" (the 50 states, plus US territories/commonwealths/possessions plus D.C. and Palau). Other questions are just supposed to permit the non-military states.

christopherhuii:
I have updated my PR. Hopefully this is an acceptable solution.

I moved the states from `State.jsx` to `src/validators/location.js`. The states are broken up into 3 separate arrays:

1. United States
2. Other US Territories
3. Military States

We can pull these in depending on which states/territories are needed.

I have updated `State.jsx` to take in a prop called `additionalStates`. There is a `propType` defined for it and it should follow a specific shape. In addition to that, `MilitaryStates.jsx` leverages `State.jsx`, so I have updated the `<State />` component to have `additionalStates` (military states).

Feedback is welcome!

ryanhofdotgov:
@christopherhuii Thanks – have you been able to find different portions in eApp that require different state lists and see that they are pulled in correctly?

ryanhofdotgov:
@christopherhuii It looks like with this PR I can now enter military states like AA in the US address section. 

Here's what I would expect (this is from master branch):
screen shot 2018-10-16 at 8 34 05 am

Here is from this PR:
screen shot 2018-10-16 at 8 34 14 am

christopherhuii:
@ryanhofdotgov Should you not be able to do that? When I was looking at the code, the `<MilitaryStates />` component has military states, but they weren't showing up in the component. So I fixed that "bug". This should not allow military states?

screen shot 2018-10-16 at 8 44 53 am

christopherhuii:
@ryanhofdotgov @hursey013 Is any place in the application you can think of where the applicant needs the options to be United States/Territories, AND Military States.

So far, I've found most fields should allow the options to be one or the other
  1. United States/Territories
  2. Military States (AA, AE, AP)

I haven't seen a situation where an applicant should have both.

ryanhofdotgov:
@christopherhuii After review, I think that is correct. We've restructured eApp to never have them in the same drop-down.

The way you can verify for yourself is to pull up the XSD, look for MilitaryStates, NoMilitaryStates, AllStates. In all the spots where the schema specifies AllStates, we've split that up visually in the form, so the drop-down never has military and non-military states in a single list.

I'm going to throw on a WIP prefix to the PR title and if it is ready to go, just remove it again. Thanks.

ryanhofdotgov:
@christopherhuii One last thing, I promise – can you address the conflict that I think just recently arose do to one of the other PRs I just merged? Thanks.

PR #933, opened 2018-10-04
Show simulate valid form link only on dev

### LINKED ISSUE #273: Remove `Simulate valid form` option on `Review`page
At the very bottom of the `Review` page in the `Review and submit` section, there is a hyperlink `Simulate valid form` that gives the user access to a debugging/testing feature. This feature allows the submit process to be exercised with invalid form data. 

Remove the hyperlink and disable the associated functionality unless the system is in `debug` mode.

hursey013:
Fixes #273 

Adds a `IsDevelopment` method to `env` to show the `Simulate valid form` only on development environments. There is no associated debugging functionality to remove, the link just sets the path to the submit view.

### DIFF
--- a/src/components/Section/Package/Package.jsx
+++ b/src/components/Section/Package/Package.jsx
@@ -113,8 +113,9 @@ class Package extends SectionElement {
       if (mark.errors > 0) {
         return false
       }
-      return true
     }
+
+    return true
   }
 
   errorCheck() {
--- a/src/components/Section/Package/Package.jsx
+++ b/src/components/Section/Package/Package.jsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import { connect } from 'react-redux'
 import { withRouter } from 'react-router'
-import { i18n, navigationWalker } from '../../../config'
+import { i18n, env, navigationWalker } from '../../../config'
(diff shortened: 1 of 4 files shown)


hursey013:
With the current configuration, as seen with the debugging link, navigating to `form/package/submit` bypasses all of the error checking on the final review. Even though the debugging link is only present in `development` I've added some additional checks to `form/package/submit` to make sure the route is inaccessible in production if the form is invalid.

src/components/Section/Package/Package.jsx
```diff
@@ -82,28 +81,40 @@ class Package extends SectionElement {
...
+        return true
+      }
+    }
+
+    for (const sectionName in tally) {
+      const mark = tally[sectionName]
+      if (mark.errors > 0) {
+        return false
+      }
+      return true
```
el-mapache: I think this will return true the first time a `mark` doesn't have any errors, is that the desired behavior?

hursey013: Nope. Updated https://github.com/18F/e-QIP-prototype/pull/933/commits/27173e6961d903fc24b91df035fe61975a789a98

Issue #860, opened 2018-09-25
Additional periods of employment activity can't indicate `Present` in `To` range.


ryanhofdotgov:
When listing "Additional periods of activity with this employer", the From/To date range field set is not permitted to have a "Present" option. It's not permitted the corresponding e-QIP XSD.

For example, see Employer 4 in `test7.json`. 

<img width="400" alt="screen shot 2018-09-24 at 5 43 00 pm" src="https://user-images.githubusercontent.com/37100189/45986641-afc24b80-c021-11e8-8889-4d1634e3e40c.png">

<img width="400" alt="screen shot 2018-09-24 at 5 43 10 pm" src="https://user-images.githubusercontent.com/37100189/45986690-f44de700-c021-11e8-85fd-f23fad5bd160.png">

PR #935, opened 2018-10-04
Allow present button to be disabled on date range component

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/860

Adds new prop to toggle the `Present` button on date ranges, specifically on additional periods of employment.

**Before:**
screen shot 2018-10-04 at 3 34 01 pm

**After:**
screen shot 2018-10-04 at 3 32 49 pm

### DIFF
--- a/src/components/Form/DateRange/DateRange.jsx
+++ b/src/components/Form/DateRange/DateRange.jsx
@@ -4,6 +4,7 @@ import DateControl from '../DateControl'
 import Checkbox from '../Checkbox'
 import Svg from '../Svg'
 import { now } from '../../Section/History/dateranges'
+import Show from '../Show'
 
 export default class DateRange extends ValidationElement {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


src/components/Form/DateRange/DateRange.jsx
```diff
@@ -290,6 +293,7 @@ DateRange.defaultProps = {
   minDateEqualTo: false,
   maxDate: new Date(),
   maxDateEqualTo: false,
+  showPresent: true,
```
el-mapache: An annoying naming comment:  I think `allowPresent` might be a little more semantic....just a personal preference though

hursey013: Yeah sort of blindly following the existing convention of using `show`, but no reason not to slowly start making those more semantic. Updated https://github.com/18F/e-QIP-prototype/pull/935/commits/c259479af0f37ae5f651b09c1dc307a11ddf61b1

src/components/Section/History/Employment/AdditionalActivity.jsx
```diff
@@ -89,6 +89,7 @@ export default class AdditionalActivity extends ValidationElement {
               <DateRange
                 name="DatesEmployed"
                 bind={true}
+                showPresent={false}
```
ryanhofdotgov: Need to change this to the new name

Issue #870, opened 2018-09-25
Explanation for  U.S. residence document doesn't persist 


ryanhofdotgov:
For this question:
Provide type of documentation he or she possesses to support U.S. residence:
If you answer "Other" and put an explanation, the explanation in the text area does not round-trip (i.e., sign in, enter data, sign out, sign-in --> data is no longer there).

Test case in screenshot is `test7.json`:

<img width="400" alt="screen shot 2018-09-25 at 1 09 48 pm" src="https://user-images.githubusercontent.com/37100189/46040222-91f9f280-c0c4-11e8-8691-1bbe0b2d56b4.png">

PR #936, opened 2018-10-04
Persist other document field for US residence

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/870

It appears two different key names were used for this particular field, `OtherDocument` on the frontend and `DocumentComments` in the backend. This standardizes the name to `DocumentComments` (which was what was already being used in the backend and within the test data).


hursey013:
No I do not believe so, although to be honest I have not dug into how the reducers are currently configured.

PR #945, opened 2018-10-05
Run react tests in parallel on CI

hursey013:
Uses CircleCI's parallelism to run react tests in parallel. Not mind blowing, but decreases test completion time down to around 7 minutes instead of 12-13 minutes. Example: https://circleci.com/gh/18F/e-QIP-prototype/1714

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ version: 2
 jobs:
   test:
     machine: true
+    parallelism: 4
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


hursey013:
@el-mapache also enables using `make test-react FILES=/path/to/specific/test` possible, running the test within docker vs locally.

el-mapache:
@hursey013 That is what I couldn't get working in the code I showed you yesterday, I don't know what the actual path to file is within the container

hursey013:
@el-mapache oh it looks like its referencing the root project directory, so I've been using something like `make test-react FILES="src/actions/ApplicationActions.test.js src/actions/AuthActions.test.js"` and it's been working.

el-mapache:
@hursey013 Yep that definitely works, I feel like I tried that for sure yesterday, but there must have been some other issue with my syntax. Regardless, this is a great addition!

Issue #755, opened 2018-08-30
Offense locations must not collect street address


ryanhofdotgov:
Offense locations must not collect street addresses, only City/County/ZipCode/State if US, City/Country if outside the US. 

Ditto for offense court locations, offense law enforcement agency.

See original SF-86 form, validation matrix, XSD. Currently worked around in e-QIP XML generation by forcing address layout in template.

Remove work-around in XML templates, template helper functions`locationOverideLayout` when UI fix is in place.

el-mapache:
@ryanhofdotgov Do we still want to perform USPS address validation (can we?) with just the city/country/state/zip?

ryanhofdotgov:
No USPS validation required/desired, but the front-end "native" US state/zip code validation should still get triggered.

el-mapache:
@ryanhofdotgov gotcha, those are still validating as normal. Thanks!

PR #947, opened 2018-10-05
Refactors address component

el-mapache:
First pass at refactoring the address component. Part of #755. The component needs to be cleaned up before we can start thinking about dynamically generating the correct fields within it.

* Removes redundant field blur, update, and focus functions
* Removes extraneous divs

**note**
We still have an `updateCountry` method. Since that value is complicated, and has additional metadata fields, I'm not sure it can be made generic in the same way, unless we somehow separate out the country and comment fields


src/components/Form/Location/Address.jsx
```diff
@@ -85,49 +71,12 @@ export default class Address extends ValidationElement {
...
-      6000
-    )
-  }
-
-  updateState(values) {
-    this.update(
-      {
-        state: values.value
-      },
-      6000
```
el-mapache: I removed all of the delay values. I'm not sure what the reasoning behind them is, or why they used the magic number of 6000 ms

ryanhofdotgov: Is it connected to giving folks an opportunity to finish typing before kicking off validation logic or the web service call to USPS validation API? I don't have anything at my finger tips, but I seem to recall seeing the usability studies where they mention they were tweaking these delay values.

ryanhofdotgov: USPS validation is not enabled by default, here are some instructions:
https://github.com/18F/e-QIP-prototype/blob/develop/docs/usps.md

src/components/Form/Location/Address.jsx
```diff
@@ -169,17 +117,9 @@ export default class Address extends ValidationElement {
         break
     }
 
-    // Clear the city when moving *from* APO/FPO and *to* something else.
-    if (
```
el-mapache: I took out this block of code, it seems like if we are moving from an international address to a domestic one, or vice versa, we would definitely want the city field to be cleared? I understand there are overlaps like Paris (texas, france) or toledo (spain, ohio), but I'm not sure its worth it complicate the code base

src/components/Form/Location/Address.jsx
el-mapache: got it! In that case, a long delay is probably fine. I'll try adding it in, and make a constant to represent the delay time if needed

el-mapache: I added the delay back. and there was no different in how fast the modal appeared. I believe that behavior is caused by another call to update, and is a more reasonable 500ms

src/components/Form/Location/Address.jsx
el-mapache: Love to get some context on this, if anyone knows!

ryanhofdotgov: APO addresses don't really have a city, but eApp uses the city field to store `APO`, `FPO`, `DPO`. Here's an example from `test7.json`:
```
 "Address": {
                    "type": "location",
                    "props": {
                      "layout": "Address",
                      "street": "No American can study",
                      "city": "APO",
                      "state": "AA",
                      "zipcode": "34003",
                      "country": "POSTOFFICE"
                    }
```
So if you are switching from APO to something else, if you don't clear city, you'll get something odd looking (e.g,. 'APO' listed in the city field).  Without looking at the code carefully, I'm not sure why you don't have to clear the country field for the transition from APO to an international address though.

Speaking generally, one of the pervasive problems across eApp is that it doesn't manage state very well as the applicant optionally enables/disables different options on the form. 

This isn't an existing bug (I think) but just an example:
If I say that I'm married, that will trigger a bunch of fields to be displayed. If I fill out that data, then later switch to saying I've never been married, those fields will get hidden, but that marriage data lingers around, gets saved to the database, re-loaded, etc.

I think this is pretty sloppy and complicates downstream logic a fair bit. If you are toggling between options (in general), all dependent state associated with that option should be cleared. 

However, to play devil's advocate, there might be some sections in the form where there might be a usability case to be made that preserving data allows the user to recover from an errant click.

el-mapache: Interesting...I'd like to hear from a more UX or design-centric person (paging @hursey013 and @line47)! I do think it's a valid point, but since the country and state were always getting changed as well, with only the city being preserved, I would err on the side of just blowing it all away.

We don't actually preserve the country value currently, its set to `''` and then reassigned if necessary via a `switch` statement

line47: I think allowing for a user to recover from unintentional click is a good idea. I'm also wondering about clearing options when a user navigates away from that page instead of when they toggle between different options.

I think it's odd that `APO` is stored in `city`. 
Why not just have an `APO/FPO/DPO` field instead of using city?

I have a few UI ideas for address updates as well that Ben and I worked on previously. I can add a new issue to show these designs if that would also be helpful.

el-mapache: I think it's set up that way because when sending mail to a military address, the `APO|FPO|DPO` replaces the city

src/components/Form/Location/Address.jsx
```diff
@@ -47,27 +47,19 @@ export default class Address extends ValidationElement {
...
   }
 
   componentDidMount() {
-    // Turn on scroll detection
-    window.addEventListener('scroll', this.scroller)
+    /**
+     * On scroll, if the user has filled out the address fields AND the data has not been
+     * validated, open a USPS address verification modal
+     */
+    window.addEventListener('scroll', this.blurForceUpdate)
```
ryanhofdotgov: @el-mapache So this is unexpected behavior... if I have entered in an address that USPS doesn't validate (or validation is offline for a weird reason) and click on "Manually correct address", every time I scroll down/up on the page, the validation api gets kicked off.

What were you aiming for here?

el-mapache: Hmm, two issues here:

1) I don't think you should see any address validation without the streets. Is there old data in your form?

2). This was actually the behavior that was in the original code, but there was a bug. What behavior do we want to see with regards to the validation modal appearing (if it is still supposed to)

ryanhofdotgov: Re: #&#xfeff;1 - Yeah, that's working as expected, no validation happening until address all filled out. 

So the only thing that was unexpected was the validation API/dialog showing on scroll.

Steps to reproduce:
1) enter in a fake US address (street, city, state, zip)
2) Validation comes up – it's not found by USPS – click on "manually check address" link on bottom right corner of modal
3) Scroll up (or scroll down) – validation API gets called, modal appears on every scroll.

If this was expected behavior previously and it was broken somehow, it's probably been broken for a long time. Not sure of a use case where scrolling alone should ever cause validation, especially one that invokes a multi-second external API call. Can you alter so it doesn't do that any more, restoring the behavior on `develop` and `master`?

el-mapache: Yep, I'll put in a new PR against develop to remove the scrolling validations, and just keep them on blur.

I _think_ the use case was this:

* user fills in a complete address
* address is validated, and is found to be innacurate
* user saves or something, leaves the form without correcting
* user comes back to the form, validations retrigger to alert the user of an error.

Although, the form should get validated on page load...

el-mapache: ahh @ryanhofdotgov looking back at the code, it looks like the code was trying to validate when the user attempted to scroll away from the form, but hasnt blurred out of a field. This seems unnecessary, since the form would still be validated when the user attempted to interact with another part of the form

ryanhofdotgov: Cool – thanks!

PR #949, opened 2018-10-08
add minDate and minDateEqualTo props to CitizenshipItem's DateRange component

### LINKED ISSUE #836: In dual citizenship date range, from date should be able to be birthday

christopherhuii:
Fixes #836 

I added the `minDate` prop with the applicant's birthday. In addition to that, I added `minDateEqualTo` , so that it allows the the citizenship beginning date to be equal to the applicant's birthday.

### DIFF
--- a/src/components/Section/Citizenship/Multiple/CitizenshipItem.jsx
+++ b/src/components/Section/Citizenship/Multiple/CitizenshipItem.jsx
@@ -85,8 +85,6 @@ export default class CitizenshipItem extends ValidationElement {
     const from = d.from || {}
     const showCurrentQuestion = to.date && from.date && !d.present
 
-    const applicantBirthday = new Date(`${from.month}/${from.day}/${from.year}`)
-
     return (
       <div className="citizenship-item">
         <Field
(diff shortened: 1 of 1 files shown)


christopherhuii:
@hursey013 I see what you're saying.

However, it doesn't look like it's using that `minDate` prop. It's only using the `from` object inside of `this.props.Dates`.

hursey013:
I _think_ that's to be expected, so if someone passes in a `minDate` the validator will use that as the minimum date, but if it's not passed in the applicant's birthdate is used https://github.com/18F/e-QIP-prototype/blob/develop/src/validators/datecontrol.js#L37-L39. Definitely a little convoluted, but that's my understanding of how it works.

christopherhuii:
Thanks for the clarification. I removed the `minDate` prop because it isn't needed. The only change is the addition of the `minDateEqualTo` prop.

Issue #658, opened 2018-08-08
experiment with frontend type checking


afeld:
We run into a number of subtle bugs due to mismatched expectations in various parts of the code around the contents of objects. Investigate / experiment with using something like [TypeScript](https://www.typescriptlang.org/) or [Flow](https://flow.org/).

cc https://github.com/18F/frontend/issues/171

PR #960, opened 2018-10-09
History/Residence: Display explanation field only when "Other" is selected

### LINKED ISSUE #576: Choosing "military housing" should not require explanation
**What did you do?**

Fill out a residence, specify it as military housing.

Not to have to justify military housing.

**What did you see instead?**

Required field with heading "Please provide an explanation".

**Additional Information**

screen shot 2018-07-17 at 10 49 49 am

christopherhuii:
Fixes #576 

This PR also simplifies the logic that displays the explanation field. If `this.props.Role` is truthy, it will check its `value`. If `value` is equal to "Other", then it will display the explanation field. The explanation field ONLY shows up when "Other" is selected.

The `|| {}` and `includes` seems unnecessary.

screen shot 2018-10-09 at 12 26 02 pm

screen shot 2018-10-09 at 12 26 38 pm

### DIFF
--- a/src/components/Section/History/Residence/ResidenceItem.jsx
+++ b/src/components/Section/History/Residence/ResidenceItem.jsx
@@ -332,8 +332,7 @@ export default class ResidenceItem extends ValidationElement {
         </Field>
         <Show
           when={
-            (this.props.Role || {}).value &&
-            !['Own', 'Rent', 'Military'].includes((this.props.Role || {}).value)
+            this.props.Role && this.props.Role.value === 'Other'
           }>
           <Field
             title={i18n.t('history.residence.label.role.explanation')}
--- a/src/components/Section/History/Residence/ResidenceItem.test.jsx
+++ b/src/components/Section/History/Residence/ResidenceItem.test.jsx
@@ -1,8 +1,12 @@
 import React from 'react'
-import { mount } from 'enzyme'
+import { shallow, mount } from 'enzyme'
 import ResidenceItem from './ResidenceItem'
 
 describe('The residence component', () => {
+  it('renders without crashing', () => {
+    shallow(<ResidenceItem />)
+  })
+
   it('no error on empty', () => {
     const expected = {
       name: 'residence'


src/components/Section/History/Residence/ResidenceItem.jsx
```diff
@@ -332,8 +332,7 @@ export default class ResidenceItem extends ValidationElement {
         </Field>
         <Show
           when={
-            (this.props.Role || {}).value &&
-            !['Own', 'Rent', 'Military'].includes((this.props.Role || {}).value)
+            this.props.Role && this.props.Role.value === 'Other'
```
ryanhofdotgov: So `this.props.Role` doesn't always have a property named `value`, but with this change that undefined result is compared to `Other`. If you put a breakpoint at this line and open up this page, you can see that happening when you create a new residence (and Role.value is not set), or you have a fresh application form.

Chrome doesn't throw an error when comparing against `undefined`, but it doesn't seem like we'd want to do comparisions like that. That's why you see this pattern all over the place, checking to see if a property is set before using it.

christopherhuii: What's wrong with comparing `undefined` to the string "Other"?

The conditional will be falsy and the condition is not satisfied so the explanation field will not display. I'm not aware of any risk or anti pattern surrounding this.

hursey013: @el-mapache and I talked about this pattern the other day, and would agree that the `|| {}` seems unnecessary, however would like to get @el-mapache's opinion again just in case this is catching something that I'm unaware of.

ryanhofdotgov:
I'll defer to @el-mapache recommends. I guess my main point is the original authors seem to have taken great pains not to access object properties that are undefined all over the code base, so want to make sure we are all on the same page before we start changing this paradigm. 

It's possible that they just needed it in one place, but decided to uniformly apply it everywhere, making code unnecessarily complicated, and we should fix that.

If it changes the calculus at all, recently we've discussed adopting some type of type checking for the frontend:
https://github.com/18F/e-QIP-prototype/issues/658

src/components/Section/History/Residence/ResidenceItem.jsx
el-mapache: While I can see why they used this pattern in other sections of the code, in this case it's unnecessary to repeat the `|| {}` guard multiple times, and @christopherhuii 's check is sufficient!

el-mapache: One thing I would like to see is less reliance overall on hard-coded strings, but that should be examined in a separate issue

el-mapache:
@ryanhofdotgov I think that is exactly what happened, they went with the pattern even when it didn't make sense. Type checking could be a good solution, as would leveraging default props in the components. 

I also notice that none of the reducers seems to have a default state, which would also be useful.

ryanhofdotgov:
Thanks everyone and sorry for my confusion @christopherhuii

christopherhuii:
@ryanhofdotgov No problem! You raised a good point, considering this is a common pattern seen across the codebase.

PR #964, opened 2018-10-09
Update Foreign Associations copy

### LINKED ISSUE #371: Remove text in Foreign Associations > Immediate Family Contact
Remove text in Foreign Associations > Immediate Family Contact

"For Section 7" (highlighted in the text block) should be removed or changed to Section 6 which is the section number for Foreign Associations

screen shot 2018-05-25 at 7 04 50 pm

christopherhuii:
Fixes #371 

Removed the text "For Section 7". ~I chose to remove the text rather than changing 7 to 6 because none of the other questions reference the section number in the description.~
I updated the text to match the form.
screen shot 2018-10-09 at 11 22 29 pm

For consistency, I felt that this section shouldn't be different.

### DIFF
--- a/src/config/locales/en/foreign.js
+++ b/src/config/locales/en/foreign.js
@@ -1372,7 +1372,7 @@ export const foreign = {
       },
       para: {
         intro:
-          '"Immediate Family" means your spouse, parents, step-parents, siblings, half and step-siblings, children, stepchildren, and cohabitant.',
+          'For this question, "Immediate Family" means your spouse, parents, step-parents, siblings, half and step-siblings, children, stepchildren, and cohabitant.',
         branch: [
           'Such as  embassy, consulate, agency, military service, intelligence or security service, etc.',
           'Answer "No" if the contact was for routine visa applications and border crossings related to either official U.S.  Government travel, foreign travel on a U.S. passport, or as a U.S. military service member in conjunction with a U.S. Government military duty.'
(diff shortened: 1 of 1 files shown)


ryanhofdotgov:
Before merging this in, we need to run this change against the NBIB policy team representative as it is a deviation from the official text of the form.

christopherhuii:
screen shot 2018-10-09 at 11 22 29 pm

On the form, it looks like the text says:
> For this question,

I have updated my PR to reflect what the form says.

ryanhofdotgov:
Yeah, the problem is that the form is re-organized in eApp and the question is repeated in the PDF. Sorry, this ticket wasn't quite ready to be picked up... it had a comment "Run by Drew" but that wasn't clear at all.

ryanhofdotgov:
It ends up that the other part of eApp does have this text already. It's just in a different position on the page. This is good to go. 

screen shot 2018-10-16 at 4 53 32 pm

ryanhofdotgov:
@christopherhuii Looks like I forgot to check that tests were passing before merging in. When you make certain changes to text, pages, etc. you'll have to rebuild the jest snapshots:
https://github.com/18F/e-QIP-prototype/blob/develop/docs/frontend.md#tests

At the moment, you have to run `yarn` to install the dependencies locally (outside the container set), then i run `yarn test -u`

You can see what it does in this PR:
https://github.com/18F/e-QIP-prototype/pull/1023

Issue #914, opened 2018-10-01
Employment validation fails for young adults


ryanhofdotgov:
**What did you do?**

Enter test data for an 18 year old applicant, with one job going back to age 16.

**What did you expect to see?**

No validation error in employment section.

**What did you see instead?**

Validation error prompting for 10 years of employment history.

**Additional Information**

Excerpt from SF-86:
"Do not list employment before your 18th birthday unless to provide a minimum of 2 years employment history."

Also see corresponding section in Validation Matrix PDF to determine appropriate validation logic.

PR #967, opened 2018-10-10
Fix length of required history based on age

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/914

The `History` component looks at the applicant's birthdate to determine how many years of history (employment, residence, employment) is required. This fixes the function that was processing the birthdate to use the correct value (`identification.ApplicantBirthDate.Date` rather than`identification.ApplicantBirthDate`). The function was also included twice, so this removes one instance of the function.

### DIFF
--- a/src/components/Section/History/History.jsx
+++ b/src/components/Section/History/History.jsx
@@ -44,6 +44,9 @@ export const sort = (a, b) => {
   return 0
 }
 
+/**
+ * Figure the total amount of years to collect for the timeline
+ */
 export const totalYears = birthdate => {
   let total = 10
   if (!birthdate) {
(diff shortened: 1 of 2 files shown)


ryanhofdotgov:
@hursey013 it seems to me like the whole paradigm of how it's calculating the ranges for which we need to collect data for is fairly imprecise.

It's calculating totalYears through some ceiling rounding, then from there doing some imprecise subtractions regardless of leap years, etc. 

Here's an example using PostgreSQL which has a bonafide date calculation library:

```
postgres=# select 'January 1, 2018'::timestamp - '10 years'::interval;
      ?column?       
---------------------
 2008-01-01 00:00:00
(1 row)

postgres=# select 'January 1, 2018'::timestamp - ((365 * 10)::text || 'days')::interval;
      ?column?       
---------------------
 2008-01-04 00:00:00
```

There were 3 leap years during this 10 year period. Since we aren't always looking for precision, but duplication of what e-QIP enforces, we need to verify what e-QIP does to cover these edge cases. I guess we should accept this PR as it is at least consistent with the imprecise paradigm of calculation and create a new issue to dive more deeply?

hursey013:
Yes, the calculations seem far from precise. I guess my thought would be that if we do improve this, it should likely be part of a larger refactor where we use an existing library like moment.js to replace the majority of the custom date logic that exists in the app today.

Issue #835, opened 2018-09-19
If you answer yes to dual citizenship, must have at least 2 entries 


ryanhofdotgov:
Per validation matrix and error response from e-QIP when testing, if you answer yes to dual citizenship you have to have at least 2 citizenships listed in Dual/multiple citizenship section.

This means if the applicant has U.S. citizenship and something else, the user should have an entry for their U.S. citizenship.

If citizenship entry is for US, do not ask:
"How did you acquire this non-U.S. citizenship you now have or previously had?"
"Have you taken any action to renounce your foreign citizenship?"
and do not prompt for a renounce explanation.

Test case 6 has this scenario. Update when resolved.

PR #971, opened 2018-10-11
Require multiple citizenship entries

hursey013:
Adds the requirement that at least two countries are added to the Dual/multiple citizenship section based on the validation matrix. Also removes the following questions if the country added is the United States:

- How did you acquire this non-U.S. citizenship you now have or previously had?
- Have you taken any action to renounce your foreign citizenship?
  - Provide explanation

**Preview:**
screen shot 2018-10-11 at 8 39 09 am


hursey013:
Closes https://github.com/18F/e-QIP-prototype/issues/835

PR #974, opened 2018-10-11
Use ToggleableAddress in lieu of Address for offenses

### LINKED ISSUE #755: Offense locations must not collect street address

el-mapache:
Fixes #755 

* Also replace custom `branch` logic in ToggleableLocation with Branch
component

**BEFORE**
screen shot 2018-10-16 at 9 05 11 am

**AFTER**
screen shot 2018-10-15 at 12 54 39 pm

**EDIT**: Added county

screen shot 2018-10-17 at 10 09 47 am


src/components/Form/Location/Layouts.js
```diff
@@ -14,5 +14,6 @@ export default {
     'US City, State, Zipcode International city',
   CITY_STATE_COUNTRY: 'City, State, Country',
   US_ADDRESS: 'US Address',
-  STREET_CITY: 'Street, City'
+  STREET_CITY: 'Street, City',
+  OFFENSE: 'US City, State, Zipcode, International city country'
```
hursey013: Would it be worth naming this more generically like the others in case this same layout ever needs to be used elsewhere in a non-offense context?

el-mapache: Probably!

el-mapache:
If someone has an idea for a more generically named 'offense' address I'd definitely entertain it. I really would not want to have to use something like: `CITY_STATE_ZIPCODE_COUNTRY_INTERNATIONAL_CITY_COUNTRY`, which seems to be the existing pattern

el-mapache:
My mistake, I thought county wasn't included when I was looking at the validation matrix, but I obviously misread it!

el-mapache:
@ryanhofdotgov Updated to include `county` for domestic offenses

api/testdata/complete-scenarios/test5.xml
```diff
@@ -4743,6 +4750,7 @@
                   <Court>
                     <Location>
                       <Place>
+                        <Street>9889 99th Ave. </Street>
```
ryanhofdotgov: Place where street is added

api/templates/location-city-state-zipcode-country.xml
```diff
@@ -0,0 +1,4 @@
+<City>{{.props.city}}</City>
+<State>{{toUpper .props.state}}</State>
+<ZipCode>{{.props.zipcode}}</ZipCode>
+<Country>{{.props.country}}</Country>
```
ryanhofdotgov: This needs to pull `.props.county` into `<County>`

api/testdata/complete-scenarios/test5.xml
ryanhofdotgov: I think everywhere a street got added unexpectedly is a place that has the wrong layout in the test case; needs to be updated to "Offense" to match what the front end would really spit out today.

It's being generated with the "Address" layout and the work-around of #755 is no longer there to prevent the default encoding for "Address".

src/components/Form/Location/Location.jsx
```diff
@@ -607,6 +608,11 @@ export default class Location extends ValidationElement {
             required={this.props.required}
           />
         )
+      // XXX This is used in ContactItem, which is foreign business contacts,
+      // would we ever need domestic fields for this?
```
ryanhofdotgov: Yeah, domestic addresses are relevant here because the form asks you where (for instance) the foreign job offer was made – and even though the job might be foreign, the offer might have been made at a location in the US.

api/location.go
```diff
@@ -159,6 +160,12 @@ func (entity *Location) Valid() (bool, error) {
 		stack = validateFields(entity, "street", "city", "state", "zipcode")
 	case LayoutStreetCity:
 		stack = validateFields(entity, "street", "city")
+	case LayoutOffense:
+		if domestic {
+			stack = validateFields(entity, "city", "state", "zipcode", "country")
```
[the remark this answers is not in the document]
el-mapache: yes, good catch!

api/templates/location-city-state-zipcode-country.xml
ryanhofdotgov: Forgot to add that would be best to rename the file so as not to reference `country` but `county`.

src/components/Form/Location/Location.jsx
el-mapache: Ah yes, I see that on the matrix. Thanks!

api/templates/location-city-state-zipcode-country.xml
el-mapache: Should we remove `<Country>` from here then? It is assumed to be the US unless it's specifically a foreign country, right?

ryanhofdotgov: Yeah, sorry – no `Country` element in this one.

el-mapache:
@ryanhofdotgov Now the XML being generated for the tests is omitting all fields after city. Not sure what needs to happen to resolve this issue

ryanhofdotgov:
@el-mapache Do you have any un-pushed local changes by chance, or is the last commit a6f01d6 the latest? I think when I looked at this commit the last time there was a bad path due to the rename?

el-mapache:
@ryanhofdotgov It looks like the file is there. The only addresses that seem to be affected are the ones with the `<County>` tag added, I think. I don't see anything that looks obviously malformed in the JSON or XML though

ryanhofdotgov:
Thanks. My local copy was in an odd state. I think just restoring "United States" to those Offense addresses will do the trick.

Issue #815, opened 2018-09-18
Logout should save before logging out


ryanhofdotgov:
**What did you do?**

Fill out some fields on a sub-section page (e.g., `Dual/multiple citizenship`). Log out. Re-log in. Navigate to sub-section page.

**What did you expect to see?**

The data I had filled out.

**What did you see instead?**

No data.

**Additional Information**

Sub-section page gets saved when you click to next page or press save manually and ?. Should also save before logging out automatically.

PR #1006, opened 2018-10-15
Save form prior to logging out

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/815

### DIFF
--- a/src/components/Navigation/Logout.jsx
+++ b/src/components/Navigation/Logout.jsx
@@ -2,6 +2,7 @@ import React from 'react'
 import { connect } from 'react-redux'
 import { i18n } from '../../config'
 import { logout } from '../../actions/AuthActions'
+import { saveSection } from '../SavedIndicator/persistence-helpers'
 
 class Logout extends React.Component {
   constructor(props) {
@@ -10,7 +11,17 @@ class Logout extends React.Component {
   }
 
   logout() {
-    this.props.dispatch(logout())
+    const application = this.props.app
+    const section = this.props.section.section
+    const subsection = this.props.section.subsection
+
+    saveSection(application, section, subsection, this.props.dispatch)
+      .then(() => {
+        this.props.dispatch(logout())
+      })
+      .catch(error => {
+        alert(error)
+      })
   }
 
   render() {
@@ -22,4 +33,13 @@ class Logout extends React.Component {
   }
 }
 
-export default connect()(Logout)
+function mapStateToProps(state) {
+  const section = state.section || {}
+  const app = state.application || {}
+  return {
+    section: section,
+    app: app
+  }
+}
+
+export default connect(mapStateToProps)(Logout)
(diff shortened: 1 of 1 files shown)


src/components/Navigation/Logout.jsx
```diff
@@ -10,7 +11,17 @@ class Logout extends React.Component {
...
 
   logout() {
-    this.props.dispatch(logout())
+    const application = this.props.app
+    const section = this.props.section.section
+    const subsection = this.props.section.subsection
+
+    saveSection(application, section, subsection, this.props.dispatch)
+      .then(() => {
+        this.props.dispatch(logout())
```
el-mapache: This pattern is used around the app, but is sort of backwards. Typically, one would want to have a dispatch handler pass the dispatch to an action, and that action would handle calling whatever functions it needed to. 

For example, in this component:

```
mapDispatchToProps = (dispatch) => ({ handleLogout: authenticationActions.logout(dispatch) })
```

Then after the `saveSection` call: `saveSection(...args).then(handleLogout)`, etc.

In this app, we usually see the pattern of the dispatch being manually passed down through parent props, and then dispatching actions directly from the component. I'm not sure if we want to start transitioning these components to use the typical redux pattern, or if its better to get these bug fixes in and open new issues to audit this code.

src/components/Navigation/Logout.jsx
```diff
@@ -22,4 +33,13 @@ class Logout extends React.Component {
   }
 }
 
-export default connect()(Logout)
+function mapStateToProps(state) {
```
el-mapache: A more es6 way to do this: `mapStateToProps = ({ section = {}, application = {} })`

Then we don't need to have the `|| {}` directly in the function. Of course, these values should be defined in a default state somewhere in the reducers, but that is another issue.

src/components/Navigation/Logout.jsx
ryanhofdotgov: I think get the bug fixes in, get automated browser tests going, then start doing across the board code audits to refactor things in the typical pattern, to ease maintenance and onboarding.

src/components/Navigation/Logout.jsx
hursey013: Just want to make sure I'm following this, so it would look like this instead?
```
mapStateToProps = ({
  section = {},
  application = {}
})

export default connect(mapStateToProps)(Logout)
```

ESLint complains about that (`Shorthand property assignments are valid only in destructuring patterns`), and would I still need to return something?

el-mapache: Sorry, the code I suggested was not all the code you'd need. Expanded:

```
function mapStateToProps({ section = {}, application = {} }) {
  return {
    section,
    app: application
  }
}
```

hursey013: Got it, thanks, updated

Issue #709, opened 2018-08-17
Entering numbers in Middle Name Field shows wrong errors 


bpdesigns:
**What did you do?**
(relates to all middle name fields)
entered 1 into the middle name and got the incorrect error 
<img width="1231" alt="screen shot 2018-08-17 at 3 17 05 pm" src="https://user-images.githubusercontent.com/15329140/44286786-5bfd5f00-a238-11e8-8a96-b3ed13f144a5.png">

entered 11  into the middle name and got 2 incorrect errors

<img width="1082" alt="screen shot 2018-08-17 at 4 10 44 pm" src="https://user-images.githubusercontent.com/15329140/44286825-7afbf100-a238-11e8-9327-ef184847df40.png">

**Expected to see**
An error stating "Only use letters, hyphens (-), periods (.), apostrophes ('), and spaces."
 - What browser and version are you using?
chrome
 - What operating system?
mac

gidjin:
I see a similar issue with the First Name field when entering numbers. Should the expected error be the same as noted here and Last Name?

PR #1008, opened 2018-10-15
Fix for 709 middle name field shows wrong errors

gidjin:
Fixes https://github.com/18F/e-QIP-prototype/issues/709


ryanhofdotgov:
@gidjin Thanks – when a PR is ready for review just add Brian or myself for this sprint.

Issue #157, opened 2018-05-23
Your history- where you went to school - person who knows you email


ryanhofdotgov:
From e-QIP-prototype-truetandem created by [Jay-Neighbours-I72](https://github.com/Jay-Neighbours-I72) : ryanhofdotgov/e-QIP-prototype-truetandem#2856  

Person who knows you branch>

- [ ] Email option should say "I don't know" instead of "Not applicable"

![screen shot 2018-04-11 at 1 53 23 pm](https://user-images.githubusercontent.com/19935974/38634193-d0595e84-3d8f-11e8-9256-e8784d63b260.jpg)
![screen shot 2018-04-11 at 1 53 18 pm](https://user-images.githubusercontent.com/19935974/38634196-d188c43e-3d8f-11e8-853d-19de1e63ca91.jpg)

PR #1017, opened 2018-10-16
Update Not applicable to instead read "I don't know"

line47:
Fixes https://github.com/18F/e-QIP-prototype/issues/157
screen shot 2018-10-16 at 2 30 37 pm

### DIFF
--- a/src/config/locales/en/reference.js
+++ b/src/config/locales/en/reference.js
@@ -32,7 +32,7 @@ export const reference = {
       comments: 'Add a comment to this question'
     },
     email: 'Provide email address for this person',
-    idk: 'Not applicable',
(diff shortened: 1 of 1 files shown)


hursey013:
It does look like there are some sections of the form that still use `Not applicable` - should we support both options, `Not applicable` and `I don't know`?

line47:
the `NotApplicable` control can be overridden with the label prop like in this example for `DelinquientItem`

```JSX
<NotApplicable
            name="ResolvedNotApplicable"
            label={i18n.t('financial.delinquent.label.notresolved')}
```

hursey013:
Got it, I see now, I thought that was a global value for the button, my bad!

PR #1024, opened 2018-10-17
First stage of error codes for save.go

macrael:
This PR makes the Save handler return error codes. Tests and the other handlers coming later.


ryanhofdotgov:
Thanks for carving this out!

Issue #898, opened 2018-09-27
Alias dates used validation error not getting caught by overall validation


ryanhofdotgov:
**What did you do?**

Alter birth date in test5.json to 01/05/1992. Load up test case. Go to overall validation review page.

**What did you expect to see?**

A validation error for dates used with alias 1. 

**What did you see instead?**

No validation error.

**Additional Information**

If you go to section 1's section review page, you'll get a validation error.

<img width="400" alt="screen shot 2018-09-27 at 1 31 25 am" src="https://user-images.githubusercontent.com/37100189/46133447-6165a600-c1f5-11e8-8b8b-b9c257f9c78a.png">

ryanhofdotgov:
If it is related to #925, probably best to to wait until #925 is resolved first.

PR #1026, opened 2018-10-17
Check min and max dates in date validator

### LINKED ISSUE #632: Invalid date of birth isn't flagged in sub-section Review, overall Review
**What did you do?**

Enter 01/01/3000 as date of birth in "Information about you" section. Go to "Information about you" Review section or go to overall Review page.

Review pages flag data of birth error. In case of overall Review section, also: don't let form be submitted until data of birth error resolved.

**What did you see instead?**

Date of birth sub-section shows error, but Review pages don't list it. Was able to get to submit stage with the bad data.

hursey013:
Fixes #632 

On the final review, as the app loops through the various sections and calls the validator of each subsection to check validity. In the case of the applicant's birthdate it was ultimately calling `DateControlValidator.isValid()` which only checks that there is a month, day, and year (https://github.com/18F/e-QIP-prototype/blob/develop/src/validators/datecontrol.js#L96-L106). However the `DateControl` component itself has some validation itself that does the min/max checks: https://github.com/18F/e-QIP-prototype/blob/develop/src/components/Form/DateControl/DateControl.jsx#L484-L512 which is why errors are displayed on the actual subsection. So calling `DateControlValidator.isValid()` directly, as the final review page does, bypasses those checks giving the impression that it's valid.

This incorporates the min/max checks into `DateControlValidator.isValid()` which should allow incorrect dates to be flagged everywhere, and hopefully catch other similar cases in which invalid dates are not being caught in the final review.

### DIFF
--- a/src/validators/daterange.js
+++ b/src/validators/daterange.js
@@ -19,11 +19,11 @@ export default class DateRangeValidator {
    * Validates the date ranges
    */
   isValid() {
-    if (!new DateControlValidator(this.from || {}).validDate()) {
+    if (!new DateControlValidator(this.from).validDate()) {
       return false
     }
 
-    if (!new DateControlValidator(this.to || {}).validDate()) {
+    if (!new DateControlValidator(this.to).validDate()) {
       return false
     }
 
--- a/src/validators/daterange.js
+++ b/src/validators/daterange.js
@@ -19,11 +19,11 @@ export default class DateRangeValidator {
    * Validates the date ranges
    */
   isValid() {
-    if (!new DateControlValidator(this.from || {}).isValid()) {
+    if (!new DateControlValidator(this.from || {}).validDate()) {
       return false
     }
 
-    if (!new DateControlValidator(this.to || {}).isValid()) {
+    if (!new DateControlValidator(this.to || {}).validDate()) {
       return false
     }
 
--- a/src/validators/datecontrol.js
+++ b/src/validators/datecontrol.js
@@ -93,7 +93,7 @@ export default class DateControlValidator {
     return date > this.minDate
   }
 
-  isValid() {
+  validDate() {
     if (
       (!this.day && !this.hideDay) ||
       (!this.month && !this.hideMonth) ||
@@ -101,8 +101,11 @@ export default class DateControlValidator {
     ) {
       return false
     }
-
     return true
+  }
+
+  isValid() {
+    return this.validDate() && this.validMaxDate() && this.validMinDate()
   }
 }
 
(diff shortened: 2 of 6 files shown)


src/validators/daterange.js
```diff
@@ -19,11 +19,11 @@ export default class DateRangeValidator {
    * Validates the date ranges
    */
   isValid() {
-    if (!new DateControlValidator(this.from || {}).isValid()) {
+    if (!new DateControlValidator(this.from || {}).validDate()) {
```
el-mapache: This `|| {}` is redundant as the `DateControlValidator` will already default to an empty object when `this.from` is undefined. This can technically cause an issue if `this.from` is set to `false` or `null`, but since it is being controlled from within the component I think we can be sure it isn't.

hursey013: Yeah, I'll try and start removing these as I come across them if it's safe. Updated https://github.com/18F/e-QIP-prototype/pull/1026/commits/1210c9202e4cc2b869858085061f2782b4f5cad0

ryanhofdotgov:
@hursey013 How does validMaxDate() and validMinDate get called on the fields of the date range object? Does it makes sense that this PR has datecontrol isValid calling min/max related methods, but on a range the corresponding validator code doesn't call them? 

In other words, there's a lack of symmetry in the datarange validator isValid and the datecontrol one. I would expect daterange isValid to call isValid on its pair of controls plus additional stuff.

I've looked a few different date-related patches recently and its all very fuzzy for me.

src/validators/daterange.js
hursey013: So I guess in this case these were needed to support `null` values being passed from tests. I updated the tests to pass in a `{}` instead, but I wonder if that is part of the explanation for why these are all around?

hursey013:
@ryanhofdotgov this is an unfortunate side effect of how the `DateRange` validator was implemented. Throughout the app, when the `DateRangeValidator` is called, it is only passed a date, not any other important data like `minDate`, `noMaxDate` (as is the case with `DateControl`). So when `DateRangeValidator` ultimately passes the date along to `DateControlValidator` it's doing so without an of that additional context which causes unexpected results.

An example I found when checking the test data was with the passport dates. The passport validator calls `DateRangeValidator` and passes it the issued and expiration dates. The `DateRangeValidator` then passes the dates to `DateControlValidator`, but since `noMaxDate` wasn't passed along with the expiration date, it gets flagged as having a date in the future by `DateControlValidator`. 

Accessing `DateRangeValidator` in this way happens in ~40 different components, so to fix it would require updating `DateRangeValidator`to take in all of the props is should be, then passing them along to `DateControlValidator`. All of the calls to `DateRangeValidator` would need to be adjusted to pass the entire DateControl object rather than just the date specifically as well.

hursey013:
https://github.com/18F/e-QIP-prototype/issues/898 is related to this issue as well.

PR #1052, opened 2018-10-22
Update identifying document numbers to set max number and make alphanumeric

### LINKED ISSUE #922: Document identifying numbers may not be longer than 30 characters
test case 7 originally had e-QIP validation errors on various document numbers because they exceeded 30 characters:

`AlienRegistrationNumber`:
`PermanentResidentCardNumber`:
`CitizenshipCertificateNumber`:
`NaturalizationCertificate`
`StateDeptForm240/DocumentNumber`

Update these fields to limit input to 30 characters. Search validation PDF for "30 Alphanumeric" and double-check there are no other similar document number fields in eApp that should be bounded in length.

The element names above are from the e-QIP XML. Identify the corresponding eApp form fields by navigating the application, inspecting test case files, or the  XML `api/templates`.

line47:
This fixes #922 

Example of the fields throwing errors when encountering a special character
screen shot 2018-10-22 at 11 22 19 am

### DIFF
--- a/src/components/Form/ForeignBornDocuments/ForeignBornDocuments.jsx
+++ b/src/components/Form/ForeignBornDocuments/ForeignBornDocuments.jsx
@@ -290,6 +290,9 @@ export default class ForeignBornDocuments extends ValidationElement {
             label="Document Number"
             {...this.props.DocumentNumber}
             className="foreign-born-document-number"
+            maxlength="30"
+            pattern="^[a-zA-Z0-9]*$"
+            prefix={"alphanumeric"}
             onUpdate={this.updateDocumentNumber}
             onError={this.props.onError}
             required={this.props.required}
--- a/src/components/Form/ForeignBornDocuments/ForeignBornDocuments.jsx
+++ b/src/components/Form/ForeignBornDocuments/ForeignBornDocuments.jsx
@@ -293,7 +293,7 @@ export default class ForeignBornDocuments extends ValidationElement {
             className="foreign-born-document-number"
             maxlength="30"
             pattern={alphaNumericRegEx}
-            prefix={"alphanumeric"}
+            prefix="alphanumeric"
             onUpdate={this.updateDocumentNumber}
             onError={this.props.onError}
             required={this.props.required}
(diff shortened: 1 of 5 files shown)


src/components/Form/ForeignBornDocuments/ForeignBornDocuments.jsx
```diff
@@ -290,6 +290,9 @@ export default class ForeignBornDocuments extends ValidationElement {
             label="Document Number"
             {...this.props.DocumentNumber}
             className="foreign-born-document-number"
+            maxlength="30"
+            pattern="^[a-zA-Z0-9]*$"
+            prefix={"alphanumeric"}
```
el-mapache: I think we can omit the `{}` in each of these instances since we aren't trying to interpolate a variable, the quoted string should be enough

src/components/Form/ForeignBornDocuments/ForeignBornDocuments.jsx
```diff
@@ -290,6 +290,9 @@ export default class ForeignBornDocuments extends ValidationElement {
             label="Document Number"
             {...this.props.DocumentNumber}
             className="foreign-born-document-number"
+            maxlength="30"
+            pattern="^[a-zA-Z0-9]*$"
```
el-mapache: Since this is repeated in a number of places over a few files, lets make this a constant somewhere in the app, and import it to the files we need it in. This way, if we have to make any adjustments to the regex, we wont have to change it in a bunch of places.

src/config/locales/en/error.js
```diff
@@ -1095,5 +1095,11 @@ export const error = {
   validMinimumCitizenships: {
     title: 'There is a problem with your entry',
     message: 'Please provide a minimum of two countries.'
+  },
+  alphanumeric: {
+    pattern: {
+      title: 'There is a problem with your entry',
+      message: 'Only use letters and numbers'
+    }
```
hursey013: Since these fields set a maxlength and are required, I think we'll need to add those validation messages in here under the `alphanumeric` key as well since that prefix is going to get applied to all of the messages.

line47: Ah thanks for that!

PR #1062, opened 2018-10-25
Add current git tag version to introduction modal

### LINKED ISSUE #844: Add a version number, queryable by end user
JITC tester wants to be able to see what version of eApp they are looking at (e.g., clicking an about button, version number in corner, etc.), so they can maintain/confirm meta-data in their testing records.

hursey013:
Fixes #844 

Uses webpack to set latest Git tag to a constant available to the app. Uses commit hash as a fallback if no tag is available. 

screen shot 2018-10-25 at 4 19 59 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,10 @@
     "clearMocks": true,
     "roots": [
       "<rootDir>/src"
-    ]
+    ],
+    "globals": {
+      "EAPP_VERSION": "MAJOR.MINOR.PATCH-IDENTIFIER"
+    }
   },
   "scripts": {
     "test": "NODE_ENV=test jest --ci --coverage --silent",
@@ -77,6 +80,7 @@
     "eslint-import-resolver-webpack": "^0.10.1",
     "eslint-plugin-import": "^2.13.0",
     "eslint-plugin-react": "^7.10.0",
+    "git-revision-webpack-plugin": "^3.0.3",
     "gulp": "^4.0.0",
     "jest": "^23.1.0",
     "redux-devtools-extension": "^2.13.2",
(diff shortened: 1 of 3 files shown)


hursey013:
Looks like we have a number of tags associated with the `develop` branch which will cause it to report inaccurate tag version numbers - would it be ok to remove those tags?

ryanhofdotgov:
I haven't made progress on the follow-up to this ticket since I talked with @hursey013 a couple of weeks ago, but here is where we are at:
* It's not clear for me still how to get the tags in GitHub the way this code expects them
* We could delete tags, but a destructive step that destroy tag history is not desirable
* Not clear how this all works when this code is imported into downstream forks and haven't spent time to test that out yet. 

I'm going to throw a [WIP] prefix to this PR for now, and if we don't resolve it within the next week or so, kick it out to the backlog.

hursey013:
As it stands, this just pulls in the last known tag + commits ahead of when it was tagged + current commit hash (ignoring commits ahead of when it was tagged + current commit hash if nothing has changed since it was tagged).  We might be able to workaround the older, inaccurate tags, by adding some logic to ignore them before a certain point, and maybe just show the branch name or commit hash in those cases.

Whether or not other folks are tagging in a similar way downstream is another story though, if we want to be safe we could just manually set a version number in some sort of global variable and just use that.

ryanhofdotgov:
@hursey013 After digging into it even more, while we are mostly using the Gitflow workflow, I haven't been using the git flow extension to manage releases.

If you use that extension, a separate release branch is created off develop, that release branch is tagged, the release branch is merged in to master, then the release branch is merged back into develop. Reference: `github.com/nvie/gitflow/issues/432#issuecomment-197343631`

If you respect that process, then `git describe` works the way you'd want and `git-revision-webpack-plugin` expects. 

I'm going to merge this PR in and start executing that aspect of the workflow.

hursey013:
Sounds good to me 👍Thanks for digging into all of this, I definitely learned a lot more about git tags than I was expecting!

PR #1063, opened 2018-10-26
Print entire contents of textarea

### LINKED ISSUE #1060: Archival PDF text areas clip large text values
**What did you do?**

Load test case 7, bring it through submit stage via debug force submit, print PDF. 

All the text value in all text areas/fields, displayed completely.

**What did you see instead?**

Text areas (almost all of them) only showing the partial text since:
* tester inserted multiple pages of text
* text area limited in display to a few dozen lines of text

**Additional Information**

See attached eApp archival and compare to equivalent e-QIP archival (or `test7.json` and load it) and note the many text areas with clipped text. Looking for "It has long been a grave" will find most of the text blocks that are affected.

[e-QIP_Request_20020667_Archival_Copy.pdf](https://github.com/18F/e-QIP-prototype/files/2510706/e-QIP_Request_20020667_Archival_Copy.pdf)
[test7-eApp-archival.pdf](https://github.com/18F/e-QIP-prototype/files/2510709/test7-eApp-archival.pdf)

line47:
Added an additional item to the textarea component to render just the value of the textarea into a div during print. This fixes #1060

**Print view**
screen shot 2018-10-26 at 12 02 54 pm

**Screen view** 
screen shot 2018-10-26 at 11 56 52 am

### DIFF
--- a/src/components/Form/Text/Text.jsx
+++ b/src/components/Form/Text/Text.jsx
@@ -52,9 +52,11 @@ export default class Text extends ValidationElement {
   render() {
     return (
       <span className="textbox-print">
+      <label className={this.labelClass} htmlFor={this.state.uid}>
+        {this.props.label}
+      </label>
         <Generic
           name={this.props.name}
-          label={this.props.label}
           ariaLabel={this.props.ariaLabel}
           placeholder={this.props.placeholder}
           type="text"
@@ -81,9 +83,6 @@ export default class Text extends ValidationElement {
           tabNext={this.props.tabNext}
           ref="text"
         />
-        <label className={`print-only ${this.labelClass}`}>
-          {this.props.label}
-        </label>
         <div
         className={`text-print print-only ${this.props.className}`}>
           {this.state.value}
--- a/src/components/Form/Text/Text.jsx
+++ b/src/components/Form/Text/Text.jsx
@@ -51,35 +51,44 @@ export default class Text extends ValidationElement {
 
   render() {
     return (
-      <Generic
-        name={this.props.name}
-        label={this.props.label}
-        ariaLabel={this.props.ariaLabel}
-        placeholder={this.props.placeholder}
-        type="text"
-        className={this.props.className}
-        disabled={this.props.disabled}
-        status={this.props.status}
-        minlength={this.props.minlength}
-        maxlength={this.props.maxlength}
-        pattern={this.props.pattern}
-        readonly={this.props.readonly}
-        required={this.props.required}
-        value={this.state.value}
-        focus={this.props.focus}
-        onChange={this.handleChange}
-        onFocus={this.props.onFocus}
-        onBlur={this.props.onBlur}
-        onError={this.handleError}
-        onKeyDown={this.props.onKeyDown}
-        onCopy={this.props.onCopy}
-        onCut={this.props.onCut}
-        onPaste={this.props.onPaste}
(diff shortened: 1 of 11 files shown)


hursey013:
Looks good, just a couple of suggestions:

I think since all of the other data is shown within an input box, for consistency's sake it might make sense to mimic an input box border around the content. It might also help distinguish the content when it is included right next to other form guidance/questions:

screen shot 2018-10-26 at 12 55 39 pm

It also looks like there is a component called SectionComments which allows people to enter supplementary comments for each section. For whatever reason it uses a normal text input instead of a textarea, but these comments may be lengthy enough that will also become cut off. Maybe we could update this to use TextArea instead?

screen shot 2018-10-26 at 12 49 47 pm

line47:
I can add a border around the text although I also wonder what the idea print scenario would be? Is that documented anywhere? 

@ryanhofdotgov Do we want to update the comment box to be a textarea like @hursey013 mentions above so it works the same way?

ryanhofdotgov:
* There's nothing really prescriptive about how the eApp archival PDF should look like, so if there is a consensus around preserving the appearance of a border, let's go ahead and do that.

* The clipping issue is affecting HTML text areas as well as HTML input fields generally, does this address that variety, or just the text areas? Both types of input have a max input length and a max input display length which can be different.

Here is a list from NBIB I got today that shows some specific examples from test case 7 (but it isn't an exhaustive list):
https://docs.google.com/spreadsheets/d/1_3t2SdVq65dZmr66t3BMnefUuOyEI7nT7teZrE2zrdw/edit#gid=986991471

* The SectionComment should be a text area for on-screen purposes, to let user know larger amount of multi-line text can be entered right? If it is a separate issue from clipping I would open up a separate issue.

line47:
update to show a border around the textareas and also show contexts of textboxes when they are larger than the input box. 

screen shot 2018-10-29 at 11 18 49 am

screen shot 2018-10-29 at 11 17 21 am

ryanhofdotgov:
@line47 Can you go through this spreadsheet:
https://drive.google.com/open?id=13RxfJqSQX6qoyqRejajn4ue5JAaSCPpaJwWin44rF7s
and double-check that all the clipping issues it lists are addressed? There are 49 issues listed, all but the 5 gold-colored rows are clipping issues.

line47:
Hey @ryanhofdotgov I checked the spreadsheet and all the issues have been addressed. There is still a lot of room for improvement when printing. 

I also noticed that in a few places there are no labels associated with `textareas` (this is really bad for accessibility) Should I open an issue for that?

ryanhofdotgov:
Thanks! Yeah, open up tickets for any accessibility issues you feel need to be addressed and any improvements. Once you resolve the CircleCI tests, then I think this is good to go.

src/components/Form/Text/Text.jsx
```diff
@@ -51,35 +51,43 @@ export default class Text extends ValidationElement {
...
-        onCopy={this.props.onCopy}
-        onCut={this.props.onCut}
-        onPaste={this.props.onPaste}
-        clipboard={this.props.clipboard}
-        tabBack={this.props.tabBack}
-        tabNext={this.props.tabNext}
-        ref="text"
-      />
+      <span className="textbox-print">
+      <label className={this.labelClass} htmlFor={this.state.uid}>
```
line47: Added the label outside the generic control to have more control of what gets shown/hidden at print

PR #1066, opened 2018-10-26
Added offense layout to location validator

hursey013:
I noticed some validation errors when testing test case 5 with the `develop` branch. The offense validator uses `LocationValidator` to validate the address field, however the location validator did not have a corresponding layout for OFFENSE so it was being returned as invalid.

### DIFF
--- a/src/validators/location.js
+++ b/src/validators/location.js
@@ -8,7 +8,7 @@ export const isInternational = location => {
   )
 }
 
-export const countryString = (country) => {
+export const countryString = country => {
   if (country && country.value) {
     if (Array.isArray(country.value)) {
       return country.value[0]
@@ -17,7 +17,7 @@ export const countryString = (country) => {
     return country.value
   }
 
-  return country;
+  return country
 }
 
 export default class LocationValidator {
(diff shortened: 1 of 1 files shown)


hursey013:
I also updated the naming of `zipcodeState` to `stateZipcode` since that is how it is referred to everywhere else is in the application.

src/validators/location.js
```diff
@@ -8,7 +8,7 @@ export const isInternational = location => {
   )
 }
 
-export const countryString = (country) => {
+export const countryString = country => {
```
el-mapache: Actually, I would leave the `()` here; eslint convention has parens surrounding the arguments of any function that either

a). has multiple arguments
b). has  code other than a statement to be returned. in other words, a function with `{}`

hursey013: This change is coming from `prettier` - I've been running it based off of https://github.com/18F/e-QIP-prototype#formatting. We have a config file setup for `prettier`, so are parens with single arguments something we want to permit?

el-mapache: Doesn't matter to me! The typical eslint default is the one I've described, that's the one I've used most commonly while at 18F, but if we already have something else defined, we can go with that. 

Does prettier run automatically? It didn't realize there was  any style linting at all, since the files in the project are so inconsistent. We may want to have it automatically run when we run our tests.

hursey013: Added a new issue to have Prettier run automatically https://github.com/18F/e-QIP-prototype/issues/1076 and will address in a separate PR if that's 👍

Issue #841, opened 2018-09-20
Don't ask certain questions when filling out unemployment time period


ryanhofdotgov:
When filling out employment history, if an entry is for an unemployed period of time, don't ask these questions:
* For this employment, in the last seven (7) years have you received a written warning, been officially reprimanded, suspended, or disciplined for misconduct in the workplace, such as a violation of security policy?
* For this employment have any of the following happened to you in the last seven (7) years?
* Reason for Leaving 

See validation matrix PDF, starting around 53, looking at the specific rules that guide this change. It is a bit convoluted since it doesn't say "if unemployed to this" but rather "if employed by military or if employed by non-military or if employed by self, etc".

PR #1079, opened 2018-10-30
Do not show certain employment questions when unemployment is selected

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/841

This uses conditional logic to hide the `Reason for leaving` and `Reprimand` questions when `Unemployment` is chosen as the employment activity. Also removed some unnecessary `<div>`s so sorry for the added noise in the diff.

### DIFF
--- a/src/components/Section/History/Employment/EmploymentItem.jsx
+++ b/src/components/Section/History/Employment/EmploymentItem.jsx
@@ -464,7 +464,6 @@ export default class EmploymentItem extends ValidationElement {
           <div className="reference">
             <Field
               title={i18n.t('reference.heading.name')}
-              titleSize="h3"
               optional={true}
               filterErrors={Name.requiredErrorsOnly}
               scrollIntoView={this.props.scrollIntoView}>
--- a/src/components/Section/History/Employment/EmploymentItem.jsx
+++ b/src/components/Section/History/Employment/EmploymentItem.jsx
@@ -258,7 +258,6 @@ export default class EmploymentItem extends ValidationElement {
    *    - Federal Contractor
    *    - Non-Government
    *    - Self Employment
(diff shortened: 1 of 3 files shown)


src/components/Section/History/Employment/EmploymentItem.jsx
```diff
@@ -456,109 +454,107 @@ export default class EmploymentItem extends ValidationElement {
...
+
+          <div className="reference">
             <Field
-              title={i18n.t(`${prefix}.heading.reference`)}
-              titleSize="h2"
-              className="no-margin-bottom"
-              scrollIntoView={this.props.scrollIntoView}
-            />
+              title={i18n.t('reference.heading.name')}
+              titleSize="h3"
```
el-mapache: `h3` is the default, no need to pass it here

src/components/Section/History/Employment/EmploymentItem.test.jsx
```diff
@@ -133,6 +133,42 @@ describe('The employment component', () => {
     expect(component.find('.reason-description').length).toBe(0)
   })
 
+  it('does not display reason for leaving for unemployment', () => {
+    const past = daysAgo(today, 365 * 3)
```
el-mapache: The setups for these two tests are identical, so we should move them into a `beforeEach` block.

If possible, I think it would be cleaner to `find` the components you expect to to not be there by their names, rather than the css selectors (`'Reprimand'` and `'ReasonLeft'`).

hursey013: Makes sense, made some tweaks: https://github.com/18F/e-QIP-prototype/pull/1079/commits/389bff2c584dc1076b98112163c5e4980356ec52

Issue #996, opened 2018-10-15
Re-generate snapshots through the container


ryanhofdotgov:
As a developer, I want to be able to re-generate the front-end Jest snapshots using the tools/packages already set-up in the existing containers, and not rely on the versions in my host environment.

This command does not work:
```docker-compose run --rm js yarn jest --updateSnapshot```

Update associated build tooling and update the documentation  (https://github.com/18F/e-QIP-prototype/blob/develop/docs/frontend.md).

Setting up `yarn` separately in the host environment and running `yarn test --updateSnapshot` does work.

PR #1081, opened 2018-10-30
Improve React Makefile test command

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/996

This allows extra flags to be passed to `make react-test`, allowing finer grained control of testing locally vs on CI. `--silent` and `--ci` are now only used on CI, and a new `FLAGS` argument allows Jest flags to be passed to the Docker container, allowing for snapshots to be generated within the container (`make test-react FLAGS="--updateSnapshot"`).

### DIFF
--- a/docs/frontend.md
+++ b/docs/frontend.md
@@ -4,25 +4,43 @@ The frontend is built in [React](https://reactjs.org/), wired up with [React Rou
 
 Here are some of the major directories:
 
-* [`/`](..): Things like the [`package.json`](../package.json) and config files for other tools live at the root directory of the repository
-* [`specs/`](../specs): [Integration tests](../specs/README.md)
-* [`src/`](../src): All the frontend source files live here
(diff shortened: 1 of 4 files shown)


docs/frontend.md
```diff
@@ -4,26 +4,44 @@ The frontend is built in [React](https://reactjs.org/), wired up with [React Rou
...
+To run a subset of tests, use the `FILES` argument, using a space as a delimiter between file paths:
+
+```shell
+make test-react FILES="src/components/Navigation/Navigation.test.jsx src/components/Navigation/SectionLink.test.jsx"
+```
+
+If changes are made that require updating a snapshot(s), pass the appropriate flag using the `FLAGS` argument:
+
+```shell
+make test-react FLAGS="--updateSnapshot"
```
el-mapache: The only issue with this is that flags aren't additive, for example, you can't have `updateSnapshot` along with other commands, without explicitly rewriting them

PR #1082, opened 2018-10-30
Add HTTP Error response codes to all API errors

macrael:
This PR introduces RespondWithStructuredError() a method for returning a standardized error list and an HTTP response code for API errors. 

It goes through all the API handlers and returns appropriate error codes for errors.

Another task will be to sort out what the JSON response for most successful responses should be. For things like save and submit we should return something.

### DIFF
--- a/api/http/logout.go
+++ b/api/http/logout.go
@@ -28,12 +28,5 @@ func (service LogoutHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// If the account is locked then we cannot proceed
-	if account.Locked {
-		service.Log.Warn(api.AccountLocked, api.LogFields{})
-		// EncodeErrJSON(w, err)
-		return
-	}
-
 	service.Log.Info(api.LoggedOut, api.LogFields{})
 }
--- a/api/http/logout.go
+++ b/api/http/logout.go
@@ -16,28 +16,22 @@ type LogoutHandler struct {
 
 // ServeHTTP will end the user session.
 func (service LogoutHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	account := &api.Account{}
 
-	// Valid token and audience while populating the audience ID
-	_, id, err := service.Token.CheckToken(r)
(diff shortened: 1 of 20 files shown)


macrael:
I'm not sure what the protocol is here, I can go back and add some basic tests to all the handlers but they were not tested before so I didn't think this task was the place to get that done.

api/http/logout.go
```diff
@@ -16,28 +16,22 @@ type LogoutHandler struct {
...
-		EncodeErrJSON(w, err)
+		RespondWithStructuredError(w, api.NoAccount, http.StatusUnauthorized)
 		return
 	}
 
 	// If the account is locked then we cannot proceed
 	if account.Locked {
 		service.Log.Warn(api.AccountLocked, api.LogFields{})
-		EncodeErrJSON(w, err)
+		// EncodeErrJSON(w, err)
```
ryanhofdotgov: There's some commented out code here, but I think this block for locked can be removed entirely? Accounts get locked on application submission, so would still need to be able to logout.

api/http/refresh.go
```diff
@@ -17,22 +17,20 @@ type RefreshHandler struct {
 
 // ServeHTTP refreshes a given token.
 func (service RefreshHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	account := &api.Account{}
 
 	// Valid token and audience
 	_, id, err := service.Token.CheckToken(r)
```
ryanhofdotgov: In other handlers you've removed this check entirely since it was already handled by the http.JWTHandler. Does it make sense to make that consistent here, and add http.JWTHandler to where `/refresh` is set up in `main.go`? Would that make all the top-level authn happen in that one place?

api/http/logout.go
macrael: Agreed. No reason to block logout on locked.

api/http/refresh.go
macrael: I like that idea. Refresh is configured without the middleware currently, should be easy to fix.

Issue #924, opened 2018-10-03
Collect secondary addresses


ryanhofdotgov:
In certain parts of the SF-86, if an APO/FPO address is indicated or an non-US address is indicated, a secondary address is solicited from the applicant. You can find these instances by searching the SF-86 PDF for "If you have indicated an" and looking at the conditional `(a)` and `(b)` options:

<img width="500" alt="screen shot 2018-10-03 at 8 53 26 am" src="https://user-images.githubusercontent.com/37100189/46422819-f7b03500-c6e9-11e8-889c-e3bca9414fe3.png">

Verify that omitting these secondary addresses was not an approved policy decision, and if not, implement them.

Edit:
Scope of this ticket is to just collect and persist the secondary addresses. Follow-on work to properly encode them into e-QIP XML is in https://github.com/18F/e-QIP-prototype/issues/1027

hursey013:
@ryanhofdotgov Should we wait on confirmation before beginning this one?

ryanhofdotgov:
@hursey013 Yeah, before we can start on this, we need policy sign off. I'll work on that, and then put this back in To Do once we have it.

ryanhofdotgov:
@hursey013 Heard an initial take from policy team – we should go ahead and implement this, following the guidance on the PDF form (screenshot above)

1) If applicant enters an APO address, *require* collection of a physical address (see fields in screenshot and in `PhysicalLocation` element in applicant XSD).

2) If applicant enters a foreign address, show an optional field set to collect an APO address, first by asking the question "Did you have an APO/FPO address at this location?".

This secondary collection workflow appears at the following 6 locations in the form:
* Applicant residence address
* Employment address (including if employment is active duty, etc.)
* Applicant's current spouse residence address
* Address of person who can verify unemployment/self-employment
* Address of person who can verify residence
* Supervisor's address

These locations where identified by searching for "If you have indicated an APO" in the SF-86 PDF and cross-checking with the instances of `PhysicalLocation` in the applicant XSD.

ryanhofdotgov:
Upon closer inspection and in testing with e-QIP, employment address has a slight different secondary address workflow from all the others. Going to separate that out into https://github.com/18F/e-QIP-prototype/issues/1050

el-mapache:
@ryanhofdotgov Do we also need phone numbers for these associated addresses?

[...]

PR #1083, opened 2018-10-30
(WIP) Collect secondary address when residence is a military address

el-mapache:
Part of #924 

Prompts the user to enter a secondary physical address when a military address is given in the residence section, or to enter an APO address when they have indicated foreign residence.

Currently, this is only present within the Residence History component

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     ]
   },
   "scripts": {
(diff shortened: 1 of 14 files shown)


el-mapache:
Closing in favor of updated PR

PR #1085, opened 2018-10-30
Add isDefined function to location validator

el-mapache:
* Ensures that a country with a value of '' is properly returned from
the `countryString` method

### DIFF
--- a/src/validators/helpers.js
+++ b/src/validators/helpers.js
@@ -1,5 +1,12 @@
 import { daysAgo, today } from '../components/Section/History/dateranges'
 
+/**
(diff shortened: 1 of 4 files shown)


el-mapache:
We may want `null` to be treated as 'defined' in the future, but this will address the bugs with the `countryString` method

christopherhuii:
@el-mapache @ryanhofdotgov 
I have verified that clicking "Outside of the United Starts" on an address field does not throw an error anymore. I believe that is the scope of this PR.

However, there's still a bug for the `APO/FPO/DPO` button. It never highlights with the green background. We might need to create an issue for that if that work isn't captured.

PR #1086, opened 2018-10-31
Removing AWS GovCloud deploy

### LINKED ISSUE #854: Turn off updates to AWS GovCloud ECR
The updates we send to AWS GovCloud ECR are no longer relevant; updates to NBIS are handled a different way.

Remove all processes that pulls down and publishes updated images there.

ryanhofdotgov:
This deployment is no longer in use. Closes #854.


macrael:
Is this the first step in moving us into the NITSL?

PR #1087, opened 2018-11-01
Collect secondary address when residence is a military or foreign address

el-mapache:
Part of #924 

Prompts the user to enter a secondary physical address when a military address is given in the residence section, or to enter an APO address when they have indicated foreign residence.

Currently, this is only present within the Residence History component

### DIFF
--- a/src/components/Form/Field/Field.jsx
+++ b/src/components/Form/Field/Field.jsx
@@ -36,6 +36,8 @@ const message = id => {
   )
 }
 
+// XXX All references to `comments` in this component refer to help text / info boxes
+
 export default class Field extends ValidationElement {
   constructor(props) {
     super(props)
--- a/src/components/Form/Location/AlternateAddress.jsx
+++ b/src/components/Form/Location/AlternateAddress.jsx
@@ -21,11 +21,10 @@ class AlternateAddress extends ValidationElement {
     this.setAlternateAddress = this.setAlternateAddress.bind(this)
   }
 
-  // XXX This is not a great pattern, but we don't have a clear
-  // way to update the state of this AlternateAddress prop when
-  // a user chooses a 'country' value that is not a military address
-  componentWillUnmount() {
-    this.props.onUpdate(alternateAddressDefaultState())
+  componentDidUpdate(prevProps) {
+    if (countryString(prevProps.country) !== countryString(this.props.country)) {
+      this.props.onUpdateCountry(alternateAddressDefaultState());
+    }
   }
 
   handleUpdate(values) {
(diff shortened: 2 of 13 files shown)


ryanhofdotgov:
@el-mapache When doing tests I ran into https://github.com/18F/e-QIP-prototype/issues/1094, so you'll want to merge in latest from develop (https://github.com/18F/e-QIP-prototype/commit/c68be771bd4a3e7be4b13235bfc79f37968b23eb) to avoid running into that. 

I didn't review the code yet – just played around with it. A few things I observed:

* When entering foreign address as primary, then choosing "Yes" for branching question, the APO address field set appears as expected and I'm able to enter the address. Save and reloading works as expected. 🎉🎉

* The `HasDifferentAddress` branching property is in a nested in a different level in the structure than the branch properties usually are. This was intentional right? -- it makes the component more self-contained.

* When entering in a primary foreign address and then choosing "Yes" to the branching question, the APO address fields are immediately marked in red. This is a different paradigm than the rest of the form, where required fields don't start off flagged in red with an error. In general, errors only appear on required fields if bad data is filled in it, or the user navigates away from it and returns (e.g., sub-section review). 

* When entering APO as primary address, I can't enter in anything for physical address. Fields are marked as red with an error to provide data, but I can't type anything into the fields. I also can't select the "Outside of the United States" button. Here's what that looks like:

screen shot 2018-11-01 at 10 17 28 pm

(macOS 10.12, Chrome 69.0.3497.100)

* The USPS address validation gets triggered when APO is specified in primary address, but not in secondary. I'm not convinced that the USPS validation API works at all for APO/FPO/DPO addresses, so the true defect may be that it gets called for the primary address in the APO, etc. scenario.

* Once all the functionality is working, the question text/field display names will have to be tweaked to match exactly what's on the SF-86 form (or get official policy sign off if a deviation is warranted). Just to clarify the terminology, an APO/FPO address is not considered a physical address.

el-mapache:
@ryanhofdotgov Thanks for checking this out! I'm going to address each of these comments here:

* `HasDifferentAddress` is actually a property of the `PhysicalAddress` data structure, which is a different address component that seems to be explicitly tied to the `Employment` set of components (specifically the self-employed options, iirc). I used it because it is self contained, as you mentioned, and because this specific instance of a branch is not exactly the same as most of the other uses of branches in the app, which are tied to `List`s.

* I know the fields are immediately marked as invalid, and I beleive it's because I don't do any of the error handling that most other components have cop/pasted across them. I would like to make the argument that we should just merge this with the behavior, and I can address it in a followup PR, before we actually push code to our staging server.

* The fourth point is a bug, I'll write a test for it + fix

* I actually omitted the address validation on the secondary APO fields, because I thought it wasn't working correctly...I can re-add if necessary

* I think most of the language is the correct barring the 'physical address' language on the secondary APO field. I tried to copy the wording exactly from the form snippets in the issue. I think this can also be part of a final cleanup PR later on, though.

ryanhofdotgov:
@el-mapache Sweet – I agree on all points. 

No need to add USPS validation for a secondary address that is an APO – I'll open up an issue to verify/remove that from the primary address field set.

el-mapache:
@ryanhofdotgov Should be ready for another look

ryanhofdotgov:
With #924 split off into #1107, #1108, and #1109, this PR closes #924

PR #1089, opened 2018-11-01
Fix mandatory counseling for not been ordered

### LINKED ISSUE #1047: For mandatory counseling, don't allow "none of the above" simultaneously with other options
For the "illegal use of drugs and drug activity" there is a section for mandatory counseling/treatment:

screen shot 2018-10-22 at 1 43 06 am

The UI is a weird mix of "check all that apply" but with one of the 5 options mutually exclusive with the other 4: "I have not been ordered... any of the above".

The applicant should not be permitted to select that option along with the other 4. 

In the section for alcohol, there is a similar checklist. However it has some kind of JavaScript to conditionally uncheck options:

screen shot 2018-10-22 at 1 43 28 am

line47:
Fixes #1047 

Adding functionality to toggle the buttons off when `I have not been ordered, advised, or asked to seek counseling or treatment by any of the above.` has been selected

The UI here is really confusing and could use some thought.

click

### DIFF
--- a/src/components/Section/SubstanceUse/Drugs/OrderedTreatment.jsx
+++ b/src/components/Section/SubstanceUse/Drugs/OrderedTreatment.jsx
@@ -66,11 +66,20 @@ export default class OrderedTreatment extends ValidationElement {
     let selected = cb.value
     let list = [...((this.props.OrderedBy || {}).values || [])]
 
-    if (list.includes(selected)) {
-      list.splice(list.indexOf(selected), 1)
-    } else {
-      list.push(selected)
-    }
+      if (list.includes(selected)) {
+        list.splice(list.indexOf(selected), 1)
+      } else {
+        if (selected !== "None" && list.includes("None")) {
+          list.splice(list.indexOf("None"), 1)
+          list.push(selected)
+        }
+        else if (selected === "None") {
+          list = [selected]
+        }
+        else {
+          list.push(selected)
+        }
+      }
 
     this.update({ OrderedBy: { values: list } })
   }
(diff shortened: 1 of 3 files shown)


line47:
CC @joshmfranklin about a design for these type of situations

src/components/Section/SubstanceUse/Drugs/OrderedTreatment.jsx
```diff
@@ -66,11 +66,20 @@ export default class OrderedTreatment extends ValidationElement {
...
 
-    if (list.includes(selected)) {
-      list.splice(list.indexOf(selected), 1)
-    } else {
-      list.push(selected)
-    }
+      if (list.includes(selected)) {
+        list.splice(list.indexOf(selected), 1)
+      } else {
+        if (selected !== "None" && list.includes("None")) {
```
el-mapache: Based on what I'm seeing in this code, it seems like a list with 'None' in it will never have any other values. If that is the case, I think we can probably combine the `if/else` statement here and put
something like:
```
if ((selected !== "None" && list.includes("None")) || selected === 'None') {
  list = [selected]
}
```

src/validators/drugorderedtreatments.js
```diff
@@ -76,6 +76,18 @@ export class DrugOrderedTreatmentValidator {
...
-    return validGenericTextfield(this.explanation) && this.validActionTaken()
+    return validGenericTextfield(this.explanation) && this.validActionTaken() && this.validOrderedBy()
+  }
+
+  validOrderedBy() {
+    if (this.orderedBy.length <= 1) {
+      return true
+    }
+    else {
+      if (this.orderedBy.includes("None")) {
```
el-mapache: Given the code in the component, will this execute? It seems like the list of options can either have multiple selections that are not 'None', or 'None'

src/components/Section/SubstanceUse/Drugs/OrderedTreatment.jsx
line47: Thanks!

src/validators/drugorderedtreatments.js
line47: Yes that right, looking back at it that else block won't run

Issue #1055, opened 2018-10-23
Tax year not persisting


hursey013:
**What did you do?**

Navigate to Financial > Taxes. Enter a year into the `Provide the year you failed to file or pay your federal, state, or other taxes (Estimated)` field. Logout and login.

**What did you expect to see?**

The date to be populated in the `Provide the year you failed to file or pay your federal, state, or other taxes (Estimated)` field.

**What did you see instead?**

No year
<img width="678" alt="screen shot 2018-10-23 at 1 29 47 pm" src="https://user-images.githubusercontent.com/1178494/47379097-f4e0a880-d6c7-11e8-8c82-1ce064c0b654.png">

PR #1092, opened 2018-11-01
Fix tax year increment bug

### LINKED ISSUE #1071: Tax year incremeented in test case #7
From NBIB testing team:

Description:
Section 26: Financial Record
Provide the year you failed to file or pay your Federal, state, or other taxes: eAPP displayed 2014 while it's translated to 2015 in eQIP

christopherhuii:
Fixes #1071 

In this PR, I have verified that test case 7 displays the correct year and starting a new form also displays the correct year.

Because `DateControl.jsx` is a heavily used component throughout the application, I am mildly nervous about the changes breaking something else. I did a bit a regression, and haven't seen any red flags yet.

The issue with the code was:
```
var date = new Date("2015");
// Wed Dec 31 2014 16:00:00 GMT-0800 (Pacific Standard Time)

date.getFullYear();
// 2014

It should be 2015.
```
Instead, I manually use `setYear` on the date in the switch statement. I am hesitant to refactor the component more because other bugs that might be introduced.

### DIFF
--- a/src/components/Form/DateControl/DateControl.jsx
+++ b/src/components/Form/DateControl/DateControl.jsx
@@ -38,6 +38,7 @@ export const datePart = (part, date) => {
     case 'year':
     case 'yy':
     case 'y':
+      d.setYear(date)
       return d.getFullYear()
   }
 
--- a/src/components/Section/Financial/Taxes/TaxesItem.jsx
+++ b/src/components/Section/Financial/Taxes/TaxesItem.jsx
@@ -139,6 +139,7 @@ export default class TaxesItem extends ValidationElement {
           <DateControl
             name="Year"
             {...this.props.Year}
+            year={this.props.Year.year || this.props.Year.value}
             className="taxes-year"
             hideMonth={true}
             hideDay={true}
--- a/src/components/Section/Financial/Taxes/TaxesItem.jsx
+++ b/src/components/Section/Financial/Taxes/TaxesItem.jsx
@@ -139,7 +139,6 @@ export default class TaxesItem extends ValidationElement {
           <DateControl
             name="Year"
             {...this.props.Year}
-            year={this.props.Year.year || this.props.Year.value}
             className="taxes-year"
             hideMonth={true}
             hideDay={true}


ryanhofdotgov:
@christopherhuii It looks like this might just be the tip of the iceberg unfortunately:
From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
"Note: parsing of date strings with the Date constructor (and Date.parse, they are equivalent) is strongly discouraged due to browser differences and inconsistencies. Support for RFC 2822 format strings is by convention only. Support for ISO 8601 formats differs in that date-only strings (e.g. "1970-01-01") are treated as UTC, not local."

ryanhofdotgov:
Actually, not that there still isn't an iceberg here, but Date.parse() doesn't appear to be the problem – as it doesn't look like date objects are being built the way you show in your example above.

ryanhofdotgov:
@christopherhuii After not being able to get `datePart` to be called like `datePart('y', '2015')` after reloading the data from the db, and was only able to get it called when loading `test7.json`, I looked into the JSON structure a bit more closely.

The work @hursey013  did in https://github.com/18F/e-QIP-prototype/issues/1055 changed the JSON structure, so the code and `test7.json` no longer matched. Here's a diff where old version is what is in repo and new version is what I generated from current version of app:

```
                   "Year": {
-                    "type": "number",
+                    "type": "datecontrol",
                     "props": {
-                      "value": "2015"
+                      "month": "1",
+                      "day": "1",
+                      "year": "2015",
+                      "estimated": false
                     }
```

This a big failing with these coarse scenario-based tests – they have to be updated as the code changes and are just stand-ins for more in-depth unit testing.

So:
1) I don't think we need this PR – we need to just fix the serialized test case to match what the code expects
2) `datePart` is still sketchy, but I haven't seen it called in the app with anything except the empty string as the value for `date`. In other words, we should re-write it to be more defensive and not call APIs that differ from browser to browser, but it doesn't seem to be the potentially critical iceberg I originally thought.

Thoughts?

christopherhuii:
I see. This makes sense why the test case was only loading an object with a key of `value` instead of an object with all the `month`, `date`, `year`, etc. keys.

```
      month: props.hideMonth ? '1' : props.month || datePart('m', props.value),
      day: props.hideDay ? '1' : props.day || datePart('d', props.value),
      year: props.year || datePart('y', props.value)
```
The part I'm still unsure about is what `props.value` is in `DateControl.jsx`. If it's always a dash separated date (e.g. 1-1-2016), we can:

```
const dateAttributes = props.value.split('-');
// (3) ["1", "1", "2016"]

const month = dateAttributes[0];
const day = dateAttributes[1];
const year = dateAttributes[2];

new Date(year, month - 1, day);
// Fri Jan 01 2016 00:00:00 GMT-0800 (Pacific Standard Time)
```

ryanhofdotgov:
@christopherhuii I added console logging during a testing session yesterday to print whenever datePart was called.

Except for the bogus test7.json data, it was only called with null? empty string?props.value

So the only place I've seen it set so far was in that test file.

christopherhuii:
@ryanhofdotgov Yeah, it's weird that `datePart` is exported, but it's not being used elsewhere. `datePart` is only used when `props.year`, `props.month`, or `props.day` is falsy. It's not clear to me when those can be falsy. The tests to indicate that it's expect to follow the pattern of a dash separated date though.

Issue #185, opened 2018-05-23
Form wide - Change all links to the underline style 


ryanhofdotgov:
From e-QIP-prototype-truetandem created by [Jay-Neighbours-I72](https://github.com/Jay-Neighbours-I72) : ryanhofdotgov/e-QIP-prototype-truetandem#2715  

Currently we are using 3 different dashed style for links. Let's use the typical underlined style for all of them to be consistent.

![screen shot 2018-03-08 at 10 14 44 am](https://user-images.githubusercontent.com/19935974/37158562-a802b886-22b9-11e8-95f4-6e98bb3455ca.jpg)

christopherhuii:
@ryanhofdotgov @line47 
> Let's use the typical underlined style for all of them to be consistent.

Is the standard underlined style a 1px solid underline?

line47:
it is `text-decoration: underline;` which is the default from the USWDS

PR #1102, opened 2018-11-05
Updates to make links standard underline style for the entire app

line47:
Fixes https://github.com/18F/e-QIP-prototype/issues/185 

screen shot 2018-11-02 at 5 06 18 pm
screen shot 2018-11-02 at 5 06 28 pm
screen shot 2018-11-02 at 5 06 37 pm
screen shot 2018-11-02 at 5 07 25 pm


line47:
Yeah I'm still having issues with viewing IE locally (long story), @christopherhuii if you wouldn't mind taking a look in IE11 that would be 💯

christopherhuii:
I will review in IE11. Thanks

christopherhuii:
Doing a first pass in Chrome:

When hovered, most fields apply the blue text color and solid underline.

In "Information about you" -> "Your identifying information", the "Remove comment" link is blue and underlined before the hover. When hovered, the text color turns to black and is still hovered.

screen shot 2018-11-07 at 9 09 28 am

christopherhuii:
Similarly, in "Your history" -> "Where you went to school", have the underline by default, the text color is blue, and does not change on hover. It's probably also worth mentioning that the underline is black in this context.

christopherhuii:
Are focus states around input fields still dashed?

If my previous comments are OK, then the PR is good to go. Once I get confirmation that my previous comments are intentional and as designed, I will approve.

line47:
Thanks @christopherhuii - those are some additional issues to work through 👍

line47:
@christopherhuii The links should be underlined by default (before hover and also when hovering)  they should have a dashed focus state. 

Seems like the only thing to fix is the black text on hover which I'm not seeing in Chrome on mac.

christopherhuii:
@line47 Sorry, I should have been more clear. It goes from light blue to darker blue.

I wasn't sure if it's intentional because the accordions go from black to light blue. If this is all intentional, then it looks good. Everything is consistent from Chrome to IE11

Issue #1108, opened 2018-11-06
Add secondary address workflow to residence verifier address collection


ryanhofdotgov:
Incorporate the secondary address workflow implemented for the residence address (#924) into the form where the residence verifier's address is collected.

Issue #1107, opened 2018-11-06
Add secondary address workflow to un/self-employment verifier address collection


ryanhofdotgov:
Incorporate the secondary address workflow implemented for the residence address (#924) into the form where the unemployment/self-employment verifier's address is collected.

PR #1113, opened 2018-11-06
Encapsulate alternate address behavior

el-mapache:
Completes #1107, #1108

* Should make it easier to reuse this functionality in the code
* Adds `AlternateAddress` to employment and the reference address section of the `History` subsection

### DIFF
--- a/src/components/Form/Location/AlternateAddress.jsx
+++ b/src/components/Form/Location/AlternateAddress.jsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import { connect } from 'react-redux';
 import { i18n } from '../../../config'
 import Field from '../Field'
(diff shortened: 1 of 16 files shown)


ryanhofdotgov:
@el-mapache Still poking it, but here is one data point:

If I add a self-employed entry, and enter an "Outside of the US" address for verifier, and then click "No" to following APO question, then:
* APO question disappears
* Address field set is cleared and set to "In the United States"

screen shot 2018-11-06 at 3 42 40 pm

ryanhofdotgov:
Also:
* Same thing happens with unemployment. And also if you click "Yes" instead of "No" to APO question
* Can't enter in an APO address for primary verifier address – clicking on the APO/FPO button leaves (or resets) the address to the domestic field set.

But the residence verifier workflow looks good :+1:

el-mapache:
Thanks, working on fixing now!

el-mapache:
@ryanhofdotgov I think this should be good to go, thanks again for reviewing

ryanhofdotgov:
I'm seeing an odd problem, but have split if off into https://github.com/18F/e-QIP-prototype/issues/1117

PR #1114, opened 2018-11-06
Show AlternateAddress component in the military employment 'supervisor' block

### LINKED ISSUE #1109: Add secondary address workflow to supervisor verifier address collection
Incorporate the secondary address workflow implemented for the residence address (#924) into the form where the supervisor's address is collected.

~Note that for active duty/national guard/etc. employment, the Supervisor's address is requested slightly differently. Only the APO --> Physical branch is enabled, the Foreign --> APO branch of the workflow should not be enabled.~ Edit: While the PDF has this deviation, e-QIP does not and that overrides the PDF in this case.

See the SF-86:
screen shot 2018-10-18 at 8 24 34 pm

el-mapache:
Fixes #1109

### DIFF
--- a/src/components/Section/History/Employment/EmploymentItem.jsx
+++ b/src/components/Section/History/Employment/EmploymentItem.jsx
@@ -438,6 +438,13 @@ export default class EmploymentItem extends ValidationElement {
             required={this.props.required}
             scrollIntoView={this.props.scrollIntoView}
           />
+          {this.props.render({
+            address: this.props.SupervisorAlternateAddress,
+            allowForeignMilitary: false,
+            belongingTo: 'SupervisorAlternateAddress',
+            country: this.props.Supervisor.Address.country,
+            onUpdate: this.update
+          })}
         </Show>
 
         <Show when={this.showReference()}>
@@ -571,7 +578,7 @@ EmploymentItem.defaultProps = {
   Status: {},
   Address: {},
   Telephone: {},
-  Supervisor: {},
+  Supervisor: { Address: {} },
   ReferenceName: {},
   ReferencePhone: {},
   ReferenceAddress: {},
--- a/src/components/Section/History/Employment/EmploymentItem.jsx
+++ b/src/components/Section/History/Employment/EmploymentItem.jsx
@@ -440,7 +440,6 @@ export default class EmploymentItem extends ValidationElement {
           />
           {this.props.render({
             address: this.props.SupervisorAlternateAddress,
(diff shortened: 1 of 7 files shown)


el-mapache:
Because this branch was cut from `ab-924-employment` I'm leaving as a WIP until #1113 is merged

hursey013:
I have been out of the loop on the alternative address work, so I could be misunderstanding it, but https://github.com/18F/e-QIP-prototype/issues/1109 indicates that there should be a different flow based on whether `active duty/national guard/etc` is selected or not - is that happening in this?

ryanhofdotgov:
@hursey013 I'm looking at this now... I think like one of the other recent tickets, this one is different in the PDF, but not different in e-QIP. i.e., it is one of those rare places where e-QIP and the PDF don't match exactly. I think this PR does what we need though (i.e., what e-QIP expects).

ryanhofdotgov:
Yeah, it looks like there is no deviation in secondary address workflow here after all. 

e-QIP makes this consistent with the previous ones, so lessoned learned to me – don't count on PDF/e-QIP to match, always verify before writing the issue description.

src/components/Section/History/Employment/EmploymentItem.jsx
```diff
@@ -453,6 +453,13 @@ export default class EmploymentItem extends ValidationElement {
             required={this.props.required}
             scrollIntoView={this.props.scrollIntoView}
           />
+          {this.props.render({
+            address: this.props.SupervisorAlternateAddress,
+            allowForeignMilitary: false,
```
ryanhofdotgov: I guess I didn't look at this carefully in the previous PRs, but I'm not following the workflow characteristic that this controls. Flipping it back and forth from false/true didn't seem to have an effect on the supervisor primary/secondary address UI.

el-mapache: the `allowForeignMilitary` switch should give us finer-grained control over whether or not we show the branch option in this component. So, it would be a bug if we actually wanted to conditionally show the branch!

It sounds like we can just remove it based on the comment above?

ryanhofdotgov: @el-mapache Yeah, I wasn't saying it being false was a bug – I was just saying that I played with the UI with it set to false, and then set it to true, re-built the app, and didn't notice any behavior change. i.e., the behavior I was expecting by a quick read of the code wasn't appearing to happen in the UI.

If you've removed it at this point, it is moot though :)

PR #1135, opened 2018-11-09
Add AlternateAddress component to CivilUnion

### LINKED ISSUE #1029: Ask about current spouse's APO/FPO
This is similar, but not exact, as: #753 and #924 

First collect either a US address or a foreign address. Do not collect a APO address initially.

Regardless of the current address, always ask "Does the person have an APO/FPO address within the United States?". The only options are Yes/No. 

If Yes, display and prompt for APO address and also a physical address.

screen shot 2018-10-18 at 8 28 01 pm

The scope of this ticket does not include encoding this data correctly into e-QIP XML – thats addressed in https://github.com/18F/e-QIP-prototype/issues/1030

el-mapache:
Fixes #1029

### DIFF
--- a/src/components/Form/Location/AlternateAddress.jsx
+++ b/src/components/Form/Location/AlternateAddress.jsx
@@ -13,8 +13,12 @@ import alternateAddress from '../../../schema/form/alternateaddress'
 const alternateAddressDefaultState = alternateAddress
 const propTypes = {
   address: PropTypes.object,
+  addressBook: PropTypes.string,
+  allowForeignMilitary: PropTypes.bool,
   belongingTo: PropTypes.string,
   country: PropTypes.string,
+  // XXX Rename this prop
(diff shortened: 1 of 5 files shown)


el-mapache:
The drop in code coverage here is due to a test I skipped; it was breaking but not actually testing anything useful

ryanhofdotgov:
@el-mapache When you are filling out the address for the first time for a spouse, it hides the address fields:

screen shot 2018-11-13 at 2 37 57 pm

If you toggle "Use current address" on and off it will show them then. 

Subjectively, I'm not sure if the new show/hide paradigm is ideal for this instance, but want to give others the opportunity to weigh in on this implementation once its on `develop`. So just looking for a changeset that will show the address field set in the initial set. Thanks.

el-mapache:
@ryanhofdotgov Updated, apologies

Issue #1048, opened 2018-10-22
Year tax was failed to be filed/paid can not be after the date settled


ryanhofdotgov:
Year tax was failed to be filed/paid can not be after the date settled. Edit: per comments below, logic is there, but error message is not correct.

line47:
I think this is part of https://github.com/18F/e-QIP-prototype/issues/1000

line47:
I think this was fixed in PR https://github.com/18F/e-QIP-prototype/pull/929

line47:
The validation is working The error message could use some updating though.
applicant birth date on this is 6/9/1979 (so both dates are after DOB)

![screen shot 2018-11-08 at 1 16 42 pm](https://user-images.githubusercontent.com/1449852/48218764-9dde0300-e358-11e8-8baf-4475a90edb88.png)

Issue #1046, opened 2018-10-22
For Drug use purchase: most recent involvement date can't be before 1st involvement


ryanhofdotgov:
Placeholder

line47:
Fixed in PR https://github.com/18F/e-QIP-prototype/pull/929

line47:
The validation is working The error message could use some updating though.
applicant birth date on this is 6/9/1979 (so both dates are after DOB)

![screen shot 2018-11-08 at 1 27 23 pm](https://user-images.githubusercontent.com/1449852/48219277-15f8f880-e35a-11e8-81b3-0e4e9b50614b.png)

PR #1136, opened 2018-11-09
Update error messages

line47:
Update error message to be descriptive about the relationship between dates.

Fixes https://github.com/18F/e-QIP-prototype/issues/1048
Fixes https://github.com/18F/e-QIP-prototype/issues/1046

screen shot 2018-11-09 at 8 03 27 am
screen shot 2018-11-09 at 8 23 00 am

### DIFF
--- a/src/components/Section/Financial/Taxes/TaxesItem.jsx
+++ b/src/components/Section/Financial/Taxes/TaxesItem.jsx
@@ -247,6 +247,7 @@ export default class TaxesItem extends ValidationElement {
               className="taxes-date"
               hideDay={true}
               minDate={(this.props.Year || {}).date}
+              prefix="taxesSatisfied"
               minDateEqualTo={true}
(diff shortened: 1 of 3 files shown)


line47:
I do have a question about how we can show an error when these fields are before the DOB?
Right now they always show the error in the screenshots that are attached. Would I just add to the current `min` message something like `The date satisfied is before the date you failed to file or pay or before your date of birth.` ?

hursey013:
Yeah I don't think we currently have a great way of working with those types of situations. The only thing I can think of is to recreate the mindate validator functionality in the component and have have it conditionally change the `prefix` field to point to either a birthdate or other date i18n string based on the results. Similar to this, but would need to be a little more involved: https://github.com/18F/e-QIP-prototype/blob/develop/src/components/Section/Relationships/Relatives/Relative.jsx#L544

Issue #582, opened 2018-07-18
 APO/FPO/DPO bug


bpdesigns:
**What did you do?**
Clicked on APO/FPO/DPO for address block

**What did you expect to see?**
 Should have turned the button green

**What did you see instead?**
the button did not change but the content for that type of address block showed up.
**Additional Information**

Also the spacing is weird in the buttons on Mac/Safari 
<img width="654" alt="screen shot 2018-07-18 at 9 40 13 am" src="https://user-images.githubusercontent.com/15329140/42885555-4877ffca-8a6f-11e8-805c-d9ba0af04661.png">

and the spacing is too narrow above "select  APO/FPO/DPO" text on both Mac/Safari and Mac/Chrome
<img width="597" alt="screen shot 2018-07-18 at 9 36 20 am" src="https://user-images.githubusercontent.com/15329140/42885570-4fdb91b4-8a6f-11e8-866f-41beaccb5835.png">

bpdesigns:
this issue relates to  #147 "Previously used APO address - not saving APO selection"

bpdesigns:
#147 relates to the selection of the button turning it green. This issue relates specifically to the spacing. I see my original comments are unclear.

Issue #147, opened 2018-05-23
Previously used APO address - not saving APO selection


ryanhofdotgov:
From e-QIP-prototype-truetandem created by [Jay-Neighbours-I72](https://github.com/Jay-Neighbours-I72) : ryanhofdotgov/e-QIP-prototype-truetandem#2871  

- [ ] When using a Previously used APO address the initial selection of APO/FPO/DPO didn't come through.
- [ ] Can't reselect the APO/FPO/DPO button

![screen shot 2018-04-18 at 2 42 53 pm](https://user-images.githubusercontent.com/19935974/38951530-ec35572a-4316-11e8-92ce-de46dacd3497.jpg)

PR #1137, opened 2018-11-09
Properly style APO address type radio button

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/582
Fixes https://github.com/18F/e-QIP-prototype/issues/147 (I think?)

Due to the way the APO option was nested in a `Show` component, the logic to apply the `checked` CSS class, along with some of the proper styles were not getting applied correctly. When looking into this I also noticed some extra markup and unnecessary CSS that I was able to remove.

There are several instances of our CSS (and javascript) that rely on a direct parent/child relationship (sibling selectors, table cells, flexbox, etc). The `Show` component sometimes requires the component it's wrapping to be inside a container `<span>` which throws that relationship off. React 16 introduced `<React.Fragment>` which will avoid this issue when we are able to upgrade.

**After:**
screen shot 2018-11-09 at 10 55 53 am

### DIFF
--- a/src/components/Form/Location/Address.jsx
+++ b/src/components/Form/Location/Address.jsx
@@ -12,7 +12,10 @@ import ApoFpo from '../ApoFpo'
 import Show from '../Show'
 import Suggestions from '../Suggestions'
 import { AddressSuggestion } from './AddressSuggestion'
-import LocationValidator, { countryString, isInternational } from '../../../validators/location'
+import LocationValidator, {
+  countryString,
+  isInternational
+} from '../../../validators/location'
 import { countryValueResolver } from './Location'
 
 export default class Address extends ValidationElement {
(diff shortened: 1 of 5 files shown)


line47:
I noticed when I toggle `APO/FPO/DPO` on and then off the button highlights `Outside of the US` would we want it to default to `Inside the US`?

apo

ryanhofdotgov:
Should these buttons even toggle? Seems like they should be straight-up radio style buttons.

line47:
Good point @ryanhofdotgov and I agree with that - I'm unsure how far down the rabbit hole you'd need to go to make that adjustment. The component being used is the `RadioGroup`

ryanhofdotgov:
OK – I would just leave things as are for this PR, and open up a different issue to make these behave more like traditional radio buttons.

hursey013:
@ryanhofdotgov I can reproduce what @line47 is seeing - looks like its a preexisting issue and looks a little more involved than a quick fix so I opened a new issue for it here https://github.com/18F/e-QIP-prototype/issues/1140

PR #1142, opened 2018-11-13
[WIP] - Radio and checkbox styling

line47:
This was some exploratory work I previously started where I was exploring stripping away the custom styling of checkboxes and radio buttons to return them to a more recognized state.
See https://codepen.io/jfrank7588/pen/xyNJMq for styling direction

- Don't hide the radio or check.
- Remove images associated with radios/checks as they were cluttering up the page and increasing the cognitive load for users.

### This work is up to Section 6: Foreign associations

**What I found along the way*
- label text sometimes used `i18n.m` and other times used `i18n.t` The `i18n.m` rendering adds a `<p>` around the text which was causing styling issues that were then overridden rather than consistently used. `Recommend using i18n.t`
- There are a lot of custom styles used for certain radio/check sections that are unique and not needed. This can be built in a more general and adaptable way. For example creating a class that shows 3 or 4 radios per row would be a better approach than creating a class targeting residenceItems.

**_testing in IE still needs to happen._** 

### Screenshots

screen shot 2018-11-13 at 9 10 04 am
screen shot 2018-11-13 at 9 10 14 am
screen shot 2018-11-13 at 9 10 23 am
screen shot 2018-11-13 at 9 10 56 am
screen shot 2018-11-13 at 9 11 18 am
screen shot 2018-11-13 at 9 11 30 am
screen shot 2018-11-13 at 9 12 11 am
screen shot 2018-11-13 at 9 12 23 am


hursey013:
@joshmfranklin with a lot of these design choices codified into the style guide, would it be ok to close this?

joshmfranklin:
@hursey013 yep makes sense to me.

PR #1152, opened 2018-11-14
[DON'T MERGE] Add codecov back in when ready 

line47:
Add coverage back in once the team decides to start using again after API coverage is at a comfortable level.

This is a placeholder PR so this step is not forgotten about. 🚀


ryanhofdotgov:
@line47 Can you make this an issue instead? You can reference the commit where you turned coverage off, or show the small diff directly in a comment.

PR #1163, opened 2018-11-16
Align form headings with exact text in SF-86

### LINKED ISSUE #1116: Tweak display text related to secondary addresses
Split off of #924 

Need to review:
* The prompt text for APO and for foreign address
 * `Does the person have an APO/FPO address within the United States?`
 * `Provide physical location data with street address, base, post, embassy, unit, and country location or home port/ fleet headquarter.`
* input field display names for physical address (e.g., `Street Address/Unit/Duty Location` and `City or Post Name`

ryanhofdotgov:
Closes #1116. Sole exception is address.apoFpoRequired, which does not have an exact correspondence in the SF-86.

### DIFF
--- a/src/components/Section/History/Employment/EmploymentItem.jsx
+++ b/src/components/Section/History/Employment/EmploymentItem.jsx
@@ -442,6 +442,7 @@ export default class EmploymentItem extends ValidationElement {
             address: this.props.SupervisorAlternateAddress,
             belongingTo: 'SupervisorAlternateAddress',
             country: this.props.Supervisor.Address.country,
+            militaryAddressLabel: i18n.t('address.militaryAddress.supervisor'),
             onUpdate: this.update
           })}
         </Show>
@@ -516,6 +517,7 @@ export default class EmploymentItem extends ValidationElement {
               address: this.props.ReferenceAlternateAddress,
               belongingTo: 'ReferenceAlternateAddress',
               country: this.props.ReferenceAddress.country,
+              militaryAddressLabel: i18n.t(`${prefix}.heading.militaryAddress`),
               onUpdate: this.update
             })}
           </div>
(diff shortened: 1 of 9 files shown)


src/config/locales/en/history.js
```diff
@@ -1355,6 +1355,7 @@ export const history = {
           'Provide the name of someone that can verify your self-employment',
         physicalAddress:
           'Is your physical work address different than your employment address?',
+        militaryAddress: 'Does your self-employment verifier have an APO/FPO address?',
```
christopherhuii: @ryanhofdotgov I can't seem to find this text being used anywhere in eApp

src/config/locales/en/history.js
```diff
@@ -1458,6 +1459,7 @@ export const history = {
     },
     unemployment: {
       heading: {
+        militaryAddress: 'Does your unemployment verifier have an APO/FPO address?',
```
christopherhuii: @ryanhofdotgov I can't seem to find this text being used anywhere in eApp. Should it be used? (same thing with previous comment)

src/components/Section/History/Employment/EmploymentItem.jsx
```diff
@@ -516,6 +517,7 @@ export default class EmploymentItem extends ValidationElement {
               address: this.props.ReferenceAlternateAddress,
               belongingTo: 'ReferenceAlternateAddress',
               country: this.props.ReferenceAddress.country,
+              militaryAddressLabel: i18n.t(`${prefix}.heading.militaryAddress`),
```
ryanhofdotgov: @christopherhuii Here's where those 2 text strings get inserted. This section has a whole bunch of text that is similar, but different between self/unemployment so they group them under common subpath and use `prefix` to distinguish, which is set to self/unemployment specific parent paths.

ryanhofdotgov:
@christopherhuii Thanks for the catch on the case. I think for little tweaks like this, I'd be interested in folks trying out the recent "Insert suggestion" feature: https://help.github.com/articles/commenting-on-a-pull-request/#adding-line-comments-to-a-pull-request

PR #1168, opened 2018-11-16
Fix sponsorship start date needs to be after DOB

### LINKED ISSUE #1044: Birthdate of sponsored person needs to be before start of foreign sponsorship
Placeholder

christopherhuii:
Fixes #1044 

Just needed to add the `minProp` so that the component knows what the minimum date allowed is.

### DIFF
--- a/src/components/Section/Foreign/Business/SponsorshipItem.jsx
+++ b/src/components/Section/Foreign/Business/SponsorshipItem.jsx
@@ -309,6 +309,7 @@ export default class SponsorshipItem extends ValidationElement {
             name="Dates"
             {...this.props.Dates}
             onUpdate={this.updateDates}
+            minDate={this.props.Birthdate && this.props.Birthdate.date}
             minDateEqualTo={true}
             onError={this.props.onError}
             className="foreign-business-sponsorship-dates"


hursey013:
This likely runs into the issue I'm working here https://github.com/18F/e-QIP-prototype/pull/1171 where the `date` prop isn't available after you log out and back in. It turns out that `date` prop is only set after a `DateControl` component is interacted with, so in this case `this.props.Birthdate.date` won't exist until you click into the person's birth date field.

christopherhuii:
@hursey013 Is there a known way to do this? There's nothing immediately obvious. to be able to swap out the error

ryanhofdotgov:
@christopherhuii Checkout `./components/Section/Relationships/RelationshipStatus/Divorce.jsx` where `minDate` is set and look at `prefix` in `DateControl.jsx`. It expects certain paths to exist after the prefix for certain errors.

hursey013:
@christopherhuii using the `prefix` prop will allow you to reference a new key in the `error` i18n file that you can add your custom messages to. Because of the way it's currently configured, you need to include a messages for `required` and another other possible errors for the new key. You can see a  recent addition from @line47 here for an example: https://github.com/18F/e-QIP-prototype/blob/develop/src/config/locales/en/error.js#L1129-L1138

christopherhuii:
@hursey013 updated the error message

src/config/locales/en/error.js
```diff
@@ -497,6 +497,12 @@ export const error = {
           message:
             "For the **from** date, the date should be after this person's date of birth."
         }
+      },
+      foreignNationalSponsorship: {
+        min: {
+          title: 'There is a problem with your entry',
+          message: 'For the **from** date, the date should be after their date of birth.'
+        }
```
hursey013: When you add a prefix, it prefixes _all_ of the error messages associated with that component unfortunately, so adding in a `required` message as well in case the date field is left blank.
```suggestion
        },
        required: {
          title: 'Your response is required',
          message: ''
       }
```

christopherhuii:
doh! thanks @hursey013

Issue #1165, opened 2018-11-16
eApp archival PDF doesn't show Employer 1 Reason for leaving in test case #9


ryanhofdotgov:
**What did you do?**

Load up test case #9, Submit, generate and save eApp archival PDF.

**What did you expect to see?**

Reason for leaving employer 1 in the archival PDF ("Employer 1 Reason").

**What did you see instead?**

No reason shown.

**Additional Information**

Might be related to work done in this section for showing/hiding things correctly based on number of years involved. Data is correctly generated in the XML sent to e-QIP.

ryanhofdotgov:
Test case 9 isn't merged into `develop` yet, but can be downloaded here:
https://github.com/18F/e-QIP-prototype/blob/rmh-alternate-address/api/testdata/complete-scenarios/test9.json

ryanhofdotgov:
I can replicate the issue with test case 1. Like test case 9, it doesn't show these questions:
* Have you attended any schools in the last 10 years?
* Have you received a degree or diploma more than 10 years ago?
(and doesn't show the Where you went to school/List the places you went to school headings at all)

Looking at my archive of test case archival PDFs, it seems to be a regression that slipped in between October 1st and November 1st.

PR #1170, opened 2018-11-16
Show reason for leaving, not just within 7 years

ryanhofdotgov:
ReasonLeft was recently fixed to appropriately show/hide fields in https://github.com/18F/e-QIP-prototype/pull/1100. However, there was redundant and incorrect logic at the EmploymentItem level that this removes.


ryanhofdotgov:
Fixes #1165

PR #1181, opened 2018-11-17
Add edu/degree Qs to history section print view

### LINKED ISSUE #1164: eApp archival PDF doesn't show test case #9 Education
**What did you do?**

Load up test case #9, which has no education listed – just "No" answers to the 2 questions. Submit. Save eApp archival PDF using the button on the submit page.

Education section printed in the PDF, with "No" to the questions listed.

**What did you see instead?**

No Education questions printed in the PDF.

**Additional Information**

This might be related to the recent work to hide/show certain things in the PDF.  Data is correctly generated in the XML sent to e-QIP.

ryanhofdotgov:
Fixes #1164

### DIFF
--- a/src/components/Section/History/History.jsx
+++ b/src/components/Section/History/History.jsx
@@ -756,6 +756,20 @@ export class HistorySections extends React.Component {
           required={true}
         />
 
+        <Branch
+          name="branch_school"
+          {...this.props.Education.HasAttended}
+          label={i18n.t('history.education.label.attendance')}
+          labelSize="h3"
+        />
+        <Show when={this.props.Education.HasAttended.value === 'No'}>
+          <Branch
+            name="branch_degree10"
+            {...this.props.Education.HasDegree10}
+            label={i18n.t('history.education.label.degree10')}
+            labelSize="h3"
+          />
+        </Show>
         <Show
           when={
             this.props.Education.HasAttended.value === 'Yes' ||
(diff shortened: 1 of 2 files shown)


gidjin:
Is it possible to include a test for existence of the new component in the History.test.jsx as well?

ryanhofdotgov:
@gidjin Not that long term we couldn't add a whole new suite of printing-related tests covering the PDF, but the tests in `History.test.jsx` doesn't test anything related to the print view. Just what is displayed/manipulated in the browser view.

For this printing/PDF change, no new components were created, just existing ones re-declared over from the existing browser view to the print view, as is the (anti-)pattern for all the sections right now (until https://github.com/18F/e-QIP-prototype/issues/614 is resolved).

So right now, the only thing representing what the print view should look like during our test suite is the snapshot `Print.test.jsx.snap`, which now correctly shows the inclusion of these 2 questions. 

In short, it's not appropriate to build out a whole new PDF/printing-centric test suite for this change. But if we already had one outside of the snapshot, it would definitely be appropriate to add/change the relevant bits.

For more context, we also need to dramatically change how we create this PDF, but there's some ongoing discussion as to how much time we should spend on it, and whether the functionality belongs outside of eApp:
https://github.com/18F/e-QIP-prototype/issues/373

gidjin:
That makes sense regarding the testing. Where can I find test case 9? I don't see it on api/testdata/complete-scenarios

gidjin:
That makes sense regarding the testing. Where can I get test case 9 I don't see it in api/testdata/complete-scenarios

ryanhofdotgov:
@gidjin checkout https://github.com/18F/e-QIP-prototype/issues/1164#issuecomment-439460500

PR #1182, opened 2018-11-19
Update component so it stores capitalized country value

### LINKED ISSUE #1149: Normalize case of countries in Country
**What did you do?**

Enter in "cambodia" in a Country field, like in the dual/multiple citizenships section. Hit Save.

Looking at the payload that gets sent to the database, expecting to see a normalized-case country name, that matches what is in the established country list: "Cambodia".

**What did you see instead?**

"cambodia" – which gets displayed as "Cambodia" in the frontend, but will be passed to XML generation as "cambodia". Could fix it in backend, but seems it should be fixed upstream as soon as data is generated.

**Additional Information**

Want to allow user to input country name in case insensitive manner, just want to persist the normalized version. State component was recently modified to fix similar issue.

Will cause e-QIP to reject application if country name is not cased as expected.

christopherhuii:
Fixes #1149 

`canada` should be `Canada`
`dominican republic` should be `Dominican Republic`
`bassas da india` should be `Bassas da India` (note: "da" is not capitalized)

### DIFF
--- a/src/components/Form/Country/Country.jsx
+++ b/src/components/Form/Country/Country.jsx
@@ -16,7 +16,8 @@ export default class Country extends ValidationElement {
     this.countries = i18n.value('countries')
 
     this.update = this.update.bind(this)
-    this.updateCountry = this.updateCountry.bind(this)
+    this.updateSingleCountry = this.updateSingleCountry.bind(this)
+    this.updateMultipleCountries = this.updateMultipleCountries.bind(this)
     this.handleError = this.handleError.bind(this)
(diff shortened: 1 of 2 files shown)


hursey013:
Looks good, nice work! I noticed a separate, but related issue that I opened a new ticket for: https://github.com/18F/e-QIP-prototype/issues/1209. 

Do you think it's worth adding a unit test for the new `filterValidCountries` function?

christopherhuii:
@hursey013 I added some small basic tests.

Issue #1197, opened 2018-11-19
Last Name field has "Initial Only" option


maryjohnston:
Last Name field has "Initial Only" option

Edit: This change is anywhere last name appears in eApp. You can check on the SF-86 and in the validation matrix/XSD, there isn't a last name initial only attribute like with middle name (e.g, <Middle Type=IO>). So lets just remove the last name initial only option entirely across the entire app.

ryanhofdotgov:
Where the SF-86 says "If you have only initials in your name, provide them and indicate "Initial only"", e-QIP (and the validation matrix page page 4), e-QIP has interpreted that to mean "initials in your [first] name". 

If there's no mention of character length in the validation matrix, then it is generally safe to assume that we shouldn't either (within reasonable limits on maximum). So a one character last name is fine.

PR #1231, opened 2018-11-28
Remove middle name and last name initial checkbox from employment verifier

### LINKED ISSUE #1196: Section 13: Self Employment Verifer has a middle name field in eAPP but not in eQIP.
Section 13: Self Employment Verifer has a middle name field in eAPP but not in eQIP. #1015

christopherhuii:
Fixes #1196 and #1197 

- Removes middle name field from the user interface
- Removes the last name initial checkbox

### DIFF
--- a/src/components/Form/Name/Name.jsx
+++ b/src/components/Form/Name/Name.jsx
@@ -44,7 +44,6 @@ export default class Name extends ValidationElement {
       middle: this.props.middle,
       middleInitialOnly: this.props.middleInitialOnly,
       hideMiddleName: this.props.hideMiddleName,
-      hideLastInitialOnly: this.props.hideLastInitialOnly,
       noMiddleName: this.props.noMiddleName,
       suffix: this.props.suffix,
       suffixOther: this.props.suffixOther,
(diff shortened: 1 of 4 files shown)


hursey013:
Looks good so far! There will also need to be some slight adjustments to the validator for when either middle name is omitted: https://github.com/18F/e-QIP-prototype/blob/develop/src/validators/name.js#L63-L84

christopherhuii:
Not quite sure I understand. I'm not sure how middle name validator is implemented (naming is a little confused), but `this.noMiddleName` is always false so it never enters the switch statement, It always returns `true` with no errors.

ryanhofdotgov:
@christopherhuii #1197 isn't limited to employment verifier, it's anywhere last name appears in eApp. You can check on the SF-86 and in the validation matrix/XSD, there isn't a last name initial only attribute like with middle name (e.g, `<Middle Type=IO>`). So lets just remove the last name initial only option entirely. 

Admittedly, the description on #1197 isn't written well – they were bulk imported by Mary from an NBIB spreadsheet. I have a task this week to go through all these bulk imported NBIB tickets and expand/clarify, so hopefully shouldn't be a problem going forward.

christopherhuii:
@ryanhofdotgov ohhhh, thanks for pointing that out. i only checked the employment verifier and didn't realize it was application wide. i undid those changes and defer the last name initial questions to #1197.

hursey013:
@christopherhuii validator update looks good, thank you.

PR #1232, opened 2018-11-29
Increase timeout for loading initial form data

### LINKED ISSUE #952: Receiving intermittent "Failed to retrieve form saved data" console error
**What did you do?**

Using app pre-loaded with test 5's test data, when computer is under heavy load (Google Video or other high CPU background processes), after logging in the "We are loading your data" screen hangs.

App to load successfully under any load.

**What did you see instead?**

App hangs on loading screen with `Failed to retrieve form saved data` console error.

hursey013:
Fixes #952 

Rather than removing the axios timeout (the default is `0` which removes the timeout), I've increased it from 10 seconds to 30 seconds to allow extra time for the initial application state to load. I also created a basic error page that will load if the timeout threshold is met, rather than leaving the loading page hanging.

To test, you can dial back the timeout to something like `1000` here https://github.com/18F/e-QIP-prototype/blob/develop/src/services/api.js#L42

**Error page:**
screen shot 2018-11-29 at 11 08 35 am

### DIFF
--- a/src/views/Error/Error.jsx
+++ b/src/views/Error/Error.jsx
@@ -1,0 +1,40 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import { i18n } from '../../config'
+
+class Error extends React.Component {
+  render() {
+    return (
+      <div className="login eapp-core" id="login">
+        <div id="seal-header" className="seal-header text-center">
+          <div className="content">
+            <img
+              src="/img/nbis-seal.png"
+              alt="National Background Investigation Services"
+            />
+            <h2>{i18n.t('login.title')}</h2>
+          </div>
+        </div>
+        <div className="content">
+          <div className="table one">
+            <div id="error" className="auth error">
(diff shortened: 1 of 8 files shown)


src/views/Error/Error.jsx
```diff
@@ -0,0 +1,40 @@
...
+          </div>
+        </div>
+      </div>
+    )
+  }
+}
+
+Error.defaultProps = {}
+
+function mapStateToProps(state) {
```
el-mapache: I would probably just remove all of the `connect` code from this file and the test since it doesn't seem to be used

src/views/Error/Error.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import { i18n } from '../../config'
+
+class Error extends React.Component {
```
el-mapache: You can export a function that returns JSX here, since you don't need any of the react lifecycle methods. The `class` is totally fine though

Issue #1010, opened 2018-10-15
ATTACHMENTS_ENABLED=0 disables all attachment functionality 


ryanhofdotgov:
`ATTACHMENTS_ENABLED=0` disables all attachment functionality, not just the functionality affecting  form uploads.

This prevents the end user from downloading the generated PDF release forms after they submit. Alter `http/attachment.go` to only disable save, update, delete. List and get will not get conditionally disabled.

Several related items made this difficult to discover:
* `ATTACHMENTS_ENABLED` defaults to true unlike our other values
* Navigation menu refactor broke auto-hiding of attachments feature
* Historical way eApp has interpreted `VARIABLENAME=`

PR #1233, opened 2018-11-29
Allow release forms to be downloaded with ATTACHMENTS_ENABLED disabled

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/1010

This removes the conditional logic tied to `ATTACHMENTS_ENABLED` for listing and retrieving attachments. This allows the release form PDFs to be downloaded after an application is submitted even if `ATTACHMENTS_ENABLED` is disabled.

I'm having a hard time figuring out how to test this in a unit test, but am open to suggestions.

### DIFF
--- a/api/http/attachment.go
+++ b/api/http/attachment.go
@@ -24,12 +24,6 @@ type AttachmentListHandler struct {
 
 // ServeHTTP serves the HTTP response.
 func (service AttachmentListHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


ryanhofdotgov:
@hursey013 Unit tests for attachment function in general (but not this issue in particular) would look something like `api/http/root_test.go`, where each operation was checked for success/failure depending on the environment variable. Though I think we need to do some mocking of the authn piece. 

We have some related issues in the backlog already, but I'll make sure there is an explicit issue to address it across the board; it is a significant gap in the backend.

Issue #811, opened 2018-09-17
Former spouse birthplace should not collect county


ryanhofdotgov:
`RelationshipStatus/Divorce.jsx` birthplace should be birthplace without county type. Need to propagate change down through XML test cases.

PR #1238, opened 2018-12-03
Remove county from former spouse birthplace

line47:
Fixes https://github.com/18F/e-QIP-prototype/issues/811
Also updated tests 👍 

screen shot 2018-12-03 at 12 06 46 pm

### DIFF
--- a/src/validators/divorce.test.js
+++ b/src/validators/divorce.test.js
@@ -137,7 +137,7 @@ describe('Divorce validation', function() {
             country: { value: 'United States' },
             city: 'Arlington',
             state: 'VA',
-            layout: Location.BIRTHPLACE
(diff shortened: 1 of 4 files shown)


line47:
Thanks, I ran the test and it passed 🤔 I'll take a look 👍

src/validators/divorce.test.js
```diff
@@ -136,7 +136,6 @@ describe('Divorce validation', function() {
           BirthPlace: {
             country: { value: 'United States' },
             city: 'Arlington',
-            county: 'Arlington',
             state: 'VA',
             layout: Location.BIRTHPLACE
```
hursey013: This layout will need to be updated as well to reflect `BIRTHPLACE_WITHOUT_COUNTY`

PR #1245, opened 2018-12-04
Add after hook to output browser errors and fail test

### LINKED ISSUE #671: check for JS errors when running specs
Will be a good additional way to catch errors.

https://stackoverflow.com/questions/29443804/save-console-messages-for-debugging-in-nightwatch-js/29502982

gidjin:
Fixes #671 Added an after hook to the feature tests that will print out the log of browser errors and throw an exception to fail the test if any of the errors are SEVERE level.


gidjin:
I agree probably could be cleaned up but I wanted the full error in the log, the exception ends up showing there and in the summary but with error logs I wanted to err on the side of too much vs too little.

PR #1058, opened 2018-10-23
Set up multi date validation

line47:
This PR fixes #1000 with a few exceptions. 
I created an additional issue for the even more complex validation #1057

PR #1251, opened 2018-12-06
Audit all date instances

### LINKED ISSUE #985: Audit for all uses of `.date` when it hasn't been set
Recently @hursey013 has been fixing issues related to the `date` property not being set in certain areas of the form. It only gets set when data is being changed by the user (or tabs through the field) and not when data is re-loaded from the database.

This problem appears to be pervasive and is likely the source of confusion/odd behavior as @line47 has been working through #773.

While I was testing #981 I had a extremely difficult time getting the data validation to trigger. 

It has code that sets `minDate` like this for the divorce date:
```
 <DateControl
            name="DateDivorced"
            prefix="divorce.date"
            className="date-divorced"
            {...this.props.DateDivorced}
            minDate={(this.props.Recognized || {}).date}
            onUpdate={this.updateDateDivorced}
            onError={this.props.onError}
          />
        </Field>
```

The problem is that `this.props.Recognized.date` is not defined when you reload the data (e.g., sign in, save data, sign out, sign in) unless you manually go change/restore the date that the civil union/ marriage was recognized.

This means that anywhere we are performing this type of validation, using this mechanism, the validation will only happen under limited circumstances.

This task is to search through the code for similar usage patterns and provide a uniform fix, ideally one that operates at a higher level and doesn't require every usage site to execute something extra.

hursey013:
Fixes #985 

**TL;DR**
When passing dates, don't worry about passing the `date` prop specifically, just pass the entire object (for example, `minDate={this.props.Birthdate}` vs `minDate={this.props.Birthdate.date}`).

---

Dates are used in pretty much every part of the app. Currently, those dates are passed around in a variety of ways, but primarily by directly passing a `date` prop like `this.props.Birthdate.date`. The trouble is, that because of the way the app was initially designed, the `date` prop does not exist until the associated `DateControl` component is updated by the end user.

I originally attempted to resolve this issue by using `connect` to add the `date` prop to the `DateControl` component when it was mounted, which worked, however because of the way the accordion is designed, the addition of the `date` prop was never propagated back to the parent component triggering an update to the store. Needless to say, this caused a variety of issues when someone would log out and log back in - no `date` props would exist, making things show up as valid when they were not.

This PR removes the need to rely on the `date` prop at all, and instead uses the day, month, and year values that are stored in the database and available on load for every date item in the app. The bulk of the changes within the PR are updating components not to send `.date`, and removing `date` from all of the tests, since including those has never reflected the true state of the app. I also consolidated a number of different functions that were handling transforming dates into the correct format, and where possible moved logic out of the components themselves and into the validators. This should have the added bonus of allowing the various points within the app that handle validation to use the same code, making the results more consistent.

---
### Date range validation
- Login on the `develop` branch
- Create or update a record in `Where you have lived`, entering 01/01/2000 as the `From` date, and 01/01/1990 as the `To` date.
- Observe a validation error
- Save the form
- Logout and log back in
- Navigate to the same record in `Where you have lived` and observe that even though the dates are wrong the field is valid.
### Not respecting minimum date
- Login on the `develop` branch
- Create or update a record in `Foreign associations > U.S. passport information`, entering 01/01/1980 as the `Issue` date, and 01/01/1970 as the `Expiration` date.
- Observe a validation error
- Save the form
- Logout and log back in
- Navigate to the same record in `U.S. passport information` and observe that even though the dates are wrong the field is valid.

## After
- Switch to the `bh_date_audit_take2` branch
- Restart server
- Log back in
- Navigate back to the input fields mentioned above and observe validation errors are shown on load


ryanhofdotgov:
@hursey013 Would you resolve the conflicts on this PR? I suspect it is a result of the recent merge of https://github.com/18F/e-QIP-prototype/pull/1058

hursey013:
Alright, this should be 👍

ryanhofdotgov:
@christopherhuii @gidjin – this is a significant shift across eApp, so just want to ensure you've had an opportunity to read the TL;DR at the top of this PR.

PR #1252, opened 2018-12-07
Last Name Initial checkbox from Name component

### LINKED ISSUE #1197: Last Name field has "Initial Only" option

christopherhuii:
Fixes #1197 

Removes the last name initial checkbox throughout the application, as well as updating all validators and tests to reflect the removal.


christopherhuii:
@ryanhofdotgov There are two more files I need to change, but I'm not super savvy with XML or how XML generation works, so maybe you could give me a hand.

`api/templates/name-last-first.xml`
`api/xml/templates/name-last-first.xml`
Both those files have a reference of `props.lastInitialOnly` in what looks like an "if" conditional. Which parts of that line do I need to remove?

ryanhofdotgov:
> Both those files have a reference of `props.lastInitialOnly` in what looks like an "if" conditional. Which parts of that line do I need to remove?

This part on line 1 should be deleted in `name-last-first.xml`:
` {{if .props.lastInitialOnly}}Type="IO"{{end}}`
but note that `api/templates` and `api/xml/templates` are the same directory. That's just a directory soft-link in `api/xml`.

(If anyone had checked that box in the eApp UI, this would have caused invalid XML to be generated)

ryanhofdotgov:
> (If anyone had checked that box in the eApp UI, this would have caused invalid XML to be generated)

To elaborate, the Type="IO" XML attribute for last name is not defined by the e-QIP XML and so if XML was generated with that attribute, it would fail XSD validation when run against the incoming application XSD located in https://drive.google.com/open?id=1SDCa2LxlzEmNn6uHhnVYfIQYPt8esOmk (these files are not redistributable and are not in our public repo)

PR #1253, opened 2018-12-07
Update employment items to sort more like a resume

### LINKED ISSUE #1179: Accordion entries do not show in consecutive order by date.
## Example use case
**What did you do?**

Entering employment periods.

Employment entries ordered within the accordion by their start date.

**What did you see instead?**

Employment entries are ordered by timing of entry. 
screen shot 2018-11-17 at 10 13 28 am

christopherhuii:
Fixes #1179 

This issue was reopened to resolve the miscommunication of how employment items were sorted. The employment items should sort like a resume does; with most recent employment ending date as first (or present). Reference the conversation [here](https://github.com/18F/e-QIP-prototype/issues/1179#issuecomment-445102251)

After the fix:
screen shot 2018-12-07 at 9 41 06 am

### DIFF
--- a/src/components/Section/History/Employment/Employment.jsx
+++ b/src/components/Section/History/Employment/Employment.jsx
@@ -62,10 +62,10 @@ export default class Employment extends SubsectionElement {
       if (
         a.Item &&
         a.Item.Dates &&
-        a.Item.Dates.from
+        a.Item.Dates.to
       ) {
-        const aDateObj = a.Item.Dates.from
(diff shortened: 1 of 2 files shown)


michaelccata:
Hey @christopherhuii that looks good to me.

Looks like we are sorting by `most recent end date` to `least recent end date`?

We can test this with users and see what makes most sense ordering wise. Sounds like it's easy to shift if we need to make a slight alteration in the future.

Issue #1250, opened 2018-12-06
Add feature flag for Single Logout


ryanhofdotgov:
In the cases where the identity provider does not support SAML Single Logout, add a feature flag to disable it. Otherwise, the browser will be redirected and likely see a confusing exception.

macrael:
Resolved by https://github.com/18F/e-QIP-prototype/pull/1256

Issue #1206, opened 2018-11-20
Sign Single Logout (SLO) requests


ryanhofdotgov:
Digitally sign the Single Logout (SLO) requests, just like AuthnRequests are signed.

macrael:
Resolved by https://github.com/18F/e-QIP-prototype/pull/1256

PR #1256, opened 2018-12-07
Sign SAML SLO Requests

macrael:
This PR signs all SLO requests for communication with WSO2. WSO2 does not currently use those signatures for anything but one day it will.

I added SAML_SLO_ENABLED as a environment variable for toggling this behavior on and off.


macrael:
This resolves https://github.com/18F/e-QIP-prototype/issues/1250 and https://github.com/18F/e-QIP-prototype/issues/1206

ryanhofdotgov:
@macrael Can you double-check that the comments on the last SLO PR got incorporated in this go-round? PR https://github.com/18F/e-QIP-prototype/pull/1155 Thanks

macrael:
Ahh thanks for the reminder!

macrael:
Original comments addressed, @ryanhofdotgov

gidjin:
This looks good to me, but I'm not familiar with SAML or go so I'll leave this for @ryanhofdotgov to merge.

api/jwt/token.go
```diff
@@ -102,18 +102,21 @@ func (service Service) SessionID(request *http.Request) string {
...
-		return claims.SessionID
+		return claims.SessionIndex
 	}
 
 	return ""
 }
 
 // NewToken generates a new Jwt signed token using a users account information
-func (service Service) NewToken(id int, sessionID string, audience string) (string, time.Time, error) {
+// ID is the user id for the token.
```
ryanhofdotgov:
```suggestion
// id is the user id for the token.
```

docs/CONFIGURATION.md
```diff
@@ -254,6 +255,14 @@ Flag to enable SAML authentication.
 **Default** - False: *empty*<br>
 **Values** - True: `1`, False: *empty*<br>
 
+## `SAML_SLO_ENABLED`
+
+Flag to enable SAML Single Logout authentication.
+
+**Target** - Front-end (web), Back-end (api)<br>
```
ryanhofdotgov: I don't think there's anything on the backend that currently looks at SAML_SLO_ENABLED (though I have a comment elsewhere where I think it should)

api/jwt/token.go
```diff
@@ -92,8 +92,8 @@ func (service Service) CurrentAudience(request *http.Request) string {
 	return ""
 }
 
-// SessionID is the session embedded in the token.
-func (service Service) SessionID(request *http.Request) string {
+// SessionIndex is the session embedded in the token.
```
ryanhofdotgov:
```suggestion
// SessionIndex returns the SAML user session identifier persisted in the token.
```

api/saml/logout_request.go
```diff
@@ -53,28 +67,195 @@ func newLogoutRequest(issuer string, username string, sessionIndex string) logou
 	}
 }
 
-// xml returns the xml for the logout request
-func (req *logoutRequest) xml() ([]byte, error) {
-	output, err := xml.MarshalIndent(req, "", "    ")
+func extractCertBody(certFile string) (string, error) {
```
ryanhofdotgov: I'd replace this with something like this pseudo code:
data, rest := pem.Decode(certFile)
str := base64.StdEncoding.EncodeToString(data)
return str

Reference:
https://golang.org/pkg/encoding/pem/#example_Decode
https://golang.org/pkg/encoding/base64/#Encoding.EncodeToString

docs/CONFIGURATION.md
macrael: yep, it does now.

api/saml/logout_request.go
macrael: Yes, that's so much neater. I tried to use that package but gave it up b/c it's encode method includes the headers. I didn't think to encode the bytes myself.

macrael:
Re the three comments above: 

1. Yes, more defensive code put in

2. I checked at the start of the function instead

3. I'm checking that the signature matches the correct one, so that's a test for the signature. I added a check with xmllint to verify the xsd. It looks like we use xmllint elsewhere so it's not a new dependency.

macrael:
@ryanhofdotgov ready to look over again

api/http/saml.go
```diff
@@ -60,9 +60,9 @@ type SamlSLORequestHandler struct {
 
 // ServeHTTP is the initial entry point for authentication.
 func (service SamlSLORequestHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if !service.Env.True(api.SamlEnabled) {
+	if !service.Env.True(api.SamlSloEnabled) {
```
ryanhofdotgov: In the absence of something at start up making sure SAML_ENABLED and SAML_SLO_ENABLED are set logically consistently, we should add the SLO check to the existing SAML check instead of replacing it.

api/http/saml.go
```diff
@@ -60,9 +60,9 @@ type SamlSLORequestHandler struct {
...
 		service.Log.Warn(api.SamlAttemptDenied, api.LogFields{})
-		http.Error(w, "SAML is not implemented", http.StatusInternalServerError)
+		http.Error(w, "SAML SLO is not implemented", http.StatusInternalServerError)
```
ryanhofdotgov: In this case, it looks like we log a generic error message but send a more detailed reason back to caller. Should be opposite (or the same detailed message). Also should say "enabled" vs "implemented".

ryanhofdotgov:
@macrael The conflicts just need to be addressed (I merged your other backend PR ahead of this one), then feel free to merge 👍

PR #1257, opened 2018-12-08
Fix #1036 time period addl employment overlap

### LINKED ISSUE #1036: Time period of additional employment activities should not overlap main entry
The time periods for additional employment activities are expected to be from a period of time before the current entry. e-QIP will enforce this. Example error:
"To Date [for additional employment activity] must not be later than [this main entry's] Dates of employment From Date, 1/2005"

From the original test case #7:
1) applicant filled out employment entry:

screen shot 2018-10-20 at 9 38 09 am

2) applicant filed out 2 additional (bogus) employment periods underneath this main employment entry.

screen shot 2018-10-20 at 9 38 20 am

screen shot 2018-10-20 at 9 38 25 am

Both additional entries were rejected by e-QIP.

gidjin:
Fix for #1036 Validate that Additional Activity under Employment does not overlap with the main Employment date range.

### DIFF
--- a/src/components/Form/DateRange/DateRange.jsx
+++ b/src/components/Form/DateRange/DateRange.jsx
@@ -151,7 +151,7 @@ export default class DateRange extends ValidationElement {
   handleError(code, value, arr) {
     arr = arr.map(err => {
       return {
-        code: `daterange.${code}.${err.code.replace('date.', '')}`,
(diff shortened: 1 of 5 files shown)


gidjin:
Thanks for looking at this Chris, but I double checked the validation matrix and you cannot enter dates for additional periods that are later than the main entry. **Date must be less than or equal to "Dates of employment" (13A.2.3) From Date .** See page 41 of the matrix. There maybe room for improvement here but I don't think it's part of the scope for this bug.

PR #1260, opened 2018-12-10
Update ISSUE_TEMPLATE.md

michaelccata:
thoughts folks?


ryanhofdotgov:
I'm OK with the changes. There's more text to delete and I'm not sure if the radio selections will make things clearer, but think we should try it out and see.

michaelccata:
Yeah I was hoping that newcomers might be able to parse it better? TBD

Issue #1174, opened 2018-11-17
Suggested auto-populated entry for date gaps is misleading (Your history)


michaelccata:
**Sections affected**

Your history / Where you have lived
Your history / Employment activities
(perhaps other branching sections where gaps appear?)

**What did you see?**

The date range for a "gap" is automatically imported into a new entry which might be misleading. The gap might include several entries but the application assumes it's one by entering the whole range. 

<img width="828" alt="screen shot 2018-11-17 at 9 23 47 am" src="https://user-images.githubusercontent.com/4422873/48662103-da8db680-ea4a-11e8-8249-af2e6dfe2d39.png">
<img width="787" alt="screen shot 2018-11-17 at 9 23 59 am" src="https://user-images.githubusercontent.com/4422873/48662104-deb9d400-ea4a-11e8-9963-6ecd824f09b3.png">

**What did you expect to see?**

Perhaps only adding the first date would be useful? Or just get rid of the auto-populate in general?

**Additional Information**

gidjin:
Has there been discussion with Mary about what she prefers as the solution to this issue?

joshmfranklin:
I would suggest leaving both dates empty and let the user decide which dates to input. Giving them more freedom to start, and if we have to circle back and give them smarter defaults (i.e. only provide `from date`) we can do that at a later time.

michaelccata:
That sounds good to me!

michaelccata:
Let's clear out the autofill and move forward.

PR #1263, opened 2018-12-12
Do not auto fill in gap dates as they may be confusing

gidjin:
Github Issue #1174

### DIFF
--- a/src/components/Section/History/Employment/Employment.jsx
+++ b/src/components/Section/History/Employment/Employment.jsx
@@ -6,11 +6,10 @@ import SubsectionElement from '../../SubsectionElement'
 import { Accordion, Branch } from '../../../Form'
 import { openState } from '../../../Form/Accordion/Accordion'
 import { newGuid } from '../../../Form/ValidationElement'
-import { today, daysAgo } from '../dateranges'
+import { today, daysAgo, extractDate, validDate } from '../dateranges'
 import { InjectGaps, EmploymentCustomSummary } from '../summaries'
 import EmploymentItem from './EmploymentItem'
 import { Gap } from '../Gap'
-import { extractDate } from '../dateranges'
 
 const byline = (item, index, initial, translation, required, validator) => {
   // If item is required and not currently opened and is not valid, show message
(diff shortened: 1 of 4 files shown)


gidjin:
Yeah I saw that. The section is automatically sorted by date. The error doesn't disappear also because the new section doesn't have a date, since the user technically still has a gap. I wonder if scrolling to the new section is enough to resolve that. Though the validation isn't rerun when you edit the new section so it doesn't go away until you go to the next section. Another option is to fill in one of the dates, like the from. I haven't tried that though to see if it resolves the error, but I image it would allow us to sort it.

@joshmfranklin any thoughts on what would be better from a user perspective?

joshmfranklin:
@gidjin here are the two things we discussed during our chat: 

Option 1: See if adding the `from date` helps with placing the new accordion in the right spot when clicking on the "Add new employer" button in a gap validation error

Option 2: Always place the gap validation error at the end of the accordion list and replace it with a new accordion when clicking on the "Add new employer" button.

gidjin:
I updated the pr to support having blank dates, but now the gap error is removed when you click to add a new range, and the new range & the gap will appear at the bottom of the list. Hopefully this helps with the usability.

christopherhuii:
I'm not sure if this is intentional or not, but I found some behavior I think is strange. This might just result in a separate GitHub issue.

screen shot 2018-12-18 at 7 56 24 am
In the image above, I have a gap in my employment, which is expected.

If I save and revisit the section, eApp will save that gap employment entry.
screen shot 2018-12-18 at 7 59 49 am

As a user, I'd expect to still see the gap entry instead of eApp automatically creating an employment item with the gap dates.

hursey013:
I'm seeing something similar to @christopherhuii, and agree that it might be a separate issue. If I'm working on the employment section for example, enter in several entires, then navigate away, and then return I see the gap warning at the bottom of the accordion (with blank dates) as expected. However if I log out and back in, return to employment, I see an entry with exact dates autopopulated in the the gap, as well as the blank entry (without dates) at the bottom of the list.

gidjin:
I was discussing this with Chris and we both see the issue described on develop as well as on my branch so it seems not related to my changes since it's a pre-existing issue. He was going to file a bug issue for it since it's not caused by these changes.

hursey013:
New issue relating to sorting residences chronologically: https://github.com/18F/e-QIP-prototype/issues/1281 - which is relevant now that the gap warning message will be displayed at the bottom of the list.

Issue #1274, opened 2018-12-14
Add basic failure path test for identification section


gidjin:
Currently the feature specs provide minimal test coverage. For this task please add a basic failure path test for the identification section. Something like going straight to the review and seeing that all the errors show up and then filling in the fields or something.

Reference: https://docs.google.com/document/d/1HoXI8i4wBcjWq_elD39BuNYdesi8cjHnnERLk4JcfT4/edit#bookmark=id.yw3fljo9pvsh

See #1212 as it implemented the minimal happy path check

PR #1275, opened 2018-12-14
Make identification e2e tests pass

### LINKED ISSUE #1212: Get Identification section spec tests passing
Reference: https://docs.google.com/document/d/1HoXI8i4wBcjWq_elD39BuNYdesi8cjHnnERLk4JcfT4/edit#bookmark=id.yw3fljo9pvsh

For this task, it is not just sufficient to get the existing Identification section tests passing, but to validate that those test cover the section, and augment them as necessary.

gidjin:
Fixes #1212 by adding a passing spec. Also this generated #1274 to cover failure example at least minimally

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ Setup steps:
 1. Disable *SAML\_ENABLED* auth and enable *BASIC\_ENABLED*
 1. Edit /etc/hosts file __OPTIONAL__
   * Add _web_ and _api_ to the end of the localhost line in /etc/hosts so you don't have to keep changing the values of API\_REDIRECT or API\_BASE\_URL in .env file back to localhost
+1. Run __make run__ to ensure services are restarted
+1. Run the purge directions from docs/test-scenarios.md to clear out old data
 
 To run the feature specs locally:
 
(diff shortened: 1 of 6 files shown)


gidjin:
I've been running the test on it's own and it passes as expected. It failed when run with the entire suite I didn't have time this round to look at why, but once all the tests are passing individually we should circle back to ensure the whole suite passes

gidjin:
Yes I've been thinking of pulling the code out into page objects so that any test can reuse the code to fill in the form. I wasn't sure on the best approach for that with Nightwatch, but there is support for page objects. I'm open to other suggestions on making them more reusable, I would like to see a test that went through one of the NBIB test scenarios too, but probably we want some basic page objects and fixtures before doing that.

README.md
```diff
@@ -68,6 +68,8 @@ Setup steps:
 1. Disable *SAML\_ENABLED* auth and enable *BASIC\_ENABLED*
 1. Edit /etc/hosts file __OPTIONAL__
   * Add _web_ and _api_ to the end of the localhost line in /etc/hosts so you don't have to keep changing the values of API\_REDIRECT or API\_BASE\_URL in .env file back to localhost
+1. Run __make run__ to ensure services are restarted
+1. Run the purge directions from docs/test-scenarios.md to clear out old data
```
ryanhofdotgov: @gidjin Markdown doesn't seem to be numbering the steps as you would expect:

screen shot 2018-12-17 at 1 56 44 pm

gidjin: Apparently not. I've updated it, I couldn't get sublists to number properly so I took them out since it always started over.

Issue #1101, opened 2018-11-05
Can't download attachments


ryanhofdotgov:
**What did you do?**

Load up test1.json, submit (web service disabled, so a no-op), click on certification PDF on final page.

**What did you expect to see?**

Certification PDF downloads.

**What did you see instead?**

No action take, 400 error in browser console.  This error in server log:
```
time="2018-11-05T10:52:54Z" level=warning msg="No identifier for attachment provided" account=2 error="strconv.Atoi: parsing \"\": invalid syntax" ip="192.168.2.156:42130"
```

**Additional Information**

In attachment GET handler, `mux.Vars()` returns an empty map.  I think this is related to https://github.com/gorilla/mux/issues/191 and the addition of the JWT middleware, which manipulated the context. 

I think solution is to upgrade gorilla mux. However, last time gorilla mux upgrade was attempted shortly before last vendor went off contract it broke the backend (OPTIONS calls were failing) and they reverted. It does not look like they opened up an issue. Here is a comment I had from that team:

"...not sure if something has changed with our HTTP mux (we have been using gorilla/mux) but if i set all the allowed HTTP methods we expect on the root router
```r.Methods("GET", "POST", "PUT", "DELETE", "OPTIONS")```
it starts to reach our handler methods."

For expediency sake, I might consider reverting the JWT handler related code, duplicating the code back in each method, then see performing library upgrade under a different issue.

ryanhofdotgov:
@macrael Just to clarify, this issue isn't test case-specific, just suggested loading up test1.json because it's the fastest way to get a complete data set that allows you to "naturally" pass validation and submit and get to the download links affected.

If you are testing against a debug build, you can also put in some bare minimum info (name, birthdate), then click a "Force submit" link in the lower left corner of the submit page.

macrael:
👍 

We can use this issue to track a better fix, probably upgrading mux.

PR #1277, opened 2018-12-17
Update Mux to restore status code fix

macrael:
The old version of Mux we were using had a bug when interacting with the standard context package, the fix to implement status codes used that package and so was reverted. This un-reverts those changes and updates mux so there is not a conflict anymore. 

I've tested this locally with CORS enabled. There was a problem before where updating Mux broke CORS OPTION requests, but that does not appear to be a problem anymore. It's probably worth testing this in a staging environment incase something is different from the local environment.


ryanhofdotgov:
@macrael In absence of something automated, have you done a (manual) full end-to-end process test with these changes? For example, some thing like:
* purge database
* load up test scenario
* change some data, save
* log out, log in
* click through section reviews
* go to final review, sign all the waivers
* submit application
* download all the digitally signed PDF waivers/releases

macrael:
I did the above locally and everything seemed to work fine.

macrael:
@ryanhofdotgov unless you think we should try and push this to the staging area for a final test I think that this looks like its working as is.

ryanhofdotgov:
Fixes #1101

PR #1279, opened 2018-12-18
Add logic in Accordion component to ensure address books are updated

### LINKED ISSUE #1178: "Previous address autofill" for Supervisor Address
**What did you do?**

Entering data for a private sector employment (non-government employment) and had to fill out multiple addresses. The functionality exists for military job postings, but not for private sector or state/local government. See screenshots below.

screen shot 2018-11-17 at 10 06 29 am

The functionality exists for military job postings, but not for private sector or state/local government. See screenshots below.

screen shot 2018-11-17 at 10 12 05 am
screen shot 2018-11-17 at 10 11 57 am

christopherhuii:
Fixes #1178 

This PR is a hack to get address books working across the application. Currently, each child of the accordion relies on the accordion to attach props to the child, which complicates things because the old address book from each employment/residence/education item overwrites the prop that was passed directly to the child component.

Each employment, residence, and education item doesn't not have the updated address books when the addresses are validated. Instead of finding a way to retrofit each of the items to include the updated address books, I opted to update the accordion component and have it manage the address book.

I'm not particularly proud of this solution, but it gets the job done.


src/components/Form/Location/Address.jsx
```diff
@@ -292,7 +292,7 @@ export default class Address extends ValidationElement {
           </RadioGroup>
         </Show>
 
-        <Show when={this.props.addressBook && book.length}>
+        <Show when={this.props.addressBook && book.length && !this.props.validated}>
```
hursey013: What's the benefit of hiding it once the address is validated? I could see the scenario where someone mistakenly selected the wrong address from the address book and needs to reselect another address.

PR #1298, opened 2018-12-24
Date range validating too early

### LINKED ISSUE #1286: Date ranges are validating too early
## Where are you in the form?

**Section:**

Any date range in the application

**Question:**

I've mostly been testing in the `other names used` section, `dates used` question

**Field:**

Date range component

## What did you see?

When I input a date into the `from date` field the entire component starts validating when I enter the year. This happens before I enter anything into the `to date`

- [x] Validation error
- [ ] Data not saving
- [ ] Other (explain below)

I expected to see that the date range doesn't get validated until both the `from date` and `to date` are entered.

This seems to be happening in any date range.

- [x] Cross-form issue
- [ ] I don't know

**Operating system**
- [ ] Windows
- [x] MacOS

**Browser**
- [x] Chrome
- [ ] Firefox
- [ ] IE
- [ ] Edge
- [ ] Other

**Version:**

screen shot 2018-12-19 at 1 18 16 pm

gidjin:
Fixes #1286 by making date range validations only validate if both fields have been touched by the user to avoid showing the error too early.

### DIFF
--- a/src/components/Form/DateControl/DateControl.jsx
+++ b/src/components/Form/DateControl/DateControl.jsx
@@ -28,6 +28,7 @@ class DateControl extends ValidationElement {
       month: props.hideMonth ? '1' : props.month,
       day: props.hideDay ? '1' : props.day,
       year: props.year,
+      touched: props.month >= 1 && props.day >= 1 && props.year >= 1,
       errors: []
     }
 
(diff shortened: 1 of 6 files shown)


hursey013:
Testing this a bit more, I'm not sure this fully addresses the issue since `touched` will get set immediately after any input in the `To` date, so while it is not shown as invalid while entering the `From` date, after the first digit is input into `From` we still run into the original issue:

screen shot 2019-01-03 at 1 41 35 pm

gidjin:
Brian, good catch, I've updated the branch so that it is not marked as touched until all 3 parts are entered. Please take another look.

hursey013:
The code looks like it's on the right track, but something must be slightly off because when running the app I'm seeing the original issue again:

screen shot 2019-01-04 at 9 45 19 am

This is the `Your history > Where you have lived` section

hursey013:
Yeah it looks like if Redux contains empty strings for the dates (rather than undefined) you get the scenario above. So basically if you were to start from a fresh account, go to `Your history > Where you have lived` hit save, log out, log back in, return to `Your history > Where you have lived` you'll see the date range validating too early.

PR #1335, opened 2019-01-10
Fix date control to stop validating the date too early

### LINKED ISSUE #1307: Single date fields validating too early
## Where are you in the form?

**Section:**
Multiple

**Question:**
Multiple

**Field:**
Multiple

Single date component (month/day/year). This doesn't seem to be happening with month/year components.

## What did you see?

Error validations are triggering as soon as a suer starts to type anything into the year field.

- [x] Validation error
- [ ] Data not saving
- [ ] Other (explain below)

Steps to reproduce:
- Find any month/day/year date control and start typing in a date
- See that as soon as you type any digit into the year field it starts to validate the entire date.

I expect each part of the date to validate on blur (i.e. as soon as the user leaves the field it will validate against its necessary rules).

- [x] Cross-form issue
- [ ] I don't know

**Operating system**
- [ ] Windows
- [x] MacOS

**Browser**
- [x] Chrome
- [ ] Firefox
- [ ] IE
- [ ] Edge
- [ ] Other

**Version:**
kapture 2018-12-28 at 11 21 45

gidjin:
Fixes #1307 

Turns out the date control toggled focus of the day field of a date control when some change was made forcing the validation to trigger. The purpose of this seems to have been made obsolete as removing the code does not change the interaction but does fix the bug. One of the tests implies that the day field should have been automatically filled in, but that no longer happens so I am removing the offending code and test, added new test to check for the bug and have verified that this works for single date field and date ranges.

### DIFF
--- a/src/components/Form/DateControl/DateControl.jsx
+++ b/src/components/Form/DateControl/DateControl.jsx
@@ -91,27 +91,6 @@ class DateControl extends ValidationElement {
     this.setState(
       { month: month, day: day, year: year, estimated: estimated, touched: touched },
       () => {
-        // Estimate touches the day so we need to toggle focus
-        const toggleForEstimation = changed.estimated
-
-        // Potential for typical day out-of-bounds (including leap year)
-        const toggleForDay = changed.year || changed.month
-
-        // Any external influence (i.e. clicking `Present` in a date range)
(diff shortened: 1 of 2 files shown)


gidjin:
Removed the unused variable and pushed an update

PR #1342, opened 2019-01-14
Update eApp to be configurable for other forms (SF85)

christopherhuii:
Please add the URL parameter `?formType=85` to view the SF-85 version of the form. No URL parameter is needed for the SF-86; it is SF-86 by default.

**ONLY** the `Your history` section has been updated to match the SF-85.

A lot is going on in this PR and I will be the first to admit that this is far from being finished. This is the first iteration of how I viewed this refactor to accommodate for the SF-85.

Start in `src/components/Main/App.jsx` in `componentDidMount` and trace the logic through there.

We will need to do much much regression to make sure all functionality in eApp continues to work and not too many new bugs have been introduced.


src/reducers/application.js
```diff
@@ -45,12 +45,3 @@ export default combineReducers({
   AddressBooks: reducer('AddressBooks'),
   Navigation: navigationReducer
 })
-
-// Or alternative...
```
gidjin: Yay for deleting commented out code! 👍

src/components/Section/History/History.test.jsx
```diff
@@ -98,7 +98,7 @@ describe('The History section', () => {
       </Provider>
     )
 
-    expect(component.find(Employment).props().totalYears).toEqual(2)
+    expect(component.find(Employment).props().totalYears).toEqual(10)
```
gidjin: This must be the test for the totalYears method mentioned earlier.

gidjin:
Also found that the next button under _where you went to school_ still says **Former federal service** and goes to that section even though formType was set to 85.

 screen shot 2019-01-14 at 11 53 31 am

The Review section under _Investigative and criminal history_ also points to **Psychological and emotional health intro** and seems to go to the that section though nothing is displayed. I'm not sure if these were part of your work or not but probably should be addressed to call those sections done.

joshmfranklin:
A couple of questions need to be changed from *7* years to *5* years, see below: 
screen shot 2019-01-15 at 3 19 08 pm

I think these questions may only be triggered if you enter a employment activity within the last 5 years. SF-86 uses 7 years to this show/hide logic may need to change from 5 to 7.

PR #1351, opened 2019-01-15
Substance use date field validation bug

### LINKED ISSUE #1339: Substance abuse counseling date present not functioning
## Where are you in the form?

**Section:**
Substance abuse > Use of alcohol > additional instances

**Question:**
Provide the date counseling or treatment **ended**

**Field:**

Date Control block

## What did you see?

- [X] Validation error
- [X] Data not saving
- [ ] Other (explain below)

Select present when filling out the section and observe the accordion header
screen shot 2019-01-09 at 5 00 39 pm

If you reload the page login again and find the field you will see that present is no longer selected and month is off by 1
screen shot 2019-01-09 at 5 00 21 pm

I expected the current month to be shown in the header (in my case January) and I expected present to still be selected when returning to the form.

I haven't tested the entire form but I believe it is localized too

- [X] Only relevant to just this question
- [ ] Cross-form issue
- [ ] I don't know

**Operating system**
- [ ] Windows
- [X] MacOS

**Browser**
- [X] Chrome
- [ ] Firefox
- [ ] IE
- [ ] Edge
- [ ] Other

**Version:**
Latest Development git hash 532416ad969fa0ae4ab9fd873a4ef17c588da6d4

gidjin:
Fixes 1338. There are two issues, first when checking present on treatment end date the month was filled in but minus one. Second when end date has yet to be filled in undefined was used for the maxDate value for treatment start date which caused strange behavior. The code fixes this by ensuring maxDate is the treatment end date or today. This doesn't fix 1339 which will need to be handled separately.

### DIFF
--- a/src/components/Section/SubstanceUse/Alcohol/ReceivedCounseling.jsx
+++ b/src/components/Section/SubstanceUse/Alcohol/ReceivedCounseling.jsx
@@ -111,7 +111,7 @@ export default class ReceivedCounseling extends ValidationElement {
       endDate = {
         date: date,
         estimated: false,
-        month: String(date.getMonth()),
+        month: String(date.getMonth()+1),
         year: String(date.getFullYear()),
         day: String(date.getDate())
       }
--- a/src/components/Section/SubstanceUse/Alcohol/ReceivedCounseling.jsx
+++ b/src/components/Section/SubstanceUse/Alcohol/ReceivedCounseling.jsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { i18n } from '../../../../config'
+import { validDate, today } from '../../History/dateranges'
 import {
   Location,
   DateControl,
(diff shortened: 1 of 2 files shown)


hursey013:
This appears to do the trick, but I'm curious if swapping the date components out for the date range component might make sense? That way we don't have to replicate the min/max logic that is built into `DateRange`. Do you know if there is something unique about this instance where is needs to remain two independent `DateControl`'s?

joshmfranklin:
@hursey013 I had originally thought that this was a specific instance in the SF86 where the `beginning` and `end` dates were asked separately. But upon further inspection of the PDF form it does look like it is a `DateRange`. If the existing components work then maybe we just write up a separate story to change this into a `DateRange` component?

gidjin:
There is a second bug with this field #1339 perhaps as part of that fix we can swap it for a date range and fix both issues at once?

PR #1352, opened 2019-01-16
Reorganizes and adds visual scroll cues to introduction/instructions modal

### LINKED ISSUE #1346: Improvements to introduction/instructions modal
- Order of copy in the introduction/instructions modal needs to be rearranged. `Persons completing this form should begin after carefully reading the preceding instructions` should precede the scrollable block of text and change `preceding` to `following`.

- In user testing, users were not aware that large blocks of scrollable text were in fact scrollable (e.g. introduction modal). Explore adding a slight inset shadow to scroll block in order to hint at ability to scroll text. See codepen for example: https://codepen.io/gillytech/pen/dlbsx

joshmfranklin:
Fixes #1346

### DIFF
--- a/src/components/Form/Introduction/Introduction.jsx
+++ b/src/components/Form/Introduction/Introduction.jsx
@@ -53,6 +53,7 @@ export class Introduction extends React.Component {
           closeable={this.props.forceOpen}
           onDismiss={this.props.onDismiss}>
           <div>
+            <h2>{i18n.t('introduction.acceptance.title')}</h2>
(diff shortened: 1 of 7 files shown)


hursey013:
Looking good. Would it be possible to place the shadow within the desktop breakpoint so it doesn't appear in the mobile view?

screen shot 2019-01-16 at 3 33 10 pm

joshmfranklin:
@hursey013 yup makes sense, should be fixed up now!

PR #1353, opened 2019-01-17
Add "I don't know" option to Selective Service

### LINKED ISSUE #1320: Applicants should be able to select "I don't know" for Selective Service Record
As an applicant
I want to be able to select "I don't know" when asked about my Selective Service Record
So that I can give the most accurate information for my SSS status

Given I am on the Selective Service Record subsection
And I have answered "Yes" to being a male born after December 31st, 1959
Then I should see a question asking me "Have you registered with the Selective Service System (SSS)?"
And I should be able to choose "I don't know"

Given I have chosen "I don't know"
Then I should see a multiline text field asking me "Provide explanation"

[Link to SF-85 content guide for this section](https://docs.google.com/presentation/d/1LCahQUxtp2lcIjjVguzivokDe-cXEcZhVWzk-6S9pc0/edit#slide=id.g44d6d9eb88_2_82)

hursey013:
Fixes #1320 

Uses the `NotApplicable` component to add an "I don't  know" option to Selective Service record.

Includes a new database migration, so make sure to restart the server when testing.

screen shot 2019-01-17 at 10 47 34 am

### DIFF
--- a/api/templates/military-selective.xml
+++ b/api/templates/military-selective.xml
@@ -9,6 +9,12 @@
     <LegalExemptionExplanation>{{textarea .props.Explanation}}</LegalExemptionExplanation>
     <LegalExemptionExplanationComment></LegalExemptionExplanationComment>
     <NumberComment></NumberComment>
+    <DNKExplanation></DNKExplanation>
+    <DNKExplanationComment></DNKExplanationComment>
+    {{- else if notApplicable .props.HasRegisteredNotApplicable | eq "True"}}
+    <LegalExemptionExplanation></LegalExemptionExplanation>
+    <LegalExemptionExplanationComment></LegalExemptionExplanationComment>
+    <NumberComment></NumberComment>
     <DNKExplanation>{{textarea .props.Explanation}}</DNKExplanation>
     <DNKExplanationComment></DNKExplanationComment>
     {{end}}
--- a/api/migrations/20190116164719_add_idk_selective_service.sql
+++ b/api/migrations/20190116164719_add_idk_selective_service.sql
@@ -1,0 +1,12 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+-- +goose StatementBegin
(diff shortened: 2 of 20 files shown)


api/templates/military-selective.xml
```diff
@@ -9,6 +9,12 @@
     <LegalExemptionExplanation>{{textarea .props.Explanation}}</LegalExemptionExplanation>
     <LegalExemptionExplanationComment></LegalExemptionExplanationComment>
     <NumberComment></NumberComment>
+    <DNKExplanation></DNKExplanation>
```
gidjin: Did you mean to add lines 12 and 13? Looks like duplicates of 18 and 19. If so do we need to fill in the explanation?

gidjin:
Looks like this works. Should we show the _Provide an explanation_ box if the applicant selects _I don't know_? Also please see the commend about the api template xml, looks like there maybe some inadvertent duplication.

api/templates/military-selective.xml
hursey013: Admittedly, this is where my lack of knowledge on how the XML is expected to look downstream comes into play. It looks like in a lot of areas we have blank XML tags in the template, like `<LegalExemptionExplanationComment></LegalExemptionExplanationComment>`, so for consistency's sake I return the empty tags in both conditions, but I'm not totally sure that is necessary. The way I have it now returns the XML exactly as it was before, the only difference is in one case the explanation is in the `LegalExemptionExplanation` tag, and in the other the `DNKExplanation`. Open to change this around if you think we can omit the empty tags.

hursey013:
@gidjin yes the explanation box should appear when you select `I don't know` - was it not appearing during your testing?

gidjin:
It appeared as expected then.

api/templates/military-selective.xml
gidjin: Ah I think I follow the code now. Sorry for the confusion it was on my end.

PR #1367, opened 2019-01-18
[WIP] Update 'Citizenship' section to support SF85

### LINKED ISSUE #1356: [SF-85] Update Citizenship section components to match SF-85
- [ ] Update Locales
- [ ] Confirm/Update form fields within each section (see [tracking document](https://docs.google.com/spreadsheets/d/1nHJhOiQFH2k0QzQQ0HkrsllDWh6ED5yi4ytbioiLjNs/edit?usp=sharing))
- [ ] Update left navigation to match the correct subsections (Subsections **NOT** part of SF-85: `Foreign passports`)
- [ ] Ensure the "back" and "next" buttons navigate to the correct sections
- [ ] `Introduction` title will need to change from `Section 4` to `Section 3`

**Note:**
We should also consider bringing the `U.S. Passport information` subsection from the `Foreign associations` section to the `Citizenship` section for all forms. Reason is that for the SF-85 it would be the only subsection within `Foreign associations` and it seems to make sense to include it in the `Citizenship` section.

christopherhuii:
Fixes #1356 

Remember, in order to test the SF85, you need to add the URL `parameter ?formType=85`.

Please remember to check for:

- locale accuracy
- left sidebar navigation accuracy
- 'back' next' button label and navigation accuracy
- form field accuracy
- both 85 and 86 are functioning as expected.

In the GitHub issue, Josh suggested:

```
We should also consider bringing the U.S. Passport information subsection from the Foreign associations section to the Citizenship section for all forms. Reason is that for the SF-85 it would be the only subsection within Foreign associations and it seems to make sense to include it in the Citizenship section.
```

I decided not to implement that at this time because it's not a showstopper. Moving the U.S. Passport subsection from Foreign associations to Citizenship will probably require a bit of discussion and estimation of effort of that change. On the front end, the lift isn't that heavy, but I am unsure of the backend effort that it might require so that things submit to eAgency correctly.


src/components/Section/Citizenship/Status/Status.jsx
```diff
@@ -982,47 +984,50 @@ export default class Status extends SubsectionElement {
               adjustFor="buttons"
               scrollIntoView={this.props.scrollIntoView}>
               <RadioGroup
-                className="citizenship-document-type"
+                className="citizenship-document-type option-list-vertical"
```
gidjin: Why were these classes added?

src/components/Section/Citizenship/Status/Status.scss
```diff
@@ -1,13 +0,0 @@
-.status {
```
gidjin: Were these unused?

christopherhuii:
deferring fora different strategy

PR #1368, opened 2019-01-18
Enable formType URL parameter on non-dev environments

christopherhuii:
Previously, the `?formType=85` URL parameter was restricted to development environments. I have removed that conditional so folks can test the SF85 out on the develop branch deploy.


christopherhuii:
Outdated PR

PR #1369, opened 2019-01-18
Add basic i18n testing

### LINKED ISSUE #698: throw an error if a translation is missing
...or at least in test mode, or at least a warning. Too easy for something to be missed right now.

```javascript
i18n.t('identification.intro.title') // "Section 1: Information about you"
i18n.t('en.foo') // "en.foo"
```

hursey013:
Fixes #698

This collects all the i18n strings used within the components and makes sure that there is a corresponding key in the locale files. Right now it's somewhat dumb in that it doesn't handle dynamic strings very well (which are unfortunately most of the error/validation messages). 

I'm also skipping the duplicate and unused string tests for the time being until I can remove the number of false positives.

This also includes fixing the initial batch of missing strings the test found - the majority of them were labels being used with the `DateControl` component which doesn't actually accept a label as a prop.


src/components/Section/Foreign/Activities/IndirectActivity/IndirectInterest.jsx
```diff
@@ -284,9 +284,6 @@ export default class IndirectInterest extends ValidationElement {
             name="Acquired"
             className="acquired"
             {...this.props.Acquired}
-            label={i18n.t(
-              `foreign.activities.indirect.interest.label.acquired`
```
gidjin: Are these still in use, if not could they be removed from the locale definition file as well?

src/config/locales/i18n.test.js
```diff
@@ -0,0 +1,65 @@
+import {
```
gidjin: I like having a static test for this, I was wondering if we should also modify the i18n library to throw a warning in the logs at least as the original story suggested since that maybe easier to handle the dynamic keys. I'm fine if that is tabled as solution for another time though.

gidjin: Are there additional steps to run this test, locally when I run `make` or `make test-react` it fails 

  FAIL  src/config/locales/i18n.test.js (21.975s)
  ● i18n › does not have missing strings

    expect(received).toBe(expected) // Object.is equality

    Expected: 0
    Received: 15

      45 |       console.log(report)
      46 |     }
    > 47 |     expect(report.length).toBe(0)
         |                           ^
      48 |   })
      49 |
      50 |   it('does not have unused strings', () => {

      at Object.<anonymous> (src/config/locales/i18n.test.js:47:27)

  ● i18n › does not have unused strings

    expect(received).toBe(expected) // Object.is equality

    Expected: 0
    Received: 2154

      55 |       console.log(report)
      56 |     }
    > 57 |     expect(report.length).toBe(0)
         |                           ^
      58 |   })
      59 |
      60 |   it.skip('does not have duplicate strings', () => {

      at Object.<anonymous> (src/config/locales/i18n.test.js:57:27)

src/components/Section/Foreign/Activities/IndirectActivity/IndirectInterest.jsx
hursey013: These are flagged by the test as being referenced in the code, but missing from the locale file. So in these cases, these strings are unneeded since DateControl doesn't accept a label, so I removed it from the jsx files, and since it's already missing from the locale file, there's no need to remove anything there.

src/config/locales/i18n.test.js
hursey013: I did a couple force pushes after I messed up trying to squash some commits, could you try pulling down the latest again and seeing if you are still getting these failures? They should have been resolved already.

hursey013: I did have the missing strings being logged to the console, but I wasn't sure if that would jump out enough to make it useful, but I think it can't hurt. I'll add it back in, and maybe only log them on dev, does that sound good?

gidjin: I think only logging them in dev would be great, if it's not too much trouble. I'll pull the latest and confirm.

gidjin: Works on latest commits thanks!

PR #1379, opened 2019-01-25
Setup javascript cache busting

### LINKED ISSUE #588: New releases don't invalidate cache?
**What did you do?**

Browse to https://eqip-prototype.fr.cloud.gov after a release.

eApp navigation menu rendered correctly.

**What did you see instead?**

In Chrome, navigation menu missing UI elements (border, menu divisions, etc.)

screen shot 2018-07-17 at 5 12 46 pm

Looked fine in Safari.

**Additional Information**

Reported by @bpdesigns. Believe this is caused by JavaScript or CSS being cached by Chrome from a previous release. Hypothesis is that we need to make sure etags is working or add some release-specific URL to the static resources so a new release won't use cached browser assets from a prior release.

Work-around is to clear browser cache, use incognito browser window, etc.

hursey013:
Fixes #588 

This adds a unique hash to the javascript filenames to avoid unintentional caching. Webpack will update the hash when it detects a change in the file's content, otherwise it remains the same. This change also splits out the vendor code into it's own js file, so that if the dependencies stay the same, and it's just a change to the app code, the vendor js file can remain untouched (thus cached).


hursey013:
Follow on changes could address cache-busting css, however that would likely be a more involved changed due to the way the css is currently being built. Ideally we would move the css into the webpack build process, eliminating the need for Gulp.

Issue #1293, opened 2018-12-20
Phantom error on education items (test case 7)


hursey013:
Steps to reproduce:
- Load test case 7
- Review education section, including final review of History section and see that there are no validation errors
- Run the final review and observe an error flagged in the in `Your history > Where you went to school`

I did some preliminary research on this. It's tied to the ReferenceName component and the `hideMiddleName` prop. `hideMiddleName` successfully hides the middle name field, and works as expected, however because of the way the validator works, the values for `ReferenceName` are pulled in from the Redux store where `hideMiddleName` is not  present. This causes the validator to flag `ReferenceName` as invalid because it appears it is missing a middle name.

joshmfranklin:
This also happens on the Employment activities subsection.

PR #1383, opened 2019-01-29
Don't include current spouse if not married

### LINKED ISSUE #1377: Sometimes eApp lets applications submit without requiring spouse citizenship
Kickback issue seems to be related to eApp not always requiring spouse citizenship. i.e., eApp will permit an application be submitted without the user specifying current spouse citizenship. Trying to get additional details.

Investigate, see how to reproduce, and resolve.

ryanhofdotgov:
Fixes #1377. Also: encode HaveFormerSpouse in alignment with validation matrix. Side effect: marital status is moved to top of Spouse section in all test cases.

Extended test6 to include triggering scenario (divorced but not currently married). All test scenarios re-submitted to e-QIP successfully (modulo test4, which has 100+ year old applicant now).

### DIFF
--- a/api/testdata/complete-scenarios/test6.json
+++ b/api/testdata/complete-scenarios/test6.json
@@ -1549,7 +1549,6 @@
                       "first": "",
                       "firstInitialOnly": false,
                       "middle": "",
-                      "hideMiddleName": true,
                       "middleInitialOnly": false,
                       "noMiddleName": false,
                       "last": "",
@@ -1731,7 +1730,6 @@
                       "first": "",
                       "firstInitialOnly": false,
                       "middle": "",
-                      "hideMiddleName": true,
                       "middleInitialOnly": false,
                       "noMiddleName": false,
                       "last": "",
@@ -3014,12 +3012,6 @@
             "value": "Yes"
           }
         },
-        "HasRegisteredNotApplicable": {
-          "type": "notapplicable",
-          "props": {
-            "applicable": true
-          }
-        },
         "RegistrationNumber": {
           "type": "text",
           "props": {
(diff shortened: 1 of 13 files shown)


api/testdata/complete-scenarios/test6.json
```diff
@@ -1549,7 +1549,6 @@
                       "first": "",
                       "firstInitialOnly": false,
                       "middle": "",
-                      "hideMiddleName": true,
```
hursey013: Was this intentional?

api/testdata/complete-scenarios/test6.json
```diff
@@ -1731,7 +1730,6 @@
...
```
hursey013: Same here, not sure if these values are no longer needed, or were just accidentally removed.

ryanhofdotgov: Those changes weren't manual – I loaded up test case 6, made my changes to spouse, and used `form` to export it back out. I had assumed that those were things that had changed in the code base. Should they still be there? It is what eApp saves right now.

hursey013: It looks like what is happening is the prop `hideMiddleName` is being set on certain `<Name>` components, which visually accomplishes the goal, but it does not appear the value is being stored in the database, which it should, since it is used in validation. But that's all unrelated to this change, so we could just restore these values and I can open up another issue, or we could remove them, but it may cause a "phantom" error on the frontend.

hursey013: Actually, looks like I already wrote this up: https://github.com/18F/e-QIP-prototype/issues/1293. Rest of the changes look good though, so I'll give it a 👍

Issue #1392, opened 2019-01-29
Implement cache-busting for CSS


ryanhofdotgov:
In issue #588, we describe some problems when loading the app with a browser that has cached versions of JavaScript and CSS.

In #1379, JavaScript caching was addressed, but CSS is still outstanding.

PR #1396, opened 2019-02-01
Add CSS cache busting

hursey013:
Fixes https://github.com/18F/e-QIP-prototype/issues/1392

Adds unique hash based on content of css/sass files to avoid stylesheets being cached longer than desired.


ryanhofdotgov:
I managed to figure it out – thanks, looks good!

Issue #1365, opened 2019-01-18
Accordion component causes eApp to hang


christopherhuii:
Steps to Reproduce

1. Start a new and clean form
2. Navigate to `4. Citizenship`
3. Click `Dual/multiple citizenship`
4. Answer `Yes` to `Do you now or have you EVER held dual/multiple citizenships?`
5. Save immediately
6. Refresh the page and log back in
7. Navigate to `4. Citizenship`
8. Click `Dual/multiple citizenship`

Bug: The page will hang and be non-responsive. When the page finished hanging, you can open the developer tools to view the console error.

```
Uncaught Error: Objects are not valid as a React child (found: object with keys {value}). If you meant to render a collection of children, use an array instead or wrap the object using createFragment(object) from the React add-ons. Check the render method of `Accordion`.
```

PR #1398, opened 2019-02-01
Fix addresses bug to default to United States

### LINKED ISSUE #1266: Address field set defaults to "Outside of US"
## Where are you in the form?

**Section:**
Relationships --> Relatives
**Question:**
Provide your relative's current address,  Provide your relative's place of birth
**Field:**
Was this person born in the United States of America?, This address is
Address block

## What did you see?
- [ ] Validation error
- [ ] Data not saving
- [X] Other (explain below)

Sign in to new/fresh form. Go to relatives section. Click Mother, Click No to "deceased?". Click Save. Logout, log back in. Go back to Mother entry. 

Address blocks for birth and current address default to "Outside US".

screen shot 2018-12-14 at 9 40 00 am

screen shot 2018-12-14 at 9 40 05 am

Default to "In US", which was displayed on the original form creation/sign-in.

- [x] Cross-form issue
- [ ] I don't know

Probably cross-form, but I don't know.

**Operating system**
- [ ] Windows
- [X] MacOS

**Browser**
- [X] Chrome
- [ ] Firefox
- [ ] IE
- [ ] Edge
- [ ] Other

**Version:**

christopherhuii:
Fixes #1266 and #1365 

This PR fixes two issues because they are somewhat related.

When a country field is fresh and untouched, it is initialized as:
```
{
  value: null
}
```

When a country field is loaded from the database, it is filled out as a string: "Canada"

When a country field is updated in the input field, it fills out as an object:
```
{
  value: "Cana"
}
```

Because of each of these cases, this is why we created the `countryString` function. I needed to tweak it a little bit so it knows how to handle `null` values.

### DIFF
--- a/src/components/Form/Location/Address.jsx
+++ b/src/components/Form/Location/Address.jsx
@@ -12,10 +12,7 @@ import ApoFpo from '../ApoFpo'
 import Show from '../Show'
 import Suggestions from '../Suggestions'
 import { AddressSuggestion } from './AddressSuggestion'
-import LocationValidator, {
-  countryString,
-  isInternational
-} from '../../../validators/location'
+import LocationValidator, { countryString } from '../../../validators/location'
 import { countryValueResolver } from './Location'
 
 export default class Address extends ValidationElement {
--- a/src/components/Form/Location/Address.jsx
+++ b/src/components/Form/Location/Address.jsx
@@ -163,13 +163,23 @@ export default class Address extends ValidationElement {
   }
 
   addressType() {
-    const country = countryString(this.props.country)
+    let country = this.props.country
+    if (typeof country === 'object') {
+      country = countryString(this.props.country)
+      if (country === '') {
+        return 'International'
+      } else if (country === null) {
+        return 'United States'
+      }
+    }
 
     if (['United States', 'POSTOFFICE'].includes(country)) {
       return country
+    } else if (country === '') {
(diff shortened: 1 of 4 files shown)


src/components/Form/Location/Address.jsx
```diff
@@ -394,7 +404,7 @@ export default class Address extends ValidationElement {
               </div>
             </div>
           </Show>
-          <Show when={isInternational(this.props)}>
```
hursey013: Looks like this is the only instance of `isInternational` so it can be removed from the imports.

hursey013:
Also out of curiosity, which aspect of this resolved #1365?

src/components/Form/Location/Address.jsx
christopherhuii: good catch, removed

christopherhuii:
> Also out of curiosity, which aspect of this resolved #1365?

Based on the console error, I was able to determine that something in the Accordion was rendering an object, which isn't allowed. After a lot of debugging, I was able to find out that the error was coming from within `src/components/Summary/Summary.jsx` because it was trying to render an object.

`this.props.left` is a country with a `null` value:
```
{
  value: null
}
```

The `countryString` function was returning the `null` object, which was causing problems in the rendering. So updating the `countryString` to account for `null` fixed it.

Issue #1034, opened 2018-10-19
Clicking "I don't know" for verifier email should clear email


ryanhofdotgov:
**What did you do?**

For education verifier for school in past 3 years, specify email (e.g. foo@example.com) Then click "I don't know". Then click on "I don't know" again to clear it.

**What did you expect to see?**

No previously stored email address

**What did you see instead?**

foo@example.com – the field was not cleared.

**Additional Information**

Likely affects other places where email is collected but optionally can be an "I don't know".

suzubara:
Taking this on, can't assign or push yet though

PR #1401, opened 2019-02-06
Don't duplicate value state in Email form component

suzubara:
Fixes https://github.com/18F/e-QIP-prototype/issues/1034

Root cause was the Email input `value` persisting in component state. It was not being used in any way that differed from `props.value` so I removed the state altogether.

### DIFF
--- a/src/components/Form/Email/Email.jsx
+++ b/src/components/Form/Email/Email.jsx
@@ -6,11 +6,6 @@ export default class Email extends ValidationElement {
   constructor(props) {
     super(props)
 
-    this.state = {
-      pattern: props.pattern,
-      value: props.value
-    }
-
     this.handleError = this.handleError.bind(this)
   }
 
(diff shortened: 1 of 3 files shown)


christopherhuii:
When I:
1. Enter an email
2. Click `I don't know`
3. Deselect `I don't know`

...the email still persists. You might need to investigate `src/components/Form/NotApplicable/NotApplicable.jsx`

@gidjin can you confirm whether this works or not? thanks

suzubara:
@christopherhuii @gidjin I'm not able to reproduce, or maybe am mis-understanding what you're experiencing. If I follow @christopherhuii 's steps, the email does not show up in the input, and I also no longer see it in the Redux state. I am specifically testing in the `History > Residence > ReferenceEmail` section, so if you are experiencing this issue elsewhere then I wonder if there is some inconsistency with how the NotApplicable components are handled in different sections..

christopherhuii:
Maybe we're on different pages so I'll provide more detail.

1. Open eApp
2. Navigate to `2. Your History` -> `Where you went to school`
3. Select `Yes` for `Have you attended any schools in the last 10 years?`
4. Provide dates of attendance: (from date: `12/12/2017`; to date: `Present`)
5. Scroll down to the heading `List a person who can verify that you attened this school`. (NOTE: SPELLING ERROR)
6. Enter a first name
7. Enter an email
8. Click `I don't know`
9. Unselect `I don't know`

Bug: The email is still filled in.

PR #1409, opened 2019-02-07
Update .babelrc to understand class arrow functions

christopherhuii:
`this` is a very important scope issue in Javascript.
This plugin will transform class properties so that we can define class properties using property initializer syntax.

Currently
```
class Component extends React.Component {
  constructor(props) {
    super(props);

    this.someFunction = this.someFunction.bind(this)
  }

  someFunction() {
    {...}
  }

  render() {
    {...}
  }
}
```

After this PR
```
class Component extends React.Component {
  constructor(props) {
    super(props);
  }

  someFunction = () => {
    {...}
  }

  render() {
    {...}
  }
}
```

The arrow functions automatically binds `this`.

Things we need to verify:
1. The code runs (duh) and functionality is not changed.
2. ESLint doesn't complain about the syntax.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-    "presets": ["es2015", "react", "stage-2"]
+    "presets": ["es2015", "react", "stage-2"],
+    "plugins": ["transform-class-properties"]
 }
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -8,6 +8,7 @@ env:
   es6: true
   node: true
   jest: true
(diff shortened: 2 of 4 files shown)


gidjin:
FYI we are holding off on merging this until after this sprint is complete to give us more time for it to shake out before making it to pre-prod.

gidjin:
I converted one class to the new syntax and tests passed, lint passed, and form still appeared to function when I went filling it out. I'm holding off approving this until the sprint changes over.

PR #1413, opened 2019-02-08
Use Babel to resolve project root & aliases

### LINKED ISSUE #1424: Use Babel to resolve project root & aliases
Configure Babel so that modules can be imported using absolute paths and/or aliases.

https://github.com/18F/e-QIP-prototype/pull/1413

suzubara:
- Adds `babel-plugin-module-resolver` and configures the root path & sets up aliases in `.babelrc`
- Adds `eslint-import-resolver-babel-module` so eslint can resolve aliases
- Use alias imports in a few existing files to make sure it works
- Fixes #1424


.eslintrc.yml
```diff
@@ -27,4 +27,17 @@ rules:
     - error
     - skipUndeclared: true
 settings:
-  import/resolver: webpack
+  import/resolver:
```
suzubara: Duplicating the config here seems to be a necessary evil due to https://github.com/tleunen/eslint-import-resolver-babel-module/issues/95 (in order for eslint to properly resolve the modules). This is fixed on `5.0.0` of `eslint-import-resolver-babel-module`, but we're locked to `^4.0.0` because 5 requires Babel 7, and upgrading Babel feels outside the scope of this work.

gidjin:
Changes look good, We are going to hold off merging this until after the sprint ends to give us more time with it locally before it goes out.

PR #1419, opened 2019-02-11
Add documentation for changes to database & API

### LINKED ISSUE #1399: Document an end-to-end change
Choosing one of the past PRs that includes a DB migration and update to the API and documenting that process.

hursey013:
Fixes #1399 

This outlines the steps necessary to persist data on the frontend, database, and API.


hursey013:
@ryanhofdotgov in regards to modifying the test scenarios (when newly added data needs to be represented) - are there any restrictions when changing those files that I should note in the documentation?

docs/updating-api-db.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Update XML template and test data
+
+After finding the corresponding XML file in `/api/templates`, make sure that the data is properly represented within the XML. It is also important to make sure that the new input is added to the corresponding `.json` file found within `/api/testdata` which is used to test the generated XML. Lastly, it's also ideal to add the new input to at least one of the complete scenarios in `/api/testdata/complete-scenarios`. This will allow the data to be pre-loaded and tested using the [load-scenario](https://github.com/18F/e-QIP-prototype/blob/develop/docs/test-scenarios.md#loading-existing-test-json-files) script.
```
ryanhofdotgov: @hursey013 One small recommended change to the guidance. 

Each one of the existing test cases 1-9 was created for OPM for a specific purpose – those are documented in `api/xml/xml_test.go`. Test case 6 was created for rejection (i.e, they weren't looking for specific areas of content) so it is a good test case to extend. I would leave the others as is. Alternatively, you can create a brand new test case.

gidjin:
Thanks the documentation looks good to me, I'm glad you included example links. I'm not marking approval as Ryan had some good additional feedback.

docs/updating-api-db.md
hursey013: That makes sense, just updated to add some clarification.

PR #1431, opened 2019-02-12
Nest routes, section/subsection refactor - Identification

### LINKED ISSUE #1422: Make use of react nested routes instead of SectionViews component
SectionViews component handles common rendering and routing features, the scope of this story is to replace the routing portion only with react nested routes. Basically SectionViews component loops through the section components and then returns the specific component, but this can easily be handled by react nested routes.

This task should be time boxed to 2 days for exploration, and if possible we can continue the refactor, but if not we can close the time box and regroup.

Note this work will befit the SF-85 work as it will be easier to make our navigation more declarative after this refactor.

suzubara:
## Description
- Dependent on https://github.com/18F/e-QIP-prototype/pull/1439
- Fixes #1422 
- Converts entire `Identification` section to use `<Route>` to render the current subsection, and starts to clean up section rendering
- For each subsection component:
  - Remove redundant & un-used props/state (replacing with config attributes where possible)
  - Update imports to use aliases
  - Use new `Subsection` component, which is almost identical to `SubsectionElement` but with minor implementation changes.
    - Went this route since it seems risky to try and convert all form sections in one go, and this will let us move forward without touching the `SubsectionElement` which is still in use by legacy code
    - Additionally, right now each subsection component depends on extending the `SubsectionElement` component *and* receiving `props` from its section parent. This definitely could use some cleanup, but these dependencies are more related to store updates, validation, and errors, and so I think should be done under a different story.
  - Use `connectIdentificationSection` to pass in methods that came from props before, and to connect the component to its appropriate store.
      - _Note:_ I don't love having to set up the `sectionConfig` object for each subsection, but it seems necessary for now in order to fit seamlessly into the rest of the architecture. I think a cleaner way to structure the subsection components may become clearer as we move further along with refactoring.
- Create explicit components for Intro and Review "subsections", use `<Review />` for print rendering since they are identical
- Created `SectionNavigation, SectionNavButton` for more modular previous/next buttons
- Set up scaffold for migrating one form section at a time into new rendering architecture

## Checklist for Reviewer
- [ ] Review code changes
- [ ] QA: No user-facing changes expected, however filling out the Identification section should be regression-tested.

### DIFF
--- a/src/components/Section/Identification/ApplicantName/ApplicantName.jsx
+++ b/src/components/Section/Identification/ApplicantName/ApplicantName.jsx
@@ -2,18 +2,66 @@ import React from 'react'
 import { i18n } from '../../../../config'
 import schema from '../../../../schema'
 import validate from '../../../../validators'
-import SubsectionElement from '../../SubsectionElement'
 import { Name, Field } from '../../../Form'
 
-export default class ApplicantName extends SubsectionElement {
+import { reportCompletion } from '../../../../actions/ApplicationActions'
+
+import connectIdentificationSection from '../IdentificationConnector'
+
+const sectionConfig = {
+  section: 'identification',
+  subsection: 'name',
+  store: 'Identification',
+  storeKey: 'ApplicantName',
+}
+
+export class ApplicantName extends React.Component {
   constructor(props) {
     super(props)
+
+    const { section, subsection, store, storeKey } = sectionConfig
+
+    this.section = section
+    this.subsection = subsection
+    this.store = store
+    this.storeKey = storeKey
+
+    this.handleError = this.handleError.bind(this)
+    this.handleCompletion = this.handleCompletion.bind(this)
     this.update = this.update.bind(this)
     this.updateName = this.updateName.bind(this)
   }
 
+  handleCompletion() {
+    const data = {
+      ...this.props,
+      ...this.state
+    }
+
+    this.props.dispatch(
+      reportCompletion(
+        this.section,
+        this.subsection,
+        this.props.validator(data)
+      )
+    )
+  }
+
+  handleError(value, arr) {
+    this.handleCompletion()
+    arr = arr.map(err => {
+      return {
+        ...err,
+        section: this.section,
+        subsection: this.subsection
+      }
+    })
+    
+    return this.props.onError(value, arr)
+  }
+
   update(queue) {
-    this.props.onUpdate({
+    this.props.onUpdate(this.storeKey, {
       Name: this.props.Name,
       ...queue
     })
@@ -25,12 +73,19 @@ export default class ApplicantName extends SubsectionElement {
     })
   }
 
+  dataAttributes() {
+    return {
+      'data-section': this.section,
+      'data-subsection': this.subsection
+    }
+  }
+
   render() {
     const klass = `section-content applicant-name ${this.props.className ||
       ''}`.trim()
 
     return (
-      <div className={klass} {...super.dataAttributes(this.props)}>
+      <div className={klass} {...this.dataAttributes(this.props)}>
         <h1 className="section-header">{i18n.t('identification.destination.name')}</h1>
         <Field
           title={i18n.t('identification.name.title')}
@@ -41,7 +96,6 @@ export default class ApplicantName extends SubsectionElement {
           <Name
             name="name"
             {...this.props.Name}
-            dispatch={this.props.dispatch}
             onUpdate={this.updateName}
             onError={this.handleError}
             required={this.props.required}
@@ -59,8 +113,6 @@ ApplicantName.defaultProps = {
   onError: (value, arr) => {
     return arr
   },
-  section: 'identification',
-  subsection: 'name',
   dispatch: () => {},
   required: false,
   validator: data => {
@@ -69,3 +121,5 @@ ApplicantName.defaultProps = {
 }
 
 ApplicantName.errors = []
+
+export default connectIdentificationSection(ApplicantName, sectionConfig)
--- a/src/components/Section/Identification/ApplicantName/ApplicantName.jsx
+++ b/src/components/Section/Identification/ApplicantName/ApplicantName.jsx
@@ -1,21 +1,26 @@
 import React from 'react'
-import { i18n } from '../../../../config'
+import { i18n } from '@config'
+
 import schema from '../../../../schema'
 import validate from '../../../../validators'
 import { Name, Field } from '../../../Form'
 
-import { reportCompletion } from '../../../../actions/ApplicationActions'
-
 import connectIdentificationSection from '../IdentificationConnector'
+import SubsectionElement from '../../SubsectionElement'
+
+import {
+  IDENTIFICATION,
+  IDENTIFICATION_NAME,
+} from '@config/formSections/identification'
 
 const sectionConfig = {
-  section: 'identification',
-  subsection: 'name',
-  store: 'Identification',
-  storeKey: 'ApplicantName',
+  section: IDENTIFICATION.name,
+  subsection: IDENTIFICATION_NAME.name,
+  store: IDENTIFICATION.store,
+  storeKey: IDENTIFICATION_NAME.storeKey,
 }
 
-export class ApplicantName extends React.Component {
+export class ApplicantName extends SubsectionElement {
   constructor(props) {
     super(props)
 
@@ -23,41 +28,10 @@ export class ApplicantName extends React.Component {
 
     this.section = section
     this.subsection = subsection
-    this.store = store
     this.storeKey = storeKey
 
-    this.handleError = this.handleError.bind(this)
-    this.handleCompletion = this.handleCompletion.bind(this)
     this.update = this.update.bind(this)
     this.updateName = this.updateName.bind(this)
-  }
-
-  handleCompletion() {
-    const data = {
-      ...this.props,
-      ...this.state
-    }
-
-    this.props.dispatch(
-      reportCompletion(
-        this.section,
-        this.subsection,
-        this.props.validator(data)
-      )
-    )
-  }
-
-  handleError(value, arr) {
-    this.handleCompletion()
-    arr = arr.map(err => {
-      return {
-        ...err,
-        section: this.section,
-        subsection: this.subsection
-      }
-    })
-    
-    return this.props.onError(value, arr)
   }
 
   update(queue) {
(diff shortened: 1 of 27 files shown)


src/components/Section/Identification/ApplicantName/ApplicantName.jsx
```diff
@@ -2,18 +2,66 @@ import React from 'react'
...
 import validate from '../../../../validators'
-import SubsectionElement from '../../SubsectionElement'
 import { Name, Field } from '../../../Form'
 
-export default class ApplicantName extends SubsectionElement {
+import { reportCompletion } from '../../../../actions/ApplicationActions'
+
+import connectIdentificationSection from '../IdentificationConnector'
+
+const sectionConfig = {
```
suzubara: This could be hoisted to a top level config

src/components/Section/Identification/ApplicantName/ApplicantName.jsx
```diff
@@ -2,18 +2,66 @@ import React from 'react'
...
+      return {
+        ...err,
+        section: this.section,
+        subsection: this.subsection
+      }
+    })
+    
+    return this.props.onError(value, arr)
+  }
+
```
suzubara: Moved these functions in here in order to decouple from `SubsectionElement`, but ultimately could be extracted and passed into each section component as part of an HOC

src/components/Section/Identification/Identification.jsx
```diff
@@ -1,20 +1,45 @@
...
+        <Route path="/form/identification/name" component={ApplicantName} />
+        <Route path="/form/identification/birthdate" component={ApplicantBirthDate} />
+        <Route path="/form/identification/birthplace" component={ApplicantBirthPlace} />
+        <Route path="/form/identification/ssn" component={ApplicantSSN} />
+        <Route path="/form/identification/othernames" component={OtherNames} />
+        <Route path="/form/identification/contacts" component={ContactInformation} />
+        <Route path="/form/identification/physical" component={Physical} />
+        <Route path="/form/identification/review" component={Review} />
+      </div>
+    )
```
suzubara: This is what we're trying to achieve instead of `SectionViews`

src/components/Section/Identification/IdentificationConnector.jsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
+
+import {
+  updateApplication,
+  reportErrors,
+} from '../../../actions/ApplicationActions'
+
+const connectIdentificationSection = (Component, { section, subsection, store, storeKey }) => {
```
suzubara: I wrote this HOC just for the Identification section, however it could definitely be abstracted and applied to any section.

src/components/Section/Identification/IdentificationConnector.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+  const mapStateToProps = (state) => {
+    const app = state.application || {}
+    const identification = app.Identification || {}
+    const errors = app.Errors || {}
+    const completed = app.Completed || {}
+
+    switch (storeKey) {
+      case 'ApplicantName':
+        return { ...identification.ApplicantName } || {}
```
suzubara: Temporary verbosity in order to get parity with how these props are currently passed into each section, can be cleaned up later.

src/components/Section/Identification/Review.jsx
```diff
@@ -0,0 +1,109 @@
...
+    return (
+      <div>
+        <ApplicantName
+          required={true}
+          scrollIntoView={false}
+        />
+  
+        <hr className="section-divider" />
+  
+        {/*
```
suzubara: Remaining sections TBD after preliminary review from the team

src/components/Section/SectionView.jsx
```diff
@@ -13,96 +13,92 @@ export class SectionViews extends React.Component {
...
-              onClick={this.handleTransition.bind(this, child.props.back)}>
-              <div className="actions back">
-                <div className="icon">
-                  <i className="fa fa-arrow-circle-left" aria-hidden="true" />
-                </div>
-                <div className="text">
-                  <div className="direction">Back</div>
-                  <div className="label">{child.props.backLabel}</div>
-                </div>
+      let previousButton = <div className="btn-cell" />
```
suzubara: TBD - need to add this back into each section

src/components/Section/Identification/ApplicantName/ApplicantName.jsx
suzubara: i.e. https://github.com/18F/e-QIP-prototype/pull/1439/files#diff-8ce44bb418dc407a198758d1fe38422e

src/components/Section/Identification/Identification.jsx
```diff
@@ -1,20 +1,108 @@
...
 import ApplicantName from './ApplicantName'
 import ApplicantSSN from './ApplicantSSN'
 import ApplicantBirthPlace from './ApplicantBirthPlace'
 import ApplicantBirthDate from './ApplicantBirthDate'
 import OtherNames from './OtherNames'
 import Physical from './Physical'
 import ContactInformation from './ContactInformation'
+import Review from './Review'
+
+const navigationConfig = [
```
suzubara: TODO move the below to high level config

src/components/Section/SectionView.jsx
suzubara: This is now completed at https://github.com/18F/e-QIP-prototype/pull/1431/files#diff-e0df189d16e0e8e1b2360e055bb7bd57R54

gidjin:
These changes look good to me. I think where you broke off tasks for future is good too.

src/components/Section/Section.jsx
```diff
@@ -70,8 +70,31 @@ class Section extends React.Component {
...
-    return <SectionViews current={this.props.section}>{sections}</SectionViews>
+    return (
+      <Switch>
+        {/* REFACTORED - These sections are rendered via <Route>s */}
+        <Route path="/form/identification" component={Identification} />
+
+        {/* TBD */}
+        <Route path="/form/:section/:subsection" render={() => (
+          <SectionViews current={this.props.section}>{this.createSections()}</SectionViews>
+        )} />
```
suzubara: ^ this should let us make these changes one section at a time

src/components/Section/Identification/Review.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+import ApplicantName from './ApplicantName'
+import ApplicantBirthDate from './ApplicantBirthDate'
+import ApplicantBirthPlace from './ApplicantBirthPlace'
+import ApplicantSSN from './ApplicantSSN'
+import OtherNames from './OtherNames'
+import ContactInformation from './ContactInformation'
+import Physical from './Physical'
+
+class Review extends React.Component {
```
christopherhuii: Can we convert this to a stateless functional component?

src/components/Section/Section.jsx
```diff
@@ -70,8 +70,31 @@ class Section extends React.Component {
   }
 
   render() {
-    const sections = this.createSections()
-    return <SectionViews current={this.props.section}>{sections}</SectionViews>
+    return (
+      <Switch>
+        {/* REFACTORED - These sections are rendered via <Route>s */}
+        <Route path="/form/identification" component={Identification} />
```
christopherhuii: I'm struggling to wrap my head around how these routes are going to work.
The SF86 will have:

```
<Route path="/form/relationships" component={Relationships} />
```
...but the SF85 will not have that section.

How can we make sure that route is not exposed? Is there a way we can use the form type config to generate this?

src/components/Section/Identification/Identification.jsx
```diff
@@ -1,349 +1,90 @@
...
-            minimumEmails={1}
-            shouldFilterEmptyItems={true}
-            defaultState={false}
-            dispatch={this.props.dispatch}
-            onUpdate={this.handleUpdate.bind(this, 'Contacts')}
-            onError={this.handleError}
-            required={true}
-            scrollIntoView={false}
-          />
+    const navButtons = (
```
christopherhuii: Is there a reason why we need to abstract this component into its own variable?

src/components/Section/shared/SectionNavigation.jsx
```diff
@@ -0,0 +1,77 @@
...
+  if (!back && !next) {
+    return null
+  }
+
+  const backButton = back ? (
+    <SectionNavButton
+      direction="back"
+      label={backLabel}
+      onClick={() => { goToSection(`/form${back.path}`)}} />
+  ) : <SectionNavButton direction="back" isEmpty={true} />
```
christopherhuii: For both the back and next button, can we remove this ternary from the component level and put it in the prop level?

```
isEmpty={back ? true : false}

// or simply

isEmpty={!!back}
```
...then put the component directly in the render instead of storing it in a variable

src/components/Section/Section.jsx
suzubara: Yes, the routes can be generated from a config (or omitted from the JSX) based on the `formType` value.

src/components/Section/Identification/Identification.jsx
suzubara: The only reason I did this was because there's a check for whether these buttons exist when rendering the `top-btns` div (see https://github.com/18F/e-QIP-prototype/pull/1431/files/128717208eec1861d840c27c4340026bbd6db792#diff-e0df189d16e0e8e1b2360e055bb7bd57R56). I'm not sure why this check exists because I don't see how the two are related, and it also seems that `navButtons` will *always* exist, but I left it in there for parity 🤷‍♀️ happy to remove it if we determine it's unnecessary.

christopherhuii: Yeah, it looks like `navButtons` will _always_ evaluate as truthy because it's a React component. Removing `navButtons` from the conditional is probably safe.
```
 {showTop && (
   <div className="top-btns"><ErrorList /></div>
  )}
```

Then there's no need for the `navButtons` variable.

suzubara: well if both `back` and `next` are false-y then the component does return `null`, but I don't see why that should make any difference to the `top-btns` div. I'll remove it.

christopherhuii: If I remember correctly, even if the component returns `null`, `navButtons` is still a React component.  The `render` shouldn't affect whether it's truthy or falsey.

Issue #1429, opened 2019-02-11
Retrieve formType from storage and include in payload to front end


gidjin:
As part of the push to support the SF-85 as well as multiple other future forms we need to implement a way for the backend to know the form type. This information is already stored in the version deployed with secure storage; however the information is not sent to the frontend in the user information and it is unclear how to support this in the version that does not use secure storage. This will include some investigation into how this information is stored and how to retrieve it from secure storage.

NOTE This story is _**NOT**_ to figure out how to use the data in the front end, but just how to retrieve it from secure storage, send it to the frontend in the payload, and how to replicate this in the version of the app that does not use secure storage.

See #1428

macrael:
1. The form type will be encoded in the GET /me/form response so that the UI can switch based on what form it pulls down from that api. 

2. eAgency sends a parameter called "level" when it provisions a user, that is saved in the accounts table on a "level" column. We will replicate that column and behavior here. 

3. eventually, but not for this PR, the form types that a user has been assigned (b/c there are edge cases where there can be multiple) should be set on the accounts table so we know which forms are available for a user to fill out. That will require the ability to specify multiple levels on the accounts table as well as an API for requesting different levels from the /me/form interface.

macrael:
We'll need something to make it easy to set the level for a user in development, I think that a binary like the `unlock` tool to toggle the form would suffice. It will come with warnings that switching a user that has already partially filled out a form has data integrity issues.

ryanhofdotgov:
What are the domain of values for `level`? To set the stage for future iterations, what ever we send to the front end should be the form type (e.g., SF-86) and a version string (e.g., 2016-11), augmenting `level` if required.

macrael:
Is that version sent by eAgency or IM? In the eAgency API it is specified as an enum of “SF85”, “SF86”, “SF85P” but I would love if level just had the version built in instead of having two separate columns for it.

ryanhofdotgov:
https://gsa-tts.slack.com/archives/CD1HQH1EY/p1551746167081300

macrael:
Code is in https://github.com/18F/e-QIP-prototype/pull/1541

macrael:
Merged

PR #1378, opened 2019-01-24
Add formType URL param to configure form type

christopherhuii:
This PR is the first step for developing new forms. In order to view different forms, we need a way to identify them.

This is a basic implementation of using URL parameters to define the form type.
By appending `?formType=` to the base URL you can define a custom form type.

For example, after you login using the URL: 

`http://localhost:3000/login` will store the following in Redux
```
authentication: {
  authenticated: true,
  token: "NpYy1zZXNzaW9uIiwiYXVkIjoiQmFzaWFfADinOWFEOWEIfjiwFOIEfcmwoei",
  formType: "SF86",
  error: ""
}
```

`http://localhost:3000/login?formType=SF85` will store the following in Redux
```
authentication: {
  authenticated: true,
  token: "NpYy1zZXNzaW9uIiwiYXVkIjoiQmFzaWFfADinOWFEOWEIfjiwFOIEfcmwoei",
  formType: "SF85",
  error: ""
}
```

`http://localhost:3000/login?formType=lolwut` will store the following in Redux
```
authentication: {
  authenticated: true,
  token: "NpYy1zZXNzaW9uIiwiYXVkIjoiQmFzaWFfADinOWFEOWEIfjiwFOIEfcmwoei",
  formType: "lolwut",
  error: ""
}
```

**Note**: This URL parameter is only available on:
  1. local dev environments
  2. develop deploy branch environment.

PR #1435, opened 2019-02-12
Implement initial ground work for form management

### LINKED ISSUE #1428: Implement top level form type branching
As part of the push to support the SF-85 as well as multiple other future forms we need to implement a way for the front end to know the form type. This information should be stored in redux and passed as a prop to the appropriate components. It may need to be in Redux Context if it has to be passed around to more than top level components. 

This work is reflected in the branch https://github.com/18F/e-QIP-prototype/compare/develop...ch-1371-identification-sf85 and in the old PR https://github.com/18F/e-QIP-prototype/pull/1378

NOTE This story is _**NOT**_ to figure out how to get the formType from the backend, but simply how to handle storing and passing it around in the front end.

See #1429

christopherhuii:
## Description

Fixes #1428 
This is an initial attempt at implementing form management. This is not a perfect solution, but it handles the basic requirements needed to handle multiple forms.

As a proof of concept, please view:
`http://API_BASE_URL/login`
`http://API_BASE_URL/login?formType=85`

You'll notice the navigation sidebar on the left has the appropriate sections (Note: subsections and questions have not been addressed yet)

There's a decent amount of work that's going on here, so I'll try to describe the body of work from a high level.
1. In development, it uses a URL parameter `?formType=` to differentiate the form type. We have an outstanding issue #1429 that will grab the form type from user data.
2. Once the applicant is logged in and the form type is stored in Redux, we need to load the correct form. A lot of the work stems from `src/components/Main/App.jsx`, specifically `componentDidMount`. This is a good place to trace through all the changes.

_Note: In order to narrow down the scope of work, I cut and pasted certain functions rather than attempt to refactor all of them._

3. Each form is described in a "schema" located in `src/config/form`. This form schema is stored in Redux.

## Checklist for Reviewer

- [ ] Review code changes
- [ ] Load up test cases and confirm data is correct
- [ ]  Fill out the form from top to bottom


src/actions/AuthActions.js
```diff
@@ -56,16 +66,17 @@ export function tokenError() {
   }
 }
 
-export function handleLoginSuccess(token) {
+export function handleLoginSuccess(token, formType = '86') {
```
suzubara: I'd recommend creating constant definitions for each of the form types, so that all of the values are centrally defined.

christopherhuii:
@suzubara What do you suggest we keep?

I've broken out the form type stuff into a separate PR (#1378)

I'm completely okay with declining this PR because it is kinda outdated work. A big reason why things are the way they are was because I was focusing on form management rather than refactoring the application.

The static constant `navigation` in `src/config/navigation.js` drives the application. It's referenced all over the application, especially in helper functions around the application. I put the form in Redux to make it easier to reference rather than refactoring.

suzubara:
@christopherhuii I think the new PR for the form type work is a good example!

When I say break things out I mean isolating changes to the code that are independent from each other. For example, updating the UI to display the correct form name, and updating the progress/score card with the correct number of sections. Additionally I noticed AuthenticatedView is removed from a few components here, which is another thing we've talked about doing (and I definitely support doing) but doesn't need to be coupled w everything else.

It also explicitly *removes* the `navigation` constant (exported from `config/navigation.js`) reference. I would recommend instead, adding whatever code is needed to make that redundant, and then subsequently removing it (this may happen over the course of several PRs, as I know that constant is widely used).

If you want to drop this PR since as you mentioned it's outdated, I would say that's up to you! It might be easier going forward than to try and reconcile it with the other work going on -- however it can still be used as a reference if that's helpful.

Also, I understand that the `props.form, props.form.sections` in Redux is meant to replace the `navigation` constant. However the configuration of each form type (what you've defined in `config/form/sf85.js, sf86.js`) is constant, and can be referenced directly by each component that needs it -- I'd reserve use of `props` for parameters that are subject to change (such as which `formType` the user is accessing).

christopherhuii:
I recommend we drop this and pick up pieces along  the way, if needed.

PR #1437, opened 2019-02-13
Styling fixes

### LINKED ISSUE #1420: Styling fixes
## Employment activities
- [x] Employment "Other" explanation textarea has large gap
- [x] Reason for leaving textare and date needs to be properly spaced

## Police record
- [x] Introduction page needs to use same page title styling as every other section

## Substance use
- [x] Review title screen says "lets make sure everything looks right" while all other review screens are titled "review your answers"

## Psychological and emotional health > Diagnosis
- [x] Health care professional name is hard to click on 
- [x] Current treatment care professional does not have title like the previous one does

joshmfranklin:
fixes #1420

### DIFF
--- a/src/components/Section/History/Employment/EmploymentActivity.jsx
+++ b/src/components/Section/History/Employment/EmploymentActivity.jsx
@@ -176,7 +176,7 @@ export default class EmploymentActivity extends ValidationElement {
           </Field>
           <Show when={this.props.value === 'Other'}>
             <Field
-              title={i18n.m('history.employment.default.activity.other.label')}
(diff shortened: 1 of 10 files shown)


joshmfranklin:
The fail seems to be linting error about indentation but the file that it is complaining about is empty `treatment.scss`. Any suggestions? Should I just get rid of the file?

gidjin:
I took a look and seems like Treatment.scss isn't empty but had a tab instead of two spaces. I made the change and the lint errors seem to have resolved locally.

PR #1439, opened 2019-02-13
Form Type, Form Section constants and config

### LINKED ISSUE #1423: Create form data structure constants
Add constants and/or config files that can act as a single source of truth for form data structure and any requirements that differ between form types.

suzubara:
## Description

Goals of this PR are:
- Provide a starting point for clearer and more flexible organization when it comes to defining and structuring form types and their sections/subsections.
- Remove dependency on single `navigation` config file, replace with pure functions that take the current formType into account.
- Default formType to `SF86` to avoid breaking current functionality.
- Fixes #1423 

## Outstanding
- [x] Create config for remaining form sections, form types
- ~~[ ] Add globally-accessible `formType` value to Redux store, defaulting to `SF86`~~ - TBD in https://github.com/18F/e-QIP-prototype/pull/1378
- ~~[ ] Modify and/or replace existing uses of `navigation.js` to handle different form types~~ - TBD in separate story

## Checklist for Reveiwer

- [ ] Review code changes


src/config/formSections.js
```diff
@@ -0,0 +1,82 @@
+import * as sections from '../constants/sections'
+
+const IDENTIFICATION = {
+  key: sections.IDENTIFICATION,
+  name: `identification`,
```
christopherhuii: Are we defaulting to back ticks for all strings? My main objections are that there's something about readability and intent when using backticks vs regular quotes.

When I see back ticks, I'm expecting something more, like interpolation or a template.
When I see a regular quote, there's nothing smart; it's just a string.

...and typing back ticks are a pain.

I don't prefer this change, but I understand the reasoning and I want to make sure we're all on the same page for the sake of consistency.

suzubara: I can revert to single quotes, it's not something I feel strongly about one way or the other

christopherhuii: Let's stick with single quotes unless there's a need for a template literal.

Thanks!

christopherhuii:
So far so good to me. The general structure looks good to me. I'm still unsure of how each subsection will have the correct questions/locale for their form type, but I'm not sure if that should be a concern at this level.

gidjin:
Agreed with Chris, these look good so far.

christopherhuii:
We're missing the Review section. I'm not exactly sure what is needed for the Review section, so it's something we need to look into a bit more.

suzubara:
This looks great, thanks for finishing out the sections @christopherhuii ! I think it's fine to proceed without the Review section for now, I'm sure as we get there it'll become clear what kind of config we'll want there.

Issue #123, opened 2018-05-23
Adding attachments - suggested icon change


ryanhofdotgov:
From e-QIP-prototype-truetandem created by [Jay-Neighbours-I72](https://github.com/Jay-Neighbours-I72) : ryanhofdotgov/e-QIP-prototype-truetandem#2940  

Users were confused with the "Other" option icon. They thought it was email.

- [ ] Suggest changing to a generic file icon or mailbox

![screen shot 2018-05-01 at 3 35 46 pm](https://user-images.githubusercontent.com/19935974/39490056-799da6d8-4d55-11e8-834b-ac456e05085e.jpg)
![screen shot 2018-05-01 at 3 35 54 pm](https://user-images.githubusercontent.com/19935974/39490057-79aa816e-4d55-11e8-9299-ac0ab91fb4fb.jpg)

PR #1447, opened 2019-02-19
[WIP] Remove unused state from form components

suzubara:
## Description

## Checklist for Reveiwer

- [ ] Review code changes


suzubara:
Putting this on the back burner for now because removing internal state from inputs requires them to be controlled, which requires consistent passing of a `value` prop, and this is opening a can of worms because of how tangled/interdependent the input components are, as well as how some components children can be rendered using `cloneElement`.

I'd recommend writing new input components to replace existing ones over time instead, which will be less likely to introduce regressions, plus may allow us to incorporate some of @joshmfranklin 's design updates. cc @gidjin @christopherhuii

PR #1449, opened 2019-02-19
Move AuthenticatedView hoc to top level

### LINKED ISSUE #1345: Move AuthenticatedView HOC to top-level components
This is a technical refactor of the code Javascript code. There is not a functional feature gained from this issue. This refactor might be necessary to unblock some of the authentication work @macrael is doing. In general this fix will simplify the complexity of the code. Theoretically, it will improve performance, but it is not clear whether those performance gains will be noticeable to a user.

There is a higher-order component (HOC) called `AuthenticatedView` in `src/views/AuthenticatedView.js`. This component checks if a user is authenticated. If a user is not authenticated, it dispatches an action that redirects a user back to the login page. This component is referenced 50+ times in the code, but it is not clear to me why it was implemented like this. This component wraps around several of the "Section" components like, `Psychological`, `Military`, `Relationships`, `Legal`, `History`, etc.

This seems overkill because the application doesn't need every section to check if a user is authenticated. It should be sufficient to wrap `AppWithForm` (`src/components/Main/AppWithForm.jsx`) because it is the top level component in eApp.

See `src/components/Main/Main.jsx`. This is where `AppWithForm` is instantiated.
```
          <Route
            exact
            path="/form/:section/:subsection*"
            component={AppWithForm}
            onEnter={this.onEnter}
          />
```

Removing the `AuthenticatedView` HOC from each section component and adding the HOC to the `AppWithForm` component isn't a huge lift. However, it is not clear how much work fixing the tests will be.

gidjin:
## Description

Fixes #1345 by removing all the references to AuthenticatedView from the form and adding it to the top level component AppWithForm.

## Checklist for Reveiwer

- [ ] Review code changes


src/components/Main/App.jsx
```diff
@@ -195,9 +195,9 @@ class App extends React.Component {
                           className="instructions mobile-hidden">
                           {i18n.t('app.instructions')}
                         </button>
-                        <Show when={this.props.authenticated}>
+                        <span>
```
christopherhuii: Do we actually need these `span` tags or are they only here to satisfy the snapshot tests? If it's only for the snapshot tests, I suggest we remove them.

gidjin: I pulled it in because that's what Show added, but looks like it doesn't buy us anything so I've removed it.

PR #1472, opened 2019-02-22
Refactor Citizenship render

### LINKED ISSUE #1454: Section Render Refactor - Citizenship section
Make same changes as https://github.com/18F/e-QIP-prototype/pull/1431 to the Citizenship section, including:
- Create `Intro, Review` components
- Create `CitizenshipConnector` for passing props into subsection components
- Use `<Route>` components to render `<Citizenship>` and its subsections
- Use `@config/formSections` config instead of `@config/navigation`
- Each subsection should extend `Subsection` instead of `SubsectionElement`
- General cleanup, use classnames, import aliases, remove unused props, etc.

christopherhuii:
Fixes #1454 

## Checklist for Reveiwer

- [ ] Review code changes


src/components/Section/Citizenship/Citizenship.jsx
```diff
@@ -1,198 +1,98 @@
...
-                {i18n.m('citizenship.intro.body')}
-              </Field>
-            </div>
-          </SectionView>
+    this.form = formTypes[props.formType]
+    this.section = this.form.find(section => (section.key === CITIZENSHIP))
+
+    // Each key in subsectionLibrary corresponds to the subsection
+    // name in the config file.
+    this.subsectionLibrary = {
```
christopherhuii: There are many subsections that are specific to a certain form. Because the Citizenship component will be shared amongst forms, I created this `subsectionLibrary` that has all the possible subsections. The `key` is the `name` attribute in the subsection config object and the `value` is the subsection component.

src/components/Section/Citizenship/Citizenship.jsx
```diff
@@ -1,198 +1,98 @@
...
-              {...this.props.Status}
-              section="citizenship"
-              subsection="status"
-              dispatch={this.props.dispatch}
-              onUpdate={this.handleUpdate.bind(this, 'Status')}
-              onError={this.handleError}
-              required={true}
-              scrollIntoView={false}
-            />
+    return this.section.subsections.map(subsection => (
```
christopherhuii: Because we don't want to expose routes that aren't applicable to a form/section, I generated the routes from the section config files. this will only expose the sections applicable to this form.

src/components/Section/Citizenship/Citizenship.jsx
```diff
@@ -1,198 +1,98 @@
...
-              {...this.props.Passports}
-              dispatch={this.props.dispatch}
-              onUpdate={this.handleUpdate.bind(this, 'Passports')}
-              onError={this.handleError}
-              scrollToBottom={this.props.scrollToBottom}
-            />
-          </SectionView>
-        </SectionViews>
+          <SectionNavigation
+            section={CITIZENSHIP}
```
christopherhuii: In previous sections, we imported all the sections. Since that's unnecessary, I only imported the Citizenship section.

src/components/Section/Citizenship/Citizenship.jsx
```diff
@@ -1,198 +1,98 @@
...
   return {
     Application: app,
     Citizenship: citizenship,
     Status: citizenship.Status || {},
     Multiple: citizenship.Multiple || {},
     Passports: citizenship.Passports || {},
     Errors: errors.citizenship || [],
-    Completed: completed.citizenship || []
+    Completed: completed.citizenship || [],
+    formType: auth.formType
```
suzubara: This should also return `...state.section` so `Citizenship` receives the current subsection (see https://github.com/18F/e-QIP-prototype/blob/develop/src/components/Section/Identification/Identification.jsx#L73). Also, since the only props Citizenship uses are `subsection, formType`, the others can be removed.

src/components/Section/Citizenship/Intro/index.js
```diff
@@ -0,0 +1,2 @@
+import Intro from './Intro'
+export default Intro
```
suzubara: Do we want to keep up this pattern of exporting from a component directory's `index.js`? I'm happy either way, just want to know if I should be doing this also!

christopherhuii: I'd prefer:
```
Intro
    index.jsx
    index.test.jsx
```

I just followed the same format for the sake of consistency.

PR #1479, opened 2019-02-25
add recusive subsection reduce function

christopherhuii:
## Description

Adds recursive function to get subsections for back/next navigation.

## Checklist for Reveiwer

- [ ] Review code changes

### DIFF
--- a/src/config/formTypes.js
+++ b/src/config/formTypes.js
@@ -42,16 +42,16 @@ export const SF85 = [
       formSections.MILITARY_SELECTIVE,
       formSections.MILITARY_HISTORY,
       formSections.MILITARY_FOREIGN,
-      formSections.MILITARY_REVIEW
-    ]
+      formSections.MILITARY_REVIEW,
+    ],
   },
   {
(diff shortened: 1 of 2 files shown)


christopherhuii:
With our current strategy, we are adding `<SectionNavigation />` at the bottom of every section. Is it possible to move it up a level into `src/components/Section/Section.jsx` or possibly, `src/views/Form/Form.jsx`?

I'm bringing it up here because it will probably affect where we flatten out the subsections.

suzubara:
to respond to your comment @christopherhuii yes I think that will make sense down the road, but recommend waiting for all section renders to be refactored so we have a full view of how that should look.

suzubara:
Also, for optimization sake I would recommend adding these exports:
```
export const FLAT_SF85 = reduceSubsections(SF85)
export const FLAT_SF86 = reduceSubsections(SF86)
// etc.
```

That way, `reduceSubsections` is executed at compile time instead of runtime. It's really more of a development helper function. This is possible since the structure of each form itself is not going to be changing within execution of the app.

PR #1480, opened 2019-02-26
Update paths for dynamic section back/next navigation

christopherhuii:
## Description

- updates the config file paths for section navigation
- updates `reduceSubsection` to add each section's full path to its object
- updates `SectionNavigation` to use the `fullPath` for back/next instead of `section` and `subsection`
- update `identification`, `citizenship`, and `military` section to use new strategy

## Checklist for Reveiwer

- [ ] Review code changes


src/config/formSections/foreign.js
```diff
@@ -4,173 +4,171 @@ import { i18n } from '@config'
...
   storeKey: 'DirectActivity',
-  label: i18n.t('foreign.subsection.activities.direct')
+  label: i18n.t('foreign.subsection.activities.direct'),
 }
 
 const FOREIGN_ACTIVITIES_INDIRECT = {
   key: sections.FOREIGN_ACTIVITIES_INDIRECT,
   name: 'indirect',
-  path: `${FOREIGN_ACTIVITIES.path}/indirect`,
+  path: 'activities//indirect',
```
suzubara: There's an extra `/` here

christopherhuii:
I'm struggling to see the advantage to have the path be `direct` versus `activities/direct`.

I prefer `activities/direct` because that's how react router is parsing the params right now. The section is `foreign` and subsection is `activities/direct`. It also makes some sense because the subsections is named `FOREIGN_ACTIVITIES_INDIRECT`, so the concept of nesting is built into the naming.

With this way, we remove the need to reference the parent to build the URL.

suzubara:
@christopherhuii My reasoning is that if the path is set as `activities/direct`, then the form object nesting is used to determine the structure for the top sections, but not for subsections. My point in the last conversation we had about this was that whatever method we use, should be consistent no matter how many levels deep the sections are nested. The complete URLs (ie `/form/foreign/activities/direct`) can be constructed & appended to each object as part of the flatten method you wrote.

src/components/Section/Identification/Identification.jsx
```diff
@@ -21,22 +23,51 @@ import Review from 'components/Section/Identification/Review'
...
- * - subsection prop is not defaulting to "intro" after login, this is prob related to keeping redux in sync with routes. Investigate & fix this.
+ * - subsection prop is not defaulting to "intro" after login.
+ * this is prob related to keeping redux in sync with routes. Investigate & fix this.
  */
 
 class Identification extends React.Component {
+  constructor(props) {
+    super(props)
+
+    this.form = formTypes[props.formType]
```
suzubara: I wouldn't set a class attribute based on props, which are subject to change.

src/helpers/navigation.js
```diff
@@ -2,34 +2,26 @@
...
-  } else if (sectionIndex < formSections.length - 1) {
-    // Go to the first subsection of the next section
-    const nextSection = formSections[sectionIndex + 1]
-    next = nextSection.subsections[0] /* eslint prefer-destructuring: 0 */
-    next.navLabel = i18n.t(`${nextSection.name}.destination.${next.name}`)
+  if (next) {
+    // TODO: Fix this hack; should work for now
+    // It works because the section is always after /form/
+    const sectionName = next.fullPath.split('/')[2]
+    next.navLabel = i18n.t(`${sectionName}.destination.${next.name}`)
```
suzubara: 👍 For a long term solution, we can update the `i18n` keys to use the section key

src/components/Section/Identification/Identification.jsx
christopherhuii: Can it change? I was under the impression that once you login, you're viewing a specific form. You can't change forms during your session. A login has access to 1 type of form.

suzubara: Right, that is true bc of the business requirements of this app, however `props` by definition can change, so this pattern wouldn't be compatible with the expected lifecycle of a React component. Although it also seems like this (and all of the top level Section components) could be a pure functional component instead, which would make this a moot point.. what do you think?

christopherhuii: I don't think this would be an issue because the form type should never change even though props can. I thought of similar to initializing initial state and never changing it (because we shouldn't be). This way we'd be avoiding extra (even if they're cheap) operations.

However, I think you're right; these can be pure functional components.

christopherhuii: I've updated the component. Let me know your thoughts

PR #1481, opened 2019-02-26
Dynamic params in History section

### LINKED ISSUE #1474: Make History total years requirement dynamic based on form type
Implement the requirement that the History section requires different time frames based on the form type for Residence, Education, Employment

suzubara:
## Description
- Fixes #1474 
- Work required to make the number of years required in the History section dynamic based on form type
- Added `i18next` library, `util/i18n.js` to start migrating from `config/i18n`
- Define & pass down years parameters based on `props.formType` in History sections
- Use year parameters for i18n strings that need it
- Use year parameter in `totalYears` function
- Employment items show reasons for leaving if employed within Employment Record # of years (i.e., 5 or 7)
- Employment items `ReasonLeft, Reprimand` use Employment Record # of years
- Lots of lint fixes, remove binding to `this`, use `classnames`, etc. (general cleanup)
- Add `messageString, titleString` to `Field message` render fn to support i18n with interpolation
- **You should be able to switch form types by adding `?formType=SF85` in the URL, and see the # of years required in the History subsections change.** This does not include changes in the sidebar nav or questions specific to a form type.

## Checklist for Reviewer

- [ ] Review code changes


suzubara:
@christopherhuii @gidjin I'd love initial feedback on this approach I'm proposing for the i18n changes:
- Added i18next library (https://www.i18next.com)
  - Note, there *is* a React-specific i18next integration, however I decided against using it because the latest version requires React 16+, and it seems to be a major rewrite from the previous version that supports React 15. Since using the vanilla i18n library is very similar to what we currently have, it seemed easier to just go with that initially, rather than try to implement the legacy `react-i18next` and then update it somewhere down the road.
- Wrote new `util/i18n.js, util/markdown.jsx` util files
  - I decided on writing a new `i18n` export altogether that maps very similarly to the existing one, but with the changes needed to use `i18next` and its features. My reasoning was that I wanted to avoid adding onto or modifying the existing `i18n` since it's heavily-used, and if we approach updating `i18n` calls on a per-file basis, migrating it to the new one should move along more quickly than if we were to update on a per-implementation basis. Similar to the new `eslint` rules, this is putting the stake in the ground that any file that needs to be updated to use the new `i18n` methods will use it for all methods in that file, instead of mixing the legacy and new.
- Implemented `i18next` call using interpolated variables in `ResidenceWrapper` to demonstrate how we can pass dynamic content into a string. I will continue to use this throughout the History section before this PR is ready for review.

src/components/Section/History/Residence/ResidenceWrapper.jsx
```diff
@@ -1,77 +1,87 @@
...
+  const years = 10
+
+  return (
+    <div>
+      <h1 className="section-header">
+        {i18n.t('history.residence.title')}
+      </h1>
+
+      <Field
+        title={i18n.t('history.residence.info', { years })}
```
suzubara: Calling this out as the `i18n.t` call with an interpolated variable

src/config/locales/en/history.js
```diff
@@ -54,12 +56,12 @@ export const history = {
...
     employment: 'Employment history',
     education: 'Schools & diplomas',
-    federal: 'Former federal service'
+    federal: 'Former federal service',
   },
   residence: {
     title: 'Where you have lived',
     info:
-      'List the places where you have lived beginning with your present residence and working back 10 years.',
+      'List the places where you have lived beginning with your present residence and working back {{years}} years.',
```
suzubara: Calling this out as the localization string that uses an interpolated variable

src/config/locales/en/history.js
```diff
@@ -453,12 +453,6 @@ export const history = {
         additionalActivity: 'Additional periods of activity with this employer',
       },
       para: {
-        done:
```
gidjin: Were these (para.done, para.exiting, para.employment) unused?

src/config/locales/en/history.js
suzubara: `i18next` will log a warning, and the variable will be replaced with an empty string (so in this example, effectively "...and working back  years."). It does seem that you can set some global defaults, which we could use, although I've been err-ing on the side of setting defaults as props in the components that are using interpolation.

src/components/Section/History/Residence/ResidenceWrapper.jsx
```diff
@@ -27,14 +27,16 @@ const ResidenceWrapper = (props) => {
 
   const residenceHasGaps = sectionHasGaps(residenceItems)
 
+  const years = 10
```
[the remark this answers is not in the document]
suzubara: That value is now used wherever a static `10` was being used, and ultimately is passed into a function (`components/Section/History/helpers.js totalYears`) which uses the birthdate value to determine the actual # of years required.

src/config/locales/en/history.js
suzubara: Yup, I did a global search for them and they are not being used. These were `history.employment.default.para...`, not to be confused with `history.employment.para...` 🤦‍♀️

gidjin: I can't find any references either, looks like this is fine. There are a couple of places that the keys for i18n are dynamically assembled but these don't seem to be any of those.

PR #1490, opened 2019-02-27
Make Date element serialization more defensive

ryanhofdotgov:
Similar to #1487. 

It seems to be possible to serialize form data in such a way where `month` property is not set on a date in the JSON serialization, but I have not been able to replicate this using the UI. 

The previous "lazy" version of the code relied on `month` being set to empty string if `not applicable` was indicated for the granted/investigation dates. This PR makes the code more defensive and makes less assumptions.

It also makes a fix to an adjacent investigation element – a clearance level of `None` was not correctly serialized.


ryanhofdotgov:
@macrael Yeah, now that I realize just how many of these variants I'm going to have to do, I'm going to revisit the function names used in that previous address-related PR.

PR #1504, opened 2019-03-05
Update U.S. visa to U.S. Visa (red foil number) on citizenship status…

### LINKED ISSUE #1503: U.S Visa should be U.S. Visa (red foil number) in the Citizenship status section
## Where are you in the form?

**Section:**
Citizenship > Citizenship status

**Question:**
Provide type of document issued

**Field:**

Radio button block

## What did you see?
It only says `U.S. Visa` on the SF-86 and it should change to `U.S. Visa (red foil number)` for the SF-86 and SF-85
- [ ] Validation error
- [ ] Data not saving
- [x] Language or content issue
- [ ] Other (explain below)

Expected is to say `U.S. Visa (red foil number)`

- [x] Only relevant to just this question
- [ ] Cross-form issue
- [ ] I don't know

**Operating system**
- [ ] Windows
- [x] MacOS

**Browser**
- [x] Chrome
- [ ] Firefox
- [ ] IE
- [ ] Edge
- [ ] Other

**Version:**

joshmfranklin:
fixes #1503

### DIFF
--- a/src/components/Section/Citizenship/Status/Status.jsx
+++ b/src/components/Section/Citizenship/Status/Status.jsx
@@ -840,7 +840,7 @@ export class Status extends Subsection {
               adjustFor="buttons"
               scrollIntoView={this.props.scrollIntoView}>
               <RadioGroup
-                className="citizenship-document-type"
+                className="citizenship-document-type option-list option-list-vertical"
                 required={this.props.required}
                 onError={this.handleError}
                 selectedValue={(this.props.DocumentType || {}).value}>
@@ -855,7 +855,7 @@ export class Status extends Subsection {
                 <Radio
                   name="document-type-visa"
                   label={i18n.t('citizenship.status.label.documenttype.visa')}
-                  value="U.S. Visa"
+                  value="U.S. Visa (red foil number)"
                   className="document-type-visa"
                   onUpdate={value => { this.updateField('DocumentType', value) }}
                   onError={this.handleError}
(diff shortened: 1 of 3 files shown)


src/components/Section/Citizenship/Status/Status.jsx
```diff
@@ -855,7 +855,7 @@ export class Status extends Subsection {
                 <Radio
                   name="document-type-visa"
                   label={i18n.t('citizenship.status.label.documenttype.visa')}
-                  value="U.S. Visa"
+                  value="U.S. Visa (red foil number)"
```
christopherhuii: @ryanhofdotgov @gidjin I'm not exactly sure how to word this, but shouldn't the value still be `U.S. Visa`?

Basically the issue is that @joshmfranklin realized was that the label for `U.S. Visa` was wrong. See image below:
screen shot 2019-03-05 at 12 21 10 pm

The label on the checkbox can be different from the value we send to the backend. Should the value we send change to `U.S. Visa (red foil number)`? Or should it remain as `U.S. Visa`?

ryanhofdotgov: The value should remain `U.S. Visa` – the legal-approved text shouldn't affect the underlying way we encode the choices.

A larger annoyance is that the enums/checkbox values across the form are often needlessly different from what the SF-86 XML uses, so we have a layer of mapping from eApp to eQIP enum/constant values. e.g., "Top Secret" vs "TopSecret". At some point it would be nice to unify those.

PR #1535, opened 2019-03-07
Pass username, URL, JWT on commandline

### LINKED ISSUE #1475: Add JWT option, etc. to load-scenario
Enhance the `load-scenario` tool to support testing in the SAML configuration used in the partner agency integration environment.

ryanhofdotgov:
Closes #1475

### DIFF
--- a/api/cmd/load-scenario/main.go
+++ b/api/cmd/load-scenario/main.go
@@ -1,14 +1,59 @@
 package main
 
 import (
+	"bufio"
 	"encoding/json"
+	"flag"
+	"fmt"
 	"log"
 	"net/http"
 	"os"
+	"os/user"
+	"path/filepath"
+	"strings"
 
 	"github.com/18F/e-QIP-prototype/api"
 	"github.com/18F/e-QIP-prototype/api/cmd"
 )
+
+const passwordFilename = ".eapass"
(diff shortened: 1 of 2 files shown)


ryanhofdotgov:
@macrael I started to go down a rabbit hole with respect to refactoring the web client code and automated testing – but I ripped that out, and decided to just keep it to strictly what we needed for expediency purposes and will re-visit at a later time.

ryanhofdotgov:
@macrael Yeah, you're going to have to change it almost every invocation :/ . Once you update the backend with the the session changes you've proposed, it could be less severe, but similar. 

It is awkward, but something like:
```
echo -n "test02: Bearer" > ~/.eapass ; \
cat >> ~/.eapass ; \
./api/bin/load-scenario -U test02 -url http://api.eapp.local.test:3000 api/testdata/complete-scenarios/test1.json
```
will block until you paste the token value in from the browser console.

Traditionally, the secrets passed on the command-line would leak from one account to another on a shared system via the process listing, but for a project managed in the open like ours, the risk is in the publicly available CircleCI logs. 

I believe the cloud.gov credentials are passed to the CF tools in a fashion similar to this PR. By simply not building in the option to pass the credentials on command-line, we help prevent another engineer from shooting themselves in the foot later on.

PR #1541, opened 2019-03-11
Add SFType and SFVersion to Accounts

macrael:
## Description
* Add two new columns to the accounts table
  * there is a default value that is then unset to
    let the columns be required but not choke on
    existing data
* Add a check on account.Save() to prevent saving
  unknown form versions
* Add a command line tool to read/set the form version
  in development.

This is the fix for https://github.com/18F/e-QIP-prototype/issues/1429

## Checklist for Reveiwer

- [ ] Review code changes


macrael:
I added some tests to the integration_test package, my intention is for that to grow some helper functions and be the basis of testing the app going forward. I think that using the db in testing is critical rather than spending a lot of time mocking it out.

Also, the code that inserts the metadata section into the /form request is a bit ugly but I think it's worth having the clean metadata section rather than having it implement the Payload format.

api/integration_test/form_version_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+	metadataBody := `
+{
+	"type": "metadata",
+	"form_type": "SF86",
+	"form_version": "2016-11"
+}`
+
+	// create request/response
+	r := httptest.NewRequest("POST", "/me/save", strings.NewReader(metadataBody))
```
ryanhofdotgov: I thought these pieces of meta-data were going to be read-only? i.e., front-end wouldn't be setting them.

api/migrations/20190305160606_add_sf_type_and_version.sql
```diff
@@ -0,0 +1,16 @@
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+-- +goose StatementBegin
+ALTER TABLE accounts ADD COLUMN sf_type VARCHAR(25) NOT NULL DEFAULT 'SF86';
```
ryanhofdotgov: I think we need to leave the backend be the arbiter of what form gets assigned and not have it spread across multiple components. For now, the backend is going to assign this default, but eventually it might come over from an external service request.

api/migrations/20190305160606_add_sf_type_and_version.sql
```diff
@@ -0,0 +1,16 @@
...
+ALTER TABLE accounts ALTER COLUMN sf_type DROP DEFAULT;
+ALTER TABLE accounts ADD COLUMN sf_version VARCHAR(25) NOT NULL DEFAULT '2016-11';
```
ryanhofdotgov: Just a comment, not asking for a change. I'm not a big fan of varchars with fixed limits in the schema, though that is in keeping with the rest of the eApp schema. See:
https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_varchar.28n.29_by_default

api/application.go
```diff
@@ -596,7 +620,11 @@ func PurgeAccountStorage(context DatabaseService, account int) {
 
 // Hash returns the SHA256 hash of the application state in hexadecimal
 func Hash(context DatabaseService, account int) string {
-	hash := sha256.Sum256(Application(context, account, true))
+	jsonBytes, err := Application(context, account, true)
+	if err != nil {
+		return ""
```
ryanhofdotgov: This only seems marginally better than just continuing to send a (bad) hash in the event of an error. There's no indication via function signature that the caller needs to check for an error. And it looks like while one call point was updated to check for the empty string, there is at least another where it was not updated.

This change motivated my questions about recover/panic and if there was a way to just catch/log these type of errors at the top-level HTTP request handler vs having to check them at every call up the stack.

api/cmd/sftype/test_sftype.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# Usage: find . | entr -c docker exec e-qip-prototype_api_1 ./cmd/sftype/test_sftype.sh
```
ryanhofdotgov: Cool – this is a good first step towards something more robust that we can copy for some of our other tools.

api/integration_test/form_version_test.go
```diff
@@ -0,0 +1,204 @@
...
+	account := api.Account{
+		Username:  "buzz1@example.com",
+		Email:     "buzz1@example.com",
+		SFType:    "SF86",
+		SFVersion: "2016-11",
+	}
+
+	_, err := account.Get(services.db, -1)
+	if err != nil {
+		if err.Error() == "pg: no rows in result set" {
```
ryanhofdotgov: Can we pull this into a constant defined in the package somewhere?

api/integration_test/form_version_test.go
```diff
@@ -0,0 +1,204 @@
+package integration_test
```
ryanhofdotgov: Really nice to see more inroads regarding integration tests – thanks! How do you call them?

api/migrations/20190305160606_add_sf_type_and_version.sql
ryanhofdotgov: To be clear – I would remove the defaults from the table schema and just have default logic at the app layer.

ryanhofdotgov: The update script should set the column values for any existing rows to be SF-86, 2016-11.

api/integration_test/form_version_test.go
macrael: Correct, this test just confirms that attempting to set them results in an error.

api/integration_test/form_version_test.go
macrael: I'm making it an explicit error in my next PR, i'll port that over here.

api/integration_test/form_version_test.go
macrael: go test ./api/integration

api/application.go
macrael: I think that using panic is possible, but is not common at all in go code. Go is pedantic in this way but very "correct". I also didn't really like that this was returning an empty string instead of an error, I'll make it an explicit err like everything else.

api/migrations/20190305160606_add_sf_type_and_version.sql
macrael: Having the default in the table logic is the only way to have NOT NULL work at the table layer. The current DB has almost no db level validation logic and I think that's a mistake. I figured this would be a good place to start experimenting with enforcing some db requirements. 

I can undo that and have the startup script set things instead and do all validation logic in the backend layer instead but I prefer this for now.

Eventually, we are going to have to have a "forms" table and this data can live there instead of on the account. But I believe that the form type is passed in during account provisioning so I think that this not null requirement can be enforced for now.

api/migrations/20190305160606_add_sf_type_and_version.sql
macrael: I can switch this to text, its an old habit that probably needs to die.

api/migrations/20190305160606_add_sf_type_and_version.sql
ryanhofdotgov: Yeah, we are on the same page in getting more constraints in the db – we should keep the not null. 

But in the update script, you should be able to add the columns, populate them with the defaults, then add the not null constraint after that.  You could even define these columns as an enum type that would constrain the value range to the map you have defined in the app layer.

To be clear, I'm just saying that the decision as to what is the default form type should solely be made at the app layer.

api/integration_test/form_version_test.go
macrael: i'll add instructions to the readme

ryanhofdotgov: Cool – can we get a Makefile target and/or an entry in `advanced.md`, etc. If we are having to manually run them at the moment, we probably want to require it as part of release process described in `releases.md`.

macrael: They are being run in CI!

PR #1551, opened 2019-03-13
Refactor Psychological section

### LINKED ISSUE #1460: Section Render Refactor - Psychological section
Make same changes as https://github.com/18F/e-QIP-prototype/pull/1431 to the Psychological section, including:
- Create `Intro, Review` components
- Create `PsychologicalConnector` for passing props into subsection components
- Use `<Route>` components to render `<Psychological>` and its subsections
- Use `@config/formSections` config instead of `@config/navigation`
- Each subsection should extend `Subsection` instead of `SubsectionElement`
- General cleanup, use classnames, import aliases, remove unused props, etc.

suzubara:
## Description

- Fixes #1460 

## Checklist for Reveiwer

- [ ] Review code changes
- [ ] Ensure form fills out and validates the same as branch develop
- [ ] Ensure back/next navigation navigates and renders the correct sections


christopherhuii:
I can't click `Yes` or `No` on the `Existing Conditions` subsection. My gut suspicion is that there's a typo somewhere

christopherhuii:
I introduce the problem because I entered the wrong `storeKey` when creating the section config files. I also pushed up a fix for the merge conflict.

christopherhuii:
Currently on develop, the psychological review section doesn't have a "next" button to get to the Review "section". This makes sense because we don't have the "Review" defined in the form config.

Should we add the "Review" stuff for the form in this PR or another?

suzubara:
@christopherhuii good catch! I added the "Review" section in the nav PR, but I do think it makes more sense to pull it into this one so the functionality isn't interrupted. I will do that.

src/helpers/navigation.js
```diff
@@ -15,14 +15,16 @@ export const getBackAndNext = (state, { currentPath }) => {
   if (back) {
     // TODO: Fix this hack; should work for now
     // It works because the section is always after /form/
-    const sectionName = back.fullPath.split('/')[2]
+    let sectionName = back.fullPath.split('/')[2]
```
suzubara: Had to do some weirdness here bc `package` is a reserved word in JS, and so the localization key is different. I think we can make this better at a later date.

Issue #1327, opened 2019-01-08
Move U.S. Passport subsection information to Citizenship section


joshmfranklin:
As an applicant
I want to be able to fill out my U.S. Passport Information when I am in the Citizenship section
Because it makes the most amount of sense while answering questions about my citizenship

Given I am in the Citizenship section of any form (85, 85P, 86)
When I complete the introduction subsection and click next
I should be taken to the U.S. Passport Information subsection

Note: This helps keep things consistent between form. Currently the U.S. Passport Information subsection is within the Foreign associates section, in the SF85 no other Foreign association subsections are present to U.S. Passport Information would be the only one. Moving it to Citizenship in all forms makes sense given the other subsections and keeps things consistent between forms. 

The new order of the Citizenship section would be:
- Introduction
- U.S. Passport information
- Citizenship status
- Dual/multiple citizenship
- Foreign passports
- Review

christopherhuii:
@joshmfranklin 

> Given I am in the Citizenship section of any form (85, 85P, 86)
> When I complete the Dual/multiple citizenship subsection and click next
> I should be taken to the U.S. Passport Information subsection

> The new order of the Citizenship section would be:
>
> - Introduction
> - U.S. Passport information
> - Citizenship status
> - Dual/multiple citizenship
> - Foreign passports
> - Review

I'm getting confused by those two statements. Which is the correct order?

joshmfranklin:
@christopherhuii Whoops sorry, I updated the bullet points but not the paragraph above
The order should be same as bullet points:
Introduction
U.S. Passport information
Citizenship status
Dual/multiple citizenship
Foreign passports
Review

ryanhofdotgov:
Some notes from our standup today:

Moving this question to a different section in the most direct manner has quite a long tail of activities:
* update backend JSON structures
* update backend complete test scenarios
* update XML code generation
* re-submit/revalidate against e-QIP
* address issue when old data from previous eApp is loaded into revised eApp frontend/backend

Suz/Chris/and I decided a couple of things:
1. It was preferable to move the question in the UI, but in a way that does not alter backend/JSON format. This in and of itself, would be treading new ground, as historically the UI and JSON serialization format are very tightly coupled.
1. Regardless of approach, this work has too much risk to incorporate into the 1.0.14 sprint/release, as we want that to be a very stable release, resolving most/all the major issues discovered in pre-prod testing rounds.

PR #1553, opened 2019-03-13
Update requiredness for US Citizen Foreign Born Documenation

### LINKED ISSUE #1505: Foreign born document number requirements
## Where are you in the form?

**Section:** Citizenship

**Question:** Provide your current citizenship status

**Field:** [Selected] I am a U.S. citizen or national by birth, born to U.S. parent(s), in a foreign country

**Component (for example address block):** Radio select

## What did you see?

Document number is required.

- [x] Validation error
- [ ] Data not saving
- [ ] Other (explain below)

*Branching needs to exist.*

If applicant selects `Other`:
1) We require a written response (this works as expected) 
2) Document number should not be required (we could also add an "I don't know" button)
3) The next four fields (DocumentNumber/Date/Place/Name) of document should be eliminated.

</br>

If applicant selects `FS-240, DS-1350, or FS-545`:
1) Document number should ~not~ be required (This is required by e-QiP when one of these three forms are selected)
2) The next three fields (Date/Place/Name) should appear (this works as expected)

*Citizenship number*
If applicant provides citizenship number (not required) then Date and Name on the certificate are required. 

 

- [x] Only relevant to just this question
- [ ] Cross-form issue
- [ ] I don't know

**Operating system**
- [ ] Windows
- [x] MacOS

**Browser**
- [x] Chrome
- [ ] Firefox
- [ ] IE
- [ ] Edge
- [ ] Other

**Version:**

gidjin:
## Description

Fixes #1505 to update allowing a user to skip one form if another is selected.

## Checklist for Reveiwer

- [ ] Review code changes


gidjin:
This updates required for Citizenship section, there are two follow on stories that are not part of this. US Passport override is not part of this and will be taken care of after #1327 and #1568

christopherhuii:
When clicking on `I am a U.S. citizen or national by birth, born to U.S. parent(s), in a foreign country`, it errors on every single question. This does not happen if any other option is selected.

ezgif com-video-to-gif

christopherhuii:
> If applicant selects Other:
> 
> 1. We require a written response (this works as expected)
> 2. Document number should not be required (we could also add an "I don't know" button)
> 3. The next three fields (Date/Place/Name) of document should be eliminated.

It's hard to tell whether 2 is working because of the `Your response is required` error. My guess would be no because it has that error.

3 isn't fulfilled because I can still see the fields for date, place, and name of the document.

src/components/Section/Citizenship/Status/Status.jsx
```diff
@@ -76,12 +79,47 @@ export class Status extends Subsection {
...
   }
 
-  updateField = (field, values) =>  {
+  updateField = (field, values) => {
     this.update({
       [field]: values,
     })
   }
 
+  isRequired = () => {
```
suzubara: This logic feels like it should live somewhere so it can be shared across the validation function & this component

src/validators/citizenship.js
```diff
@@ -1,7 +1,135 @@
...
+  certificateIssued,
+  certificateName,
+}) => validGenericTextfield(certificateNumber)
+    && validDateField(certificateIssued)
+    && new NameValidator(certificateName).isValid()
+
+export const isCertificatePartial = ({ certificateNumber, certificateIssued, certificateName }) => {
+  if (!_.isEmpty(certificateNumber) && !_.isEmpty(certificateNumber.value)) {
+    return !validateCertificate({ certificateNumber, certificateIssued, certificateName })
+  } if (!_.isEmpty(certificateIssued) && (
```
suzubara: This seems like a redundant use of `_.isEmpty` if you have to check each of the object's keys as well

gidjin:
@suzubara & @christopherhuii I believe I've addressed all the above issues please review again.

src/components/Section/Citizenship/Status/Status.jsx
```diff
@@ -43,6 +44,34 @@ export class Status extends Subsection {
...
+      documentNumber: DocumentNumber,
+      documentIssued: DocumentIssued,
+      placeIssued: PlaceIssued,
+      documentName: DocumentName,
+      certificateNumber: CertificateNumber,
+      certificateIssued: CertificateIssued,
+      certificateName: CertificateName,
+    }
+    this.isCertificateRequired = isCertificateRequired(data)
+    this.isDocumentRequired = isDocumentRequired(data)
```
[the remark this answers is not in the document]
gidjin: Thanks for the feedback, I've moved these.

ryanhofdotgov:
@suzubara @christopherhuii Just going to state the obvious in order to be explicit as I move the parent issue to the next sprint:
* We should defer merging this PR in until the next sprint.
* One of you will have to address any required changes as John off-boards. Thanks!

PR #1565, opened 2019-03-15
Fix issues with Substance Use

### LINKED ISSUE #1564: Substance component not found
Seeing this error on https://eqip-prototype-dev.fr.cloud.gov/ when trying to submit the form. 

Screen Shot 2019-03-15 at 12 00 04 PM

christopherhuii:
## Description
Fixes #1564 
- Fixes issues with `Negative Impact` subsection not saving
  - Fix: The naming was wrong. `NegativeImpact` should be `NegativeImpacts` (plural)
## Checklist for Reveiwer

- [ ] Review code changes

### DIFF
--- a/src/components/Section/Section.jsx
+++ b/src/components/Section/Section.jsx
@@ -25,7 +25,8 @@ const storeToComponentMap = {
   History,
   Legal,
   Psychological,
-  SubstanceUse,
+  // TODO: Redux, backend, XML, etc. all has Substance named as such instead of SubstanceUse
+  Substance: SubstanceUse,
   Package,
 }
 
--- a/src/components/Section/SubstanceUse/SubstanceUseConnector.jsx
+++ b/src/components/Section/SubstanceUse/SubstanceUseConnector.jsx
@@ -72,8 +72,8 @@ const connectSubstanceUseSection = (Component, {
         return { ...substance.OrderedTreatments, addressBooks } || {}
       case 'VoluntaryTreatments':
         return { ...substance.VoluntaryTreatments, addressBooks } || {}
-      case 'NegativeImpact':
-        return { ...substance.NegativeImpact } || {}
+      case 'NegativeImpacts':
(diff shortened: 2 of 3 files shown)


ryanhofdotgov:
Would you do a little digging into root cause/source merge that caused this regression and what we need to put in place to prevent similar problem going forward?

christopherhuii:
This situation is probably the result of carelessness. The fields reflected changes, but didn't save because of a typo (`NegativeImpact` vs. `NegativeImpacts`). There's a bit of inconsistency in some of the field naming.

The `Substance component not found` console log wasn't an issue. That console log only shows up because we still have the non-refactored sections rendering the old way. Once all sections have been converted, all the old code will be removed.

ryanhofdotgov:
What do we need to change so we can catch this type of defect? Do we need our Selenium tests working, or is it possible to catch this at the unit test level?

src/components/Section/Section.jsx
```diff
@@ -25,7 +25,8 @@ const storeToComponentMap = {
   History,
   Legal,
   Psychological,
-  SubstanceUse,
+  // TODO: Redux, backend, XML, etc. all has Substance named as such instead of SubstanceUse
+  Substance: SubstanceUse,
```
suzubara: 👍 as we touch areas of the code that reference section/subsection names, I think we should gradually change them to all use the `key` values defined in the section constants to avoid issues going forward.

christopherhuii:
I'm not exactly sure how to write a test for this. It wasn't saving because the naming on the frontend and backend weren't matching up.

ryanhofdotgov:
@christopherhuii I don't want to drag this out longer than is useful, but it seemed like there were two symptoms of the root cause:
1) front-end wouldn't validate the form data even if an answer was supplied (this doesn't involve backend)
2) form data wouldn't save properly to backend

Shouldn't is be possible for  #⁠1 to be caught by a unit test?

PR #1566, opened 2019-03-18
Date comparison error messages

### LINKED ISSUE #1467: Error messages for comparison dates need to be correct
For some subsections we are comparing two dates within the same subsection to make sure that the second date can not come before the first. For a number of different sections the error message on the second date is displaying `The date should be after your date of birth.` These error messages should be specific to the date they should not come before. Below are the following subsections that need revised error messages:

- [x] Foreign associations > U.S. Passport information
- [x] Foreign associations > Foreign activities > Direct control
- [x] Foreign activities > Indirect control
- [x] Financial record > Bankruptcy
- [x] Financial record > Delinquent payments
- [x] Substance use > Illegal use of drugs and drug activity > Purchase

joshmfranklin:
Fixes #1467

### DIFF
--- a/src/config/locales/en/error.js
+++ b/src/config/locales/en/error.js
@@ -1376,5 +1376,15 @@ export const error = {
       title: 'There is a problem with your entry',
       message: 'The date of discharge of is before the date filed.'
     }
+  },
+  delinquent: {
+    required: {
+      title: 'Your response is required',
+      message: ''
+    },
+    min: {
+      title: 'There is a problem with your entry',
+      message: 'The date the issue was resolved is before the date the issue began.'
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


src/config/locales/en/error.js
```diff
@@ -1336,5 +1336,65 @@ export const error = {
...
+    }
+  },
+  drugPurchase: {
+    required: {
+      title: 'Your response is required',
+      message: ''
+    },
+    min: {
+      title: 'There is a problem with your entry',
+      message: 'The date of most involvement is before the date of first involvement.'
```
suzubara: @joshmfranklin Think this should read `The date of most recent involvement...`

PR #1584, opened 2019-03-22
Move U.S. Passport section from Foreign to Citizenship

### LINKED ISSUE #1327: Move U.S. Passport subsection information to Citizenship section

christopherhuii:
## Description
Fixes #1327 

This is a part of an effort to help users have a simplified experience when filling out their citizenship. It makes more sense for the U.S. Passport information to be in Citizenship rather than Foreign.

However, these aren't simple changes. In this PR:
- The user will see that the U.S. Passport section has moved from Foreign to Citizenship
- U.S. passport information is **still** stored under Foreign in the Redux store in order to avoid backend, database, and XML changes.
- No backend or database migrations have changed because we currently don't have a plan for backwards compatibility.
- **ONLY** the flow of how an applicant fills out the form changes.

Notes:
I have an issue with the naming throughout this application. The one letter differences aren't clear and can easily cause confusion and mistakes.
U.S. Passport information is recognized as `Passport` in Redux. and it's URL is `/form/citizenship/passport`

Foreign Passports information is recognized as `Passports` in Redux and its URL is `/form/citizenship/passports`

To add to the confusion, the backend recognizes U.S. passport information as `ForeignPassport` (because before this PR, Passport lived in the Foreign section).

There should be future iteration to rename all the things and make things 1:1 on the frontend and backend (e.g. backend will store U.S. Passport information in the citizenships table and generate correct XML). See #1608 

Note: There are some hacks in place to make this work. They have comments pointing to them to be deleted.
`src/selectors/validation.js`
`src/actions/ApplicationActions.js`

## Checklist for Reviewer

- [ ] Review code changes


christopherhuii:
**DO NOT MERGE AS A PART OF 1.0.14**

suzubara:
@christopherhuii re: your comments on the naming, I agree & that was my intention with adding the form section constants, to give us clear naming for each subsection that we can move to (and change if needed). where possible I think we should start trying to use those as identifiers instead of the section `name` or path

src/validators/validate.js
```diff
@@ -356,13 +140,12 @@ export const walkValidationTree = data => {
...
     // When the property is not specific to this instance
     // skip it and go to the next.
-    if (!data.hasOwnProperty(property)) {
-      continue
-    }
 
-    const result = walkValidationTree(data[property])
-    if (!result) {
-      return false
+    if (!data[property]) {
```
suzubara: This looks like it should be `if (data[property]) {...` since it's accessing that value?

suzubara:
@christopherhuii I'm getting redirected to the `/error` page when I try to login as a new user with no data

christopherhuii:
Ah got it. I think I fixed it

Issue #1522, opened 2019-03-06
Failures for 2019-03-05 pre-prod XML 8 of 9


ryanhofdotgov:
Pre-prod SF-86 8 of 9:
https://drive.google.com/open?id=12xYfbfdKNJAyu5hVIURXHzKe0tbf9EJn

Exceptions when submitted to e-QIP:
```
WARN[0002] gov.opm.eqip.ws.webservice.ApplicantDataException: SF86_5_S12-Education: Provide a response for "Telephone number". 
WARN[0002] gov.opm.eqip.ws.webservice.ApplicantDataException: SF86_5_S13A-EmploymentUnemploymentInformation: Provide a response for "Telephone number". 
WARN[0002] gov.opm.eqip.ws.webservice.ApplicantDataException: SF86_5_S14-SelectiveServiceRegistration: Answer "Have you registered with the Selective Service System (SSS)?" question yes, no or I don't know. 
```

ryanhofdotgov:
Looks like some of errors are also seen in https://github.com/18F/e-QIP-prototype/issues/1516 and https://github.com/18F/e-QIP-prototype/issues/1521

ryanhofdotgov:
Closing this as duplicate.

Issue #1521, opened 2019-03-06
Failures for 2019-03-05 pre-prod XML 7 of 9


ryanhofdotgov:
Pre-prod SF-86 7 of 9:
https://drive.google.com/open?id=1S-vQUPyeRPGCpp6OEacHwksmjZEMPR1w

Exceptions when submitted to e-QIP:
```
WARN[0004] gov.opm.eqip.ws.webservice.ApplicantDataException: SF86_5_S09-Citizenship: Do not provide a country for "Documentation of U.S. Citizen Born Abroad: Place of issuance" when selecting a state or territory. 
WARN[0004] gov.opm.eqip.ws.webservice.ApplicantDataException: SF86_5_S13A-EmploymentUnemploymentInformation: Provide a response for "Supervisor telephone number". 
WARN[0004] gov.opm.eqip.ws.webservice.ApplicantDataException: SF86_5_S14-SelectiveServiceRegistration: Answer "Have you registered with the Selective Service System (SSS)?" question yes, no or I don't know. 
```

ryanhofdotgov:
Looks like one of the errors is also in https://github.com/18F/e-QIP-prototype/issues/1516

ryanhofdotgov:
Exception 1 addressed by #1594, Exception 2 is a dup of #1516, and Exception 3 is resolved by #1596.

PR #1594, opened 2019-03-26
Correctly encode IDontKnow for Selective Service registration

ryanhofdotgov:
Partially addresses #1521 (selective service issue) and the corresponding error in #1522


ryanhofdotgov:
All 10 complete test scenarios were successfully validated against XSDs and e-QIP service.

PR #1619, opened 2019-04-04
Update redux store to use form data from /form endpoint

### LINKED ISSUE #1590: Use FormType and FormVersion in the /form meta section to switch form type
Reference:  EN-3431

Presently, we are using a GET URL param to switch which form type we are displaying in the frontend. Now that that is being returned in the /form call we should use that instead. 

There is a new section in /form called "Metadata" that looks like this:
```
{
  "Metadata": {
    "type": "metadata",
    "form_type": "SF86",
    "form_version": "2016-11"
  }
  ...
}
```

christopherhuii:
## Description
Fixes #1590 

Previously, we used the query params (e.g. `?formType=SF85`) to set the form type. This PR leverages:
```
Metadata: {
    form_type: '',
    form_version: '',
}
```
from `/forms` to populate the `formType` and `formVersion` in Redux.

Because the data comes with the form data and not from login, I moved the `formType` from `authentication` to `application`.

**Note:**
1. Query params (e.g.`?formType=SF85` ) will no longer work after this PR is merged. We need logins with different form types and versions to test the different forms/versions.
2. `formVersion` is stored in Redux, but it is not being used anywhere yet.
 

## Checklist for Reviewer

- [ ] Review code changes


src/actions/AuthActions.js
```diff
@@ -26,6 +27,17 @@ export function login(username, password) {
...
       .then((response) => {
+        /**
+         * In local development and staging, we have the ability to manually
+         * set the formType by using URL params.
+         */
+        if (env.IsDevelopment() || env.IsStaging()) {
+          const params = location.search
+          const query = queryString.parse(params)
+          if (query.formType) {
+            window.formType = query.formType
```
christopherhuii: This is a little hacky because by the time the form data is retrieved in `ApplicationActions`, but URL params are gone. This is the workaround.

src/actions/ApplicationActions.js
```diff
@@ -34,7 +34,9 @@ export function getApplicationState(done) {
         }
         return api.form().then((r) => {
           formData = r.data
-          dispatch(updateApplication('Settings', 'formType', formData.Metadata.form_type))
+          const formType = window.formType ? window.formType : formData.Metadata.form_type
```
christopherhuii: If there's a URL param `formType`, it always has higher priority than the `formType` that comes with the form data.

src/actions/AuthActions.js
suzubara: seems fine enough for now!

PR #1627, opened 2019-04-23
85 small copy changes [EN 3653]

joshmfranklin:
Completes EN-3653


src/components/Section/Package/Submit/__snapshots__/index.test.jsx.snap
```diff
@@ -91,7 +91,7 @@ Not a guarantee of acceptance, but all required fields are complete."
           onClick={[Function]}
           type="button"
         >
-          Submit your SF-86
+          Submit your
```
suzubara: @joshmfranklin Ideally this snapshot shouldn't change. This can be fixed by providing a `formType` prop to the test case. lmk if you want to look at it together!

ryanhofdotgov:
@joshmfranklin I'm thinking of some ways we might be able to have some semi-automated integration between the public PRs and the internal issues.

Would you mind referencing the internal issues using the `EN-` prefix? e.g., "Completes EN-3653"? I'll update this PR description to match that convention.

src/components/Section/Package/Submit/__snapshots__/index.test.jsx.snap
joshmfranklin: @suzubara yeah I'd love to look at that together!

PR #1629, opened 2019-04-23
Encode "I don't know" clearance level

### LINKED ISSUE #1489: Add support for "I don't know" clearance level in XML generation
After https://github.com/18F/e-QIP-prototype/issues/1488 is resolved fix legal-investigations-investigated.xml, line 16 to handle case where "I don't know" is provided as answer to what clearance level was received. 

Alter test6 to use that as a value for clearance level in the investigation history entry. Re-validate against XSDs, re-submit to e-QIP test instance.

ryanhofdotgov:
Closes #1489. Update test6 so it contains an "I don't know" clearance level. Successfully validated by e-QIP.

### DIFF
--- a/api/templates/legal-investigations-investigated.xml
+++ b/api/templates/legal-investigations-investigated.xml
@@ -13,8 +13,11 @@
     <OtherAgency>{{text $Item.AgencyExplanation}}</OtherAgency>
     {{end}}
     {{end}}
-    <!-- XXX https://github.com/18F/e-QIP-prototype/issues/1489 -->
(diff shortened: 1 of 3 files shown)


macrael:
Why is "not applicable" rendered as "unknown"?

ryanhofdotgov:
> Why is "not applicable" rendered as "unknown"?

In the UI, this particular not applicable widget has the text "I don't know". 

As far as e-QIP: it encodes this "I don't know" a little differently that almost anywhere else in the XML – this element doesn't have an `IDontKnow` attribute but instead has "Unknown" as one of the enumeration values.

PR #1631, opened 2019-04-25
Sets CORS Access-Control-Max-Age header

ryanhofdotgov:
Closes EN-3143

Sets CORS Access-Control-Max-Age header via the CORS_MAX_AGE environment variable. Defaults to 10 minutes, which was influenced by the Chrome team's decision to cap max age in their browser to
that duration.

### DIFF
--- a/api/http/cors.go
+++ b/api/http/cors.go
@@ -3,10 +3,13 @@ package http
 import (
 	"net/http"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/18F/e-QIP-prototype/api"
 )
+
+const defaultMaxAge = 600
 
 // CORSHandler is the handler for CORS.
 type CORSHandler struct {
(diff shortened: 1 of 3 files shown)


ryanhofdotgov:
Tested under macOS 10.14 Chrome, Firefox, Safari. Not in a position to test under IE11 or Edge in a VM, so will confirm after merge and deploy to cloud.gov.

Note: Safari does not show successful preflight requests in the network monitoring console. (And failed preflights in Safari reportedly get displayed in the Resources tab, not the Network tab).

ryanhofdotgov:
@macrael Could you clarify – the default is already set in the `cors.go` code, there is just a comment block that has a comment about it in `env.Configure`. I don't really like the comment there since it is duplicative, but that's the convention in the file.

macrael:
Sorry, I was just pointing out that I liked that you put the default in the code that needed it instead of putting it in env.Configure() like a lot of our other defaults.

PR #1632, opened 2019-04-25
[EN-3658] Update section mapping

suzubara:
## Description

- This PR contains the next set of work required in order to fully deprecate the legacy `navigation.js` config file(s). It includes migrating usage related to looping over/finding form sections from the legacy config to the newer form config (which are specific to form type). Also includes removing some dead code leftover from the render refactor, which made big strides in decoupling from the legacy `navigation.js`.
- `ErrorList.jsx` - Updated to find the form section associated with errors based on new section keys instead of looping over the old navigation array via `navigationWalker`
  - This required changing all of the values of `data-section, data-subsection` attributes to use the new unique section keys instead of legacy names. This is low risk because those data attributes were *only* being used by `ErrorList`. But that is why this PR touches every section component file :(
- `Print.jsx` - Deleted legacy `Print.jsx` (not being used anymore) and added tests to new `Print/index.jsx`
- `Section.jsx` - Removed code that was used to render sections using `SectionView`. Since all sections have been refactored to render using Routes, this code is not being used anymore.
- `SectionTitle.jsx` - Refactored to use current router location to render the current section breadcrumbs instead of `this.props.section` and `navigation.js`. This also involved adding a `breadcrumbs` array to the `reduceSubsections` function.
- Miscellaneous eslint fixes

## Checklist for Reviewer

- [ ] Review code changes
- [ ] Clicking through the form and loading/verifying test scenarios should be sufficient QA for these changes.


suzubara:
@joshmfranklin This PR includes displaying the current section in the header breadcrumbs, so you may want to review how it transitions as you scroll down.

PR #1636, opened 2019-04-29
Remove DSN phone number option from all telephone blocks

christopherhuii:
## Description
This PR removes all DSN phone number references from the front end code. Previously, applicants had the ability to toggle between a domestic, DSN, and international phone number. Now they only have options for domestic and international phone numbers.

## Checklist for Reviewer

- [ ] Review code changes


src/components/Form/Telephone/Telephone.jsx
```diff
@@ -122,131 +76,118 @@ export default class Telephone extends ValidationElement {
...
-      this.updateNumber('DSN')
-    })
-  }
-
-  updateDomesticFirst(values) {
-    const domestic = { ...this.state.domestic, first: values.value }
-    this.setState({ domestic }, () => {
+  updateDomesticFirst = (values) => {
+    this.setState({
+      first: values.value,
```
suzubara: @christopherhuii I think these update functions still need to set `...this.state.domestic` etc because when I tried testing this and entering a value, nothing actually appears in the input.

christopherhuii:
Sigh, sorry about that. Things should be good now

PR #1637, opened 2019-04-30
[EN-3668] App. state validation spike - Identification

suzubara:
## Description

This PR contains a proposed model for ongoing work to refactor the JS classes currently used by eApp to validate the form at all levels. Changes include:
- Add `validate.js` library to use as a standard format for writing declarative models for all objects that require validation
- Write validation model and tests for `name, phone, ssn, email, eyeColor, hairColor, height, weight, sex` based on existing validation rules
- Created `models/validate.js` that can be used to set any global `validate.js` options, and exports a method (`validateModel`) to validate an object against a given model.
- Use new validation models for the following subsections:
  - Identification / Name
  - Identification / SSN
  - Identification / Contact Information
  - Identification / Physical
- Minor cleanup (ESLint fixes, created some `constants/enums` files)

Note: I omitted validation for the Date of Birth, Place of Birth, and Other names sections because they involve validating locations and dates, both of which will require additional overhead work that I would like to accomplish in a separate PR.

If accepted, the proposed strategy going forward is to:
- Continue to make changes on a section-by-section basis until the entire form has been covered.
- Implement `validate.js` models for low-level nodes _only._ For example, in this PR there are models for `phone` and `email`, but validating the _Identification / Contacts_ section as a whole is still left up to functions in the `validators/identificationcontacts.js` file. The reason for this is that while `validate.js` does have support for validating nested objects, it is limited, and using it at multiple levels would make it difficult to change the schema for models in the future. For now I think using functions in the existing `validators/` files is sufficient, and we can address this at a later point if needed.
- Update existing classes so that all logic is extracted into pure functions, and use `validateModel` for low-level nodes as mentioned above. Existing class methods that are in-use (such as `isValid()`) should be preserved so that they can still be called throughout the app, but the underlying implementation of all methods can be shifted to functions. Once this work is completed, the calling of validators itself can be shifted to use the new functions so that the legacy classes can be fully deprecated.
- Extract constants such as enums and regex patterns out to `constants/` files where possible, and shared logic/models out to util functions or models where appropriate.
- Add new tests and verify existing tests still pass when refactoring validation logic, aim for high test coverage. Since validation is purely functional this should be straightforward.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
     "remove-markdown": "^0.3.0",
     "reselect": "^4.0.0",
     "smoothscroll-polyfill": "^0.3.6",
-    "uswds": "^1.6.9"
+    "uswds": "^1.6.9",
+    "validate.js": "^0.12.0"
   },
   "husky": {
     "hooks": {
--- a/src/constants/enums/eyeColorOptions.js
+++ b/src/constants/enums/eyeColorOptions.js
@@ -1,0 +1,12 @@
+export default [
+  'Black',
+  'Blue',
+  'Brown',
+  'Gray',
+  'Green',
+  'Hazel',
+  'Maroon',
+  'Multicolored',
+  'Pink',
+  'Unknown',
+]
--- a/src/constants/enums/hairColorOptions.js
+++ b/src/constants/enums/hairColorOptions.js
@@ -1,0 +1,16 @@
+export default [
+  'Bald',
+  'Black',
+  'Blonde',
+  'Brown',
+  'Gray',
+  'Red',
+  'Sandy',
+  'White',
+  'Blue',
+  'Green',
+  'Orange',
+  'Pink',
+  'Purple',
+  'Unknown',
+]
--- a/src/constants/enums/nameSuffixOptions.js
+++ b/src/constants/enums/nameSuffixOptions.js
@@ -1,0 +1,15 @@
+export default [
+  '',
+  'Jr',
+  'Sr',
+  'II',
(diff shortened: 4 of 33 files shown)


suzubara:
I want to also note that in this plan, we can expect significant overhead early on that will lighten up as the work progresses. There are a number of models that are shared by most of the form sections, such as name, date, location, etc. which will require more time and attention, but then will be shared across the remaining sections. Therefore I expect this work to be slow going at first, but quickly pick up past the Relationships section.

suzubara:
@macrael I don't have a strong feeling about `model` vs `schema` and happy to change it if that's preferable!

macrael:
It’s a bit of bike shedding, but if model is a common name for this I would use that, if not I think I would prefer schema. I think of models as the actual object you are passing around, these are descriptions of those objects but aren’t the object themselves. 

-MacRae

suzubara:
@macrael Yeah that's valid. I was about to rename but realized that `src/schema` is already a thing full of these "schema" definitions for mapping the API data. in order to avoid confusion, I'd prefer keeping them totally separate for now.. how do you feel about renaming at a later date once perhaps the API data mapping has been cleaned up?

macrael:
Validate.js calls them “constraints” that’s not terrible. But I’m fine with either constraints or model, up to you.

suzubara:
ok I'm gonna leave it models for now, and should be an easy change if we decide to rename in the future.

PR #1638, opened 2019-04-30
[EN-3657] App validation saga

suzubara:
## Description

- This PR contains the *final* 🎉  set of work required in order to fully deprecate the legacy `navigation.js` config file(s). It includes replacing uses of `navigationWalker` with newer functions that determine the valid state of the form and its sections.
- `redux-saga` is added in order to provide a flow for updating the valid status of each form section that can be triggered just by dispatching the Redux action `validateFormData`.
- Replaced `validateApplication` function (which was only called on login) with the new `validateApplication` saga
  - This saga will loop through the form sections based on the current `formType`, and update the `isValid` attribute for each one in Redux. This can be improved upon further in the future, but for now it makes improvements without changing the signature of the Redux data.
- Remove old `navigation-helpers` code that is no longer being used
- Delete legacy `navigation.js` config files
- ESLint fixes
- **Note:** This PR introduces the use of `redux-saga` into the eApp code base. Sagas use a relatively new feature of ES6 called generators, which require a polyfill for a regenerator runtime. The codebase _already_ includes the `@babel/polyfill` (https://babeljs.io/docs/en/babel-polyfill) which satisfies these requirements, but it's worth noting the added dependency on this polyfill.

## Checklist for Reviewer

- [ ] Review code changes
- [ ] Verify that when you log into an account with valid and invalid data, the sidebar displays the correct valid/invalid status for each section.
- [ ] Verify that as you complete the form, each section's valid status displays correctly in the sidebar.


christopherhuii:
The sidebar (score card and navigation) doesn't update when the form initially loads.

1. Fill out a subsection until it validates with a green dot
2. Log out and log back in
3. Notice the subsection doesn't have a green dot

Alternatively:

1. Load up any test case (I used test case 6)
2. Notice that side bar doesn't update with the number of sections complete nor do any green dots show up.

christopherhuii:
1. Load up a fully validated application (e.g. test case 6)
2. Notice under `Review and submit`, applicants can click on `Review` and `Submit`.

In `develop`, the applicant has to click on `Review` first then proceed to `Submit`. I'm not 100% sure if that extra review/validation is necessary, but that is a difference from how `develop` flows.

suzubara:
1) Fixed the issue where the sidebar shows the correct valid status on log in, by moving the initial form data load into sagas (!)

2) As for the difference in the behavior with "Review & Submit", I'm kind of torn. Since the valid state is always accurately reflected in Redux now, if the form is 100% complete and valid, navigating to the "Review" page doesn't actually do anything. If the form is not complete/valid, then navigating to Review will still show the summary of sections with errors and prevent the user from going to the Submit page. If we want to maintain the existing behavior, we can either:
- add Redux flag that gets set to true when the user goes to the Review page for the first time, making the "Submit" sidebar link clickable
- keep the "Submit" sidebar link disabled always, so that the user always has to go through the Review page.
Curious to know what @joshmfranklin thinks?

suzubara:
I just looked into that discrepancy in the Review/Submit behavior and found that actually the Submit link is disabled on `develop` because the current error check is returning a false positive (looks related to the moving of the US Passports section). So I think the proper behavior on `develop` is in fact for the Submit link to be enabled if the form is complete and valid.

christopherhuii:
loaded test cases, clicked through and filled out the form. changes look good. minor differences as a result of the refactoring, but haven't noticed any showstoppers. i'll let josh get a chance before we merge.

PR #1640, opened 2019-05-02
[EN-3662] Switch To Simple Storage

macrael:
## Description

This is a large PR. It includes a new storage system I'm dubbing "Simple Storage" that saves the entire application as JSON rather than serializing each and every item into its own postgres column. 

The PR presently updates /save and /form to use the new serializer and updates the Rejector service to use them as well. There is still work to be done before this is all finished. 

## Checklist for Reviewer

- [ ] Review code changes


ryanhofdotgov:
We're running the database in our integration tests now... how much effort is involved to run the backend and do a full round trip with `load-scenario` and `form` on all our scenarios as part of the integration test suite?

macrael:
This PR is now in a reviewable state. The only thing I would like to do before merging is go through the old storage code and delete all of it.

This is a big PR! The main thing it does is replaces the old db store for applications and for attachments and creates a new storage system (simplestorage) for both of them. To be confident about these changes, I have added a lot of tests in the /api/integration package, testing via our APIs that /save and /form (and /status and the attachment apis) all work.

The integration tests have helpers for saving sections as well as saving entire test files. I've got it testing all the test sections we have, but haven't yet expanded those tests to cover all the complete-scenarios yet. It's worth talking about wether we'd rather have that testing actually using the load-scenario tool or wether testing it inside this package as is is enough. 

In addition to these apis, I've also created some admin handlers for kickback and for submission so that code for those APIs can be tested here. 

Design-wise, I've made Application a top level object, and made it so most of our attendant services (store, pdf, xml) all operate directly on that object instead of passing around db handles everywhere. This cleaned up the implementation of the yes/no clearing code in the kickback service considerably, which was a major driving change for this work.

The old DB code is not entirely gone, it's still being used to save and load accounts and transmissions. That can be fixed in a future PR. It was important to me that applications and attachments be implemented to make sure that the design of the store worked for those complicated cases.

ryanhofdotgov:
@macrael Just FYI, the last commit to remove the dead code adds a fair amount of noise to an already large PR, so I've created a throwaway PR to help review the code minus that last commit:
https://github.com/18F/e-QIP-prototype/pull/1663

ryanhofdotgov:
@macrael I've added a new issue EN-3780 to address some follow-on documentation updates that need to be done in the next sprint after we get this in place. Would you review that issue and add to the list as necessary?

macrael:
@ryanhofdotgov This PR is ready to review again. It has all the deletions in it so it might be easiest just to go through the recent commits one by one, they are fairly isolated to fixing a single issue each.

api/postgresql/db.go
```diff
@@ -168,7 +168,7 @@ func (service *Service) Save(query ...interface{}) error {
 		err := service.Insert(q)
 		if err != nil {
 			// oh dear god, only try and update if it fails because of a duplicated primary key.
-			if strings.HasSuffix(err.Error(), `_pkey"`) {
+			if strings.HasPrefix(err.Error(), "ERROR #23505 duplicate key value violates unique constraint") && strings.HasSuffix(err.Error(), `_pkey"`) {
```
ryanhofdotgov: It's hair splitting, but I was thinking of something more simple, like `strings.Contains(err.Error(), "duplicate key value violates unique constraint")`

Maybe we'll have time this year to revisit and use `ON CONFLICT DO UPDATE` which became available in PostgreSQL 9.5.

macrael: I'm also hoping that after this push we'll have time to move the last two tables into simple storage instead of continuing to use the old db code.

ryanhofdotgov: @macrael What 2 tables are that?

macrael: Looking at my notes it's 3. accounts, transmissions, and basic auth memberships.

ryanhofdotgov: I'm not sure those are good candidates for the document-oriented simple storage – let's discuss if you pick that work up.

macrael: I wasn't clear, I want to continue using the tables the same way, but re-implement the db code to just use sqlx and live inside the storage package rather than keeping the old db code around.

PR #1642, opened 2019-05-06
[EN-3656] Storybook init

suzubara:
## Description

- Basic setup needed to get Storybook component library up and running
- Upgraded to Babel v7 and related packages/config (this includes fixing some broken imports)
- Added initial add-ons for Storybook: knobs, actions, a11y, jest
- Set CSS asset paths, hoisted imports to eqip.scss
- Added stories for `ScoreCard, Radio, RadioGroup`

## Checklist for Reviewer

- [ ] Verify eApp basic functionality in Chrome & IE
- [ ] Verify you can start storybook with `yarn storybook` (outside of Docker, so you will need to `yarn install` locally)
- [ ] Verify font & image assets show up in both eApp & Storybook
- [ ] Review code changes


ryanhofdotgov:
@suzubara There's a ton of dependency updates here – long overdue. What is your level of confidence with respect to stability? Do we need to focus on any manual cross-browser testing this sprint, especially IE11, beyond the regular level?

Any way we can easily tell if all these dependencies have IE11 compatibility in mind?

suzubara:
@ryanhofdotgov Good question. The bulk of the dependency updates here are related to upgrading Babel from v6 to v7. Because Babel is primarily a developer tool used in the webpack build, the greatest area of potential risk is in the act of running webpack to build the assets themselves (that's why this update bubbled up a few broken imports that I needed to fix in this PR). In other words, if the engineers are able to run webpack in their environment, that should verify the stability of the upgrade.

The one exception is in the `@babel/polyfill` package, which *does* import code executed at runtime in the browser. The only change to that in Babel 7 is that early stage ECMAScript proposals are no longer included. I don't believe we are using any of those features, but I do think it's worth doing a basic manual run through of the form in IE11 just to verify.

Ultimately I know it looks like a major update to dependencies; it is, however, one I've done before on a large codebase with no issues, and that is because the majority of the changes are to the developer build pipeline itself rather than the compiled code.

PR #1643, opened 2019-05-07
[EN-3669] Remove local comment state in Field and remove "Add a comment" from eApp

christopherhuii:
## Description

This bug occurred because of the mismatching of data being passed from `props` and usage of `state`. The `Field` component was using internal `state` to preserve the text inside the textarea rather than letting it be controlled by the `props` that it was being passed.

The internal `state` has been remove and it relies on the `commentsValue` that it is passed.

Edit: This PR took an interesting turn. Upon further review of the SF86 and validation matrix, a number of the optional "Add a comment" fields are extraneous and can be removed. The optional "Add a comment" field that I originally was trying to fix will be removed with this PR.

| Ticket | Title | Validation Matrix Page |
| ------ | ----- | ---------------------- |
| EN-2902 | Remove comment from "Select your sex" | 7 |
| EN-2903 | Remove "Add a comment" from school address | 32 |
| EN-2904 | Remove "Add a comment" from reason leaving | 55 |
| EN-2905 | Remove "Add a comment" for investigation completed | 194 |
| EN-2906 | Remove "Add a comment" from clearance granted | 195-196 |
| EN-2907 | Remove "add a comment" from residence verifier relationship radio | 27-28 |

## Checklist for Reviewer

- [ ] Review code changes


ryanhofdotgov:
@christopherhuii If this comment field wasn't being persisted before, can you make sure we get this field represented in one of the complete test scenarios? 

You'll have to load up one of the existing scenarios (even test10.json), add the comment, save, serialize back out to json, run the XML tests, confirm comment is now in the XML, update fixture, test with e-QIP, etc. – that test scenario documentation should walk you through the steps, but ping me if you get stuck. There are a lot of moving parts, so it can be a bit confusing the first couple of times through.

ryanhofdotgov:
Based on quick chat with @christopherhuii: if we are going to keep the "Add comment" function around for future use in a select places (it's currently excised from the entire form), let's add a test that actually exercises it.

suzubara:
@christopherhuii @ryanhofdotgov @joshmfranklin  Wait, so are we indiscriminately getting rid of *all* comments blocks? This help text will probably need to be addressed in that case:
image

ryanhofdotgov:
@suzubara Thanks – looks like there is some cleanup here to do. The comment removal should not have been done indiscriminately – the intent was that @christopherhuii carefully look at the validation matrix/XSDs where this additional comment fields were present, and either remove them if e-QIP did not support them or make sure that comments were persisting all the way through to the SF-XML if e-QIP did support.

My understanding is that all these types of comment blocks added in eApp have no support in e-QIP, so we have to remove them all. When they were originally added in eApp, they weren't looking at e-QIP compatibility at all. When e-QIP compatibility is no longer required, we can add them back in, once there is support for them downstream from us.

Let's go ahead and merge this in. @christopherhuii would you review the help text for each question, identify those that need updating and pull in @joshmfranklin to get assistance as required. I'll open up a second ticket for that.

ryanhofdotgov:
Follow on task is in EN-3774

PR #1645, opened 2019-05-07
Breakout syscalls based on build target so app can be built cross platform

macrael:
## Description

It has been impossible to build any of our commands that used web client.go on MacOS because the echo implementation there used two syscalls that only exist on linux. This change uses files that are conditionally built for specific platforms to set the correct syscall as a variable. 

## Checklist for Reviewer

- [ ] Review code changes


macrael:
That is definitely a concern. I've mostly just been running tests in macOS which has gone well so far for me.

PR #1621, opened 2019-04-04
Add AlternateAddress to employment item address

suzubara:
## Description

- Fixes #1050 
- Adds the alternate address flow to Employment items based on the requirements described in the test cases below.
  - The alternate address flow displays follow-up fields if a given address is marked as `APO/FPO/DPO` or `Outside the U.S.`
  - If the address is marked `APO/FPO/DPO`, a secondary address block is displayed under the label `Provide physical location data with street address, base, post, embassy, unit, and country location or home port/fleet headquarter.`
  - If the address is marked `Outside the U.S.`, the follow up yes/no question is displayed: `Do you or did you have an APO/FPO address while at this location?`
    - If the answer is `Yes`, a secondary address block is displayed under the label `Provide APO/FPO address.`
    - If the answer is `No`, no secondary address block is displayed.
- TODO:
  - [x] Verify label copy
  - [ ] B/E / XML updates to persist new address keys? - need a new task for this
  - [x] UX question

## Test Cases
- [ ] **User is employed by the military**
  - The question `Is/was your physical work address different than your employer's address?` should NOT appear on the form
  - If the employer address (duty station) is `APO/FPO/DPO` or `Outside the U.S.`, the alternate address flow should appear underneath the employer address fields
- [ ] **User is not employed by the military and physical work address is not different**
  - The question `Is/was your physical work address different than your employer's address?` should appear on the form (mark "No")
  - If the employer address is `APO/FPO/DPO` or `Outside the U.S.`, the alternate address flow should appear underneath the physical work address question
- [ ] **User is not employed by the military and physical work address is different**
  - The question `Is/was your physical work address different than your employer's address?` should appear on the form (mark "Yes")
  - If the employer address is `APO/FPO/DPO` or `Outside the U.S.`, the alternate address flow should NOT appear underneath the physical work address question
  - If the physical work address is `APO/FPO/DPO` or `Outside the U.S.`, the alternate address flow should appear underneath the physical work address fields

## Checklist for Reviewer

- [ ] Review code changes
- [ ] Review changes for Database effects
- [ ] Verify change works in IE browser

More details about this can be found in [docs/review.md](docs/review.md)

PR #1650, opened 2019-05-15
Adding XML support for alternate employment addresses

ryanhofdotgov:
Closes EN-3665 and builds off work in PR #1621.


ryanhofdotgov:
test10.json has been validated against XSDs and e-QIP.

PR #1657, opened 2019-05-17
[EN-3672] Validation Refactor - Identification

suzubara:
## Description

- This PR adds a lot of groundwork for the validation refactor, based on existing WIP refactoring the Identification & History sections, including:
  - New enums/constants and helpers for location validation
  - Functions & tests for date and date range validation
  - Defined a generic `location` validation model, which can be used to create specific location "layouts" (such as address and birthplace)
  - Created custom validators for general use throughout the app
- Uses new validation models for *all* Identification subsections

## Checklist for Reviewer

- [ ] Review code changes


src/models/validate.js
```diff
@@ -63,3 +89,8 @@ export const validateModel = (data, model, options) => {
 }
 
 export default validateModel
+
+/** require Yes or No */
+export const hasYesOrNo = {
```
christopherhuii: Is there a reason why `hasYesOrNo` is in `validate.js` rather than broken out into a different file like the rest of the validators?

suzubara: @christopherhuii everything that's in the `src/models/validators` is a custom validator function that gets initialized in this `validate.js` file. `hasYesOrNo` is just a shorthand for this `inclusion` rule, since it is pretty common. I wasn't sold on how to organize something like that so I just stuck it here for now 🤷‍♀

PR #1661, opened 2019-05-21
Fix missing checkmarks for sidenav and input fields

joshmfranklin:
This fixes the missing checkbox icon from the sidenav and input fields. We had added `$asset-path` to the `eqip.scss` file to get the stylesheets working for Storybook, however this change added an extra `/` to the asset path in the app. This caused the icon url to be invalid.

This change does cause the styles to not load on Storybook.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,9 @@
     "globals": {
       "EAPP_VERSION": "MAJOR.MINOR.PATCH-IDENTIFIER"
     },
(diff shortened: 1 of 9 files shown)


suzubara:
Instead of removing the `/` from `$asset-path`, I removed it from the CSS that references `$asset-path`. I also added some stories for the `SectionLink` component in order to verify the icon shows up in both the app & storybook.

PR #1664, opened 2019-05-22
[EN-3774] Remove "Add a comment" locale from various parts of the application

christopherhuii:
## Description
In #1643 we removed the optional `Add a comment` field from various parts of the application. Because of that change, we need to update and remove some of the help text around eApp.

The only visible place in eApp I could find anything related to adding a comment was in `Your identifying information` subsection regarding sex. There are a lot of unused locale so I tried doing my best to delete unused ones about comment.

## Checklist for Reviewer

- [ ] Review code changes


ryanhofdotgov:
@christopherhuii Sex is not the same as gender, so since we don't have a way to persist a comment about that at the question level, please remove that locale entry/help text above the question entirely – thanks. Once we don't have to maintain 100% backwards compatibility with e-QIP, we can add it back in.

PR #1668, opened 2019-05-23
[EN-3704] Remove APO type for address of employment (Self Employment)

christopherhuii:
## Description
This PR removes the radio button to select APO address for `Provide the address of employment` when an applicant selects `Self-employment`. This change reflects what is valid in the validation matrix and e-QIP.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/components/Form/Location/Location.jsx
+++ b/src/components/Form/Location/Location.jsx
@@ -629,6 +629,7 @@ export default class Location extends ValidationElement {
             isEnabled={this.props.isEnabled}
             isPoBoxAllowed={this.props.isPoBoxAllowed}
             disableToggle={this.props.layout === Location.US_ADDRESS}
+            showPostOffice={this.props.showPostOffice}
             onBlur={this.handleBlur}
             onUpdate={this.updateAddress}
             onError={this.handleError}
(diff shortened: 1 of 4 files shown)


src/components/Form/Location/Location.jsx
```diff
@@ -868,6 +869,7 @@ Location.defaultProps = {
   addressBooks: {},
   addressBook: '',
   isPoBoxAllowed: true,
+  showPostOffice: true,
```
ryanhofdotgov: Does this line addition enable APO addresses for all locations by default?

christopherhuii: Yes it does, but that's a mistake. It should be `false` as default.

It should be `false` because in `src/components/Form/Location/Address.jsx` line 585, it's `false` by default.

src/components/Section/History/Employment/EmploymentItem.test.jsx
```diff
@@ -222,4 +222,14 @@ describe('The employment component', () => {
...
     expect(EmploymentItem.defaultProps.onError(null, [])).toEqual([])
   })
+
+  it('hides APO for address of employment if self employed', () => {
+    const props = {
+      EmploymentActivity: { value: 'SelfEmployment' },
+    }
+
+    const component = createComponent(props)
+    const employmentAddressParent = component.find("[data-fieldName='employmentAddress']")
```
suzubara: 👍 to adding a `data` attribute for test selectors! Can we call it something more specific like `data-test-id` to indicate that's what it's for?

christopherhuii:
@suzubara I think this is ready for review again. Thanks

PR #1669, opened 2019-05-23
[EN-3680] Validation Refactor - Relationships / People

suzubara:
## Description

- Validation model and tests for People, Person
- Includes function for determining whether there are gaps in date ranges in a given duration
- Adds requireNumber option to phone model in order to require a number value

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/helpers/date.js
+++ b/src/helpers/date.js
@@ -57,3 +57,57 @@ export const dateWithinRange = (date, range) => {
 
   return boundary <= createDateFromObject(date)
 }
+
+/** Convert date objects to luxon objects and sort by from date */
+export const sortDateRanges = (ranges) => {
+  const sortFn = (a, b) => {
+    if (a.from < b.from) return -1
+    if (a.from > b.from) return 1
+    return 0
+  }
+
+  return [...ranges].map(r => ({
+    from: createDateFromObject(cleanDateObject(r.from)),
+    to: r.present ? today : createDateFromObject(cleanDateObject(r.to)),
+  }))
+    .sort(sortFn)
+}
+
+/**
+ * All dates being operated on are luxon date objects
+ * @param {*} coverage - duration object that needs to be covered
+ * @param {*} ranges - ranges to check against coverage duration
+ */
+export const findTimelineGaps = (coverage, ranges) => {
+  const coverageDuration = Duration.fromObject(coverage)
+  const requiredRange = {
+    from: today.minus(coverageDuration),
+    to: today,
+  }
+
+  // prepare & sort ranges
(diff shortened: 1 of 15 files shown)


src/helpers/date.js
```diff
@@ -57,3 +57,57 @@ export const dateWithinRange = (date, range) => {
...
+
+  return [...ranges].map(r => ({
+    from: createDateFromObject(cleanDateObject(r.from)),
+    to: r.present ? today : createDateFromObject(cleanDateObject(r.to)),
+  }))
+    .sort(sortFn)
+}
+
+/**
+ * All dates being operated on are luxon date objects
```
ryanhofdotgov: This is fantastic!

src/models/validators/durationCoverage.js
```diff
@@ -0,0 +1,26 @@
+import { validate } from 'validate.js'
+import { validateModel } from 'models/validate'
+import { findTimelineGaps } from 'helpers/date'
+
+const durationCoverageValidator = (value, options = {}) => {
+  if (validate.isEmpty(value)) return null // Don't validate if there is no value
+
+  const { requiredDuration } = options
+  if (!requiredDuration) return 'Invalid options'
```
ryanhofdotgov: These string return values seems a little odd – what the intended usage by the caller?

src/validators/people.js
```diff
@@ -1,107 +1,37 @@
...
 
   validCount() {
-    return ((this.list || {}).items || []).reduce((acc, cur) => {
-      const valid = new PersonValidator(cur.Item).isValid()
-      return valid ? acc + 1 : acc
-    }, 0)
-  }
-
-  validYearRange() {
-    const julianMax = julian(daysAgo(today, 365 * minimumYears))
```
ryanhofdotgov: I'm very happy to see this code replaced :)

src/models/validators/durationCoverage.js
ryanhofdotgov: (I didn't see a caller of `durationCoverageValidator` so didn't want it to hold up the PR.)

suzubara: @ryanhofdotgov So the paradigm for `validate.js` custom validators is to return `null` if the value passes validation, and an error message if it doesn't. We aren't actually using any of these error messages currently, since right now `validate.js` is only being used to update the valid state in Redux, and also because we use the error message locale files. I'm thinking that when we get to the UI refactor, we can look at whether or not to change this, and how to use these messages better (for example, maybe returning a key from the locale file instead). For now I've just been returning a string that describes what the error itself is.

PR #1670, opened 2019-05-23
[EN-3782] Update broken complete test scenarios

christopherhuii:
## Description
This PR fixes the following test cases. They all validate now.

1. test03
2. test04
3. test05
4. test07 
5. test09

- Fixed employment model/validator for `Additional` employments. It was executing for all employment activity types, but it should only be executed for "Other" employment activity types.
- Removed middle name from education verifier to align with validation matrix, eApp backend/XML, and e-QIP
- Updated test case 7 by removing employment reference middle name and adding a required contact email.
- Added missing name suffix option.
- Fix relevant tests.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/constants/enums/nameSuffixOptions.js
+++ b/src/constants/enums/nameSuffixOptions.js
@@ -2,6 +2,7 @@ export default [
   '',
   'Jr',
   'Sr',
+  'I',
   'II',
   'III',
   'IV',
--- a/src/models/validators/branchCollection.js
+++ b/src/models/validators/branchCollection.js
@@ -7,7 +7,7 @@ const branchCollectionValidator = (value, options = {}) => {
   const { items } = value
   const { validator } = options
 
-  if (!items || !items.length) return 'Collection is empty'
+  if (!items || !items.length) return null
   if (!validator) return 'Invalid validator'
 
   const hasNo = items.some(i => i.Item && i.Item.Has && i.Item.Has.value === 'No')
--- a/src/models/validators/branchCollection.js
+++ b/src/models/validators/branchCollection.js
@@ -7,7 +7,7 @@ const branchCollectionValidator = (value, options = {}) => {
   const { items } = value
   const { validator } = options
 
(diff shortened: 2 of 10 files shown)


src/models/validators/branchCollection.js
```diff
@@ -7,7 +7,7 @@ const branchCollectionValidator = (value, options = {}) => {
   const { items } = value
   const { validator } = options
 
-  if (!items || !items.length) return 'Collection is empty'
+  if (!items || !items.length) return null
```
ryanhofdotgov: @suzubara This looks like similar pattern to something I commented on in https://github.com/18F/e-QIP-prototype/pull/1669 – what's the intent around these error strings as return values?

suzubara: @ryanhofdotgov see my comment: https://github.com/18F/e-QIP-prototype/pull/1669#discussion_r287362832

@christopherhuii what test case is this fixing? is a branch collection supposed to be valid if it's empty?

src/constants/enums/nameSuffixOptions.js
```diff
@@ -2,6 +2,7 @@ export default [
   '',
   'Jr',
   'Sr',
+  'I',
```
ryanhofdotgov: Nice catch!

src/models/validators/branchCollection.js
christopherhuii: Gotcha, this change has been reverted

src/validators/employment.test.js
```diff
@@ -701,7 +701,7 @@ describe('Employment component validation', () => {
       {
         data: {
           EmploymentActivity: {
-            value: 'SelfEmployed',
+            value: 'StateGovernment',
```
ryanhofdotgov: @christopherhuii Is there an opportunity to have a test that verifies that the additional question isn't answered/set e.g., for self-employment?

PR #1673, opened 2019-05-24
[EN-3694] Submission errors on specific police offense scenario

ryanhofdotgov:


ryanhofdotgov:
EN-3786 is resolved with #1674, which is about to be merged. `test10.json` was updated to reflect changes in #1674 and is now validated by e-QIP.

ryanhofdotgov:
The "one space" <Street> elements ended up being a red herring to resolve EN-3694, but fixed that here anyways.

PR #1675, opened 2019-05-28
[EN-3650] Update Graffle, template PDFs, known-good files to July 2017 version

ryanhofdotgov:
Incorporates the updated Graffle and template PDFs that were posted to EN-3650.


ryanhofdotgov:
Let's do it as a separate PR, because I want to tweak the 85 one to match an identifier in the corresponding XML schema, and put a placeholder for the 85P.

PR #1676, opened 2019-05-28
[EN-3692] Audit use of DoNotKnow in XML

ryanhofdotgov:
Makes XML generation code more consistent, defensive when DoNotKnow is an option. Helps backend code be resilient to frontend situations where toggling a branch doesn't always clear state of closed off branch.

### DIFF
--- a/api/templates/email-optional.xml
+++ b/api/templates/email-optional.xml
@@ -1,0 +1,6 @@
+{{$dnk := notApplicable .DoNotKnow}}
+{{if $dnk | ne "True"}}
+<Email>{{email .Email}}</Email>
+{{else}}
+<Email DoNotKnow="{{$dnk}}"></Email>
+{{end}}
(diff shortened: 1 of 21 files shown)


api/templates/email-optional.xml
```diff
@@ -0,0 +1,5 @@
+{{if notApplicable .DoNotKnow | eq "True"}}
+<Email DoNotKnow="True"></Email>
+{{else}}
+<Email>{{email .Email}}</Email>
```
macrael: The other templates you changed (like date-month-year-optional) seem to re-implement the non-optional template instead of just calling it like you did here. Is there any reason to not standardize on one behavior or the other?

ryanhofdotgov: I had a couple of thoughts in mind:

* I changed the existing *-optional templates to use the same logic pattern used elsewhere. It prevents weird empty space/blank line elements in the XML output that downstream devs had commented on in the past and makes code pattern simpler and instantly recognizable – but at the cost of a repeated element tag here and there.

*  Whether or not there were multiple call sites in the template files drove whether or not to create a separate template file, Go function, addition to function map. Many of them were one offs and it wouldn't actually save lines of code.

Even ones that look there are multiple call sites (like `LegalName`) actually are different because one might need nested first/middle/last and another might need just first/last.

PR #1682, opened 2019-05-30
[EN-3700] Change form versions for consistency with e-QIP

ryanhofdotgov:
July 2017 is the most recent version of SF-86, which is identical to the November 2016 version. It is the revision e-QIP references.

The XML XSDs we are using for the 86 and 85 use '2017-07' and '2017-12-draft7' respectively, so this change creates consistency there as well.

Testing shows that the stored, serialized JSON meta-data for form version is automatically updated once account.form_version is updated.

### DIFF
--- a/api/account.go
+++ b/api/account.go
@@ -21,11 +21,11 @@ var (
 var knownFormVersions = map[string][]string{
 
 	"SF86": []string{
-		"2016-11",
+		"2017-07",
 	},
 
 	"SF85": []string{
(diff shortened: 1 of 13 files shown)


ryanhofdotgov:
@macrael Can you take a look at the CircleCI failures on this PR – I think it might be related to the other issue you're looking at.

ryanhofdotgov:
@macrael false alarm – it's not related at all. :)

PR #1683, opened 2019-05-30
[EN-3781] Mobile layout tweaks

joshmfranklin:
Fixes EN-3781

Updates to styling for improved mobile use.


src/components/Form/DateRange/DateRange.scss
```diff
@@ -13,6 +13,13 @@
     width: 5rem;
     text-align: center;
     padding-top: 7rem;
+
+    &:before {
+      content: '\f30b';
+      font-family: 'Font Awesome 5 free';
```
christopherhuii: Is there a reason why the `font-family` needs to be `Font Awesome 5 free` and not just `font-family: FontAwesome`?

I'm not familiar with `'Font Awesome 5 free'

ryanhofdotgov:
@joshmfranklin Would you add a comment about what browsers you've tested these CSS changes under? 

I saw some test updates and some unexpected code changes in Telephone.jsx. Could you clarify if there are non-layout bug fixes in here as well? Thanks!

src/components/Form/DateRange/DateRange.scss
joshmfranklin: Font Awesome 5 Free is the updated font name that Font Awesome uses to declare which icon set to use. There is now a "free" version and a paid version so they separated the fonts into two categories. We are using the free version.

joshmfranklin:
> @joshmfranklin Would you add a comment about what browsers you've tested these CSS changes under?
> 
> I saw some test updates and some unexpected code changes in Telephone.jsx. Could you clarify if there are non-layout bug fixes in here as well? Thanks!

@ryanhofdotgov the changes have been tested in Chrome responsive dev tools so far, I am pairing with MacRae to get a device setup for testing changes locally. We can wait to merge this PR until I've tested further on a device.

The updates to the `.jsx` files were mostly lint required changes along with a couple class updates and the addition of some wrapping divs in the `Telephone.jsx` in order to accommodate for better mobile layouts. The changes to the `SummaryCounter.test.jsx` file were made because of updates to class names that the test was looking for.

ryanhofdotgov:
Thanks! If you could hit IE11 as well, that will round off what we minimally need.

PR #1684, opened 2019-05-30
[EN-3795] Fixes the filling out of mandatory counseling and navigation to voluntary subsection

christopherhuii:
## Description
This PR fixes two issues that come from a single root cause.

- `Have any of the following ordered, advised, or asked you to seek counseling or treatment as a result of your illegal use of drugs or controlled substances?` does not save.
- Cannot navigate to the next section

It's not exactly clear how or when this bug was introduced, but the issue in the code has to do with the variable type. The function was expecting an object, but it was being passed an array. This issue is similar to the `values`, but that seems to occasionally reoccur. I've updated that component function and a number of tests to make sure this doesn't happen again.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/components/Section/SubstanceUse/Drugs/OrderedTreatment.jsx
+++ b/src/components/Section/SubstanceUse/Drugs/OrderedTreatment.jsx
@@ -11,7 +11,7 @@ import {
   Textarea,
   DateRange,
   Telephone,
-  Show
+  Show,
 } from '../../../Form'
 import DrugType from './DrugType'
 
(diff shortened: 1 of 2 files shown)


ryanhofdotgov:
@christopherhuii Can you clarify what you mean by "This issue is similar to the values, but that seems to occasionally reoccur"?

joshmfranklin:
@christopherhuii not sure if this has something to do with my setup or the changes made, but when i clear my db and then reload the app but the first question in `Mandatory counseling or treatment` is always answered `no` by default.

joshmfranklin:
@christopherhuii I'm not sure if this is a problem with my setup or something to do with the changes for this issue, but after I clear my db locally the first question in `Mandatory counseling or treatment` is always answered `no` and there are questions answered in `Voluntary counseling or treatment` as well.

christopherhuii:
@ryanhofdotgov A similar example is countries. The value (as in key value pair) can be a string or an object.

Example:
```
// string
country: 'United Status'

// object
country: {
  value: 'United States',
}
```

This had been a problem in the past.

In this situation, `this.props.OrderedBy` is expected to be an object with the `values` key. For some reason, it was just the value.

@joshmfranklin I'm not seeing that issue on my end. Maybe I can take a look at what you're seeing later.

PR #1685, opened 2019-05-31
[EN-3799] Fix countryString if an empty object

suzubara:
## Description

Fixes an issue with the `countryString` normalizer function. All uses of this function expect it to return a value that is render-able by React (ie a string, boolean, or null/undefined -- but not an array or object). However, with the existing function if the parameter is an empty object or an object that is missing a `value` key it will just return the object, causing an error at render time. This PR fixes this use case and also includes linter fixes.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/validators/location.js
+++ b/src/validators/location.js
@@ -106,16 +106,18 @@ export class Geocoder {
 
 export default class LocationValidator {
   constructor(data = {}) {
-    this.layout = data.layout
+    if (data) {
+      this.layout = data.layout
 
-    // Data
-    this.street = data.street
-    this.street2 = data.street2
-    this.city = data.city
-    this.state = data.state
-    this.zipcode = data.zipcode
-    this.county = data.county
-    this.country = countryString(data.country)
+      // Data
+      this.street = data.street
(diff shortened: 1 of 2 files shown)


src/validators/location.js
```diff
@@ -1,43 +1,123 @@
...
     if (isDefined(country.value)) {
       if (Array.isArray(country.value)) {
         return country.value[0]
       }
 
       return country.value
-    } else if (country.value === null) {
+    }
+
+    if (country.value === null) {
```
christopherhuii: Can we move the conditional check for `country.value === null` and `country.value === undefined` nested at the top of `if (validate.isHash(country)) {` and delete `if (isDefined(country.value)) {`.

Maybe something like this

```
export const countryString = (country) => {
  if (validate.isHash(country)) {
    if (country.value === null || country.value === undefined) {
      return country.value
    }

    if (Array.isArray(country.value)) {
      return country.value[0]
    }

    return country.value
  }
  return country
}
```

suzubara: 👍 good call

PR #1688, opened 2019-05-31
[EN-3690] Update eApp front-end submission

christopherhuii:
## Description
This story was originally intended to help the applicant download the archival PDF form in a way that flows intuitively.

There were a number of changes suggested
- [x] Change the submission confirmation browser alert to a modal so it grabs the attention of the applicant
- [x] Add the "loading" spinner to indicate that the application is being submitted
- [ ] Redesign the archival PDF download/print page so it's easier to use and understand

The first two changes have been completed and are included in this pull request. I would like to consider deferring re-stying/designing the download/print page until the backend portion is completed. This way, we won't be blocked by other teams in addition to the time it'll get us too get that endpoint setup..

## Checklist for Reviewer

- [ ] Review code changes
- [x] Verify change works in IE browser


ryanhofdotgov:
There's a checkbox on in the PR template, but just a friendly reminder to test these changes under IE11. I know the PR template says it's a reviewer responsibility, but ideally the dev submitting the PR should take a pass at the changes under IE11 as well as Chrome. Thanks!

christopherhuii:
PR has been checked on IE11. I have noticed that submitting does take a little bit longer. This might be due to the VM rather than IE. So if things look like they're hanging up, just wait a little bit and they should resolve.

joshmfranklin:
I checked this branch out and what I am seeing is that the first two steps of the updated submission process are working (updated confirmation modal and spinner), the redirect to the downloads page is not working yet so that will need to be completed in order for this to be merge-able.

christopherhuii:
@joshmfranklin Not quite sure I understand the issue. After the spinner, the page navigates to the "old" print and PDF download page. The new download page will be implemented in a separate task. There's a "part 2" issue.

Does it redirect to the "old" download page? or does it hang?

joshmfranklin:
@christopherhuii this was a hiccup with my local branch. This part of the PDF workflow looks good to me.

PR #1692, opened 2019-06-03
Fix race condition in /save call

macrael:
## Description

The new save implementation introduced a race condition where data could be lost if too many save calls were happening when there wasn't yet an application row for the given user. Fix is just to try and update again if you fail both to update and then create. There are definitely faster solutions to this problem but since it is only possible for it to happen to a user when they are filling out their first section and then never again it doesn't seem worth the complication. 

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/api/http/save.go
+++ b/api/http/save.go
@@ -70,7 +70,8 @@ func (service SaveHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Save to storage and report any errors
-	if saveErr := service.Store.SaveSection(section, id); saveErr != nil {
+	saveErr := service.Store.SaveSection(section, id)
+	if saveErr != nil {
 		if saveErr == api.ErrApplicationDoesNotExist {
 			// if the application doesn't exist, we need to create it.
 			newApplication := api.BlankApplication(account.ID, account.FormType, account.FormVersion)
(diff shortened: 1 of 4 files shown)


api/http/save.go
```diff
@@ -70,17 +70,32 @@ func (service SaveHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 			if createErr != nil {
-				service.Log.WarnError(api.EntitySaveError, createErr, api.LogFields{})
-				RespondWithStructuredError(w, api.EntitySaveError, http.StatusInternalServerError)
-				return
+				// This should happen but rarely, but there is a race condition here where multiple /save calls
+				// get ApplicationDoesNotExist above. In that case, some of them will get ApplicationExists here,
+				// but it's safe for them to just try again.
+				if createErr == api.ErrApplicationAlreadyExists {
+					service.Log.Info("Having to double save due to /save race", api.LogFields{})
+					saveAgainErr := service.Store.SaveSection(section, id)
```
ryanhofdotgov: This going to happen on every new login – it's going to be pretty common, so I would log it at debug level, not info

api/http/save.go
```diff
@@ -70,17 +70,32 @@ func (service SaveHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 			createErr := service.Store.CreateApplication(newApplication)
 			if createErr != nil {
-				service.Log.WarnError(api.EntitySaveError, createErr, api.LogFields{})
-				RespondWithStructuredError(w, api.EntitySaveError, http.StatusInternalServerError)
-				return
+				// This should happen but rarely, but there is a race condition here where multiple /save calls
+				// get ApplicationDoesNotExist above. In that case, some of them will get ApplicationExists here,
+				// but it's safe for them to just try again.
+				if createErr == api.ErrApplicationAlreadyExists {
+					service.Log.Info("Having to double save due to /save race", api.LogFields{})
```
ryanhofdotgov:
```suggestion
					service.Log.Debug("Having to double save due to /save race", api.LogFields{})
```

PR #1696, opened 2019-06-04
Clear No Branches on Kickback

macrael:
## Description

This PR clears all the questions answered "no" that need to be cleared when the application is kicked backed (aka rejected) to the applicant. 

The list of Nos that have to be cleared are in a document Josh compiled. 

## Checklist for Reviewer

- [ ] Review code changes


macrael:
There's some extra commits in the history here but the "files changed" tab is a clean diff from develop.

macrael:
I'm up to 108/189 of the no's cleared. It's starting to move pretty fast. I think I found a legit xml that the tests are now going to be failing for.

macrael:
This PR is massive and ready to go!

api/foreign.go
```diff
@@ -280,6 +303,55 @@ func (entity *ForeignTravel) Valid() (bool, error) {
...
+// ClearNos clears the "no" answers on application rejection
+func (entity *ForeignTravel) ClearNos() error {
+	entity.HasForeignTravelOutside.ClearNo()
+
+	// One of the few "Yes" answers that must be cleared
+	if entity.HasForeignTravelOfficial != nil && entity.HasForeignTravelOfficial.Value == "Yes" {
+		entity.HasForeignTravelOfficial.Value = ""
+	}
+
+	questionedErr := entity.List.ClearBranchItemsNo("Questioned")
```
sojeri: this method would be a great place to start with a list of keys and loop through clearing the branch items & returning any error that surfaces. :)

api/foreign.go
```diff
@@ -337,6 +409,41 @@ func (entity *ForeignActivitiesBenefits) Valid() (bool, error) {
...
+		benefitNames := []string{"ContinuingBenefit", "FutureBenefit", "OneTimeBenefit", "OtherBenefit"}
+
+		for _, benefitName := range benefitNames {
+
+			benefitItem, benefitErr := benefitSetItem.GetItemValue(benefitName)
+			if benefitErr != nil {
+				return errors.Wrap(benefitErr, fmt.Sprintf("error fetching benefit named %s from foreign activities", benefitName))
+			}
+
+			// Sometimes, OtherBenefit if it hasn't been selected seems to be sent as a textarea, but in that case we can't clear it
```
sojeri: this sounds like something we might want to look into so we at least know why it is happening. not sure where we're tracking work, but is this something it's worth adding a small chore or story for?

api/foreign.go
```diff
@@ -679,6 +856,32 @@ func (entity *ForeignBusinessConferences) Valid() (bool, error) {
 	return entity.List.Valid()
 }
 
+// ClearNos clears the "no" answers on application rejection
+func (entity *ForeignBusinessConferences) ClearNos() error {
```
sojeri: this whole method, bar the fist line, is nearly identical to its counterpart in ForeignActivitiesDirect.

I wonder if it might be possible to abstract this logic out & to have both call into the common logic with string value for the event Items collection & maybe a custom error string to output?

api/foreign.go
```diff
@@ -1021,6 +1278,13 @@ func (entity *ForeignBusinessVentures) Valid() (bool, error) {
 	return entity.List.Valid()
 }
 
+// ClearNos clears the "no" answers on application rejection
+func (entity *ForeignBusinessVentures) ClearNos() error {
+	entity.HasForeignVentures.ClearNo()
```
sojeri: probably out of the scope of this PR since we want it in today, but an idea I had for abstracting this is:
- if we could have a generic property name that references the same value as `HasForeignVentures` for `ForeignBusinessVentures` or `HasForeignVoting` for `ForeignBusinessVoting`, then these methods could potentially all be abstracted to a single source.

api/legal.go
```diff
@@ -457,6 +518,13 @@ func (entity *LegalTechnologyManipulating) Valid() (bool, error) {
 	return entity.List.Valid()
 }
 
+// ClearNos clears the "no" answers on application rejection
+func (entity *LegalTechnologyManipulating) ClearNos() error {
+	entity.HasManipulating.ClearNo()
```
sojeri: all of these Legal ones are following the same pattern, too. if we could have that generic type define `ClearNos`, then whatever constructor or initializer would only have to map the generic property type to all these type-specific ones like `HasManipulating`.

api/psychological.go
```diff
@@ -134,6 +165,36 @@ func (entity *PsychologicalConsultations) Valid() (bool, error) {
...
+
+	for _, incident := range entity.List.Items {
+		appealsItem, getErr := incident.GetItemValue("Appeals")
+		if getErr != nil {
+			return getErr
+		}
+
+		appeals, ok := appealsItem.(*Collection)
+		if !ok {
+			return errors.New("Unexpected type found trying to clear the competence appeals")
```
sojeri: `competence appeals` should probably be `consultations` here

api/relationships.go
```diff
@@ -145,6 +182,17 @@ func (entity *RelationshipsCohabitants) Valid() (bool, error) {
 	return entity.CohabitantList.Valid()
 }
 
+// ClearNos clears any questions answered nos on a kickback
+func (entity *RelationshipsCohabitants) ClearNos() error {
+
+	entity.HasCohabitant.ClearNo()
+
+	entity.CohabitantList.ClearBranchNo()
```
sojeri: way to ruin my abstraction proposal, cohabitation ;_;

though theoretically this one could override from the base

macrael:
The New XML files submit to eQIP successfully.

api/foreign.go
macrael: I changed ClearBranchItemsNo to take multiple keys to clear

api/foreign.go
macrael: I created [EN-3822]

api/foreign.go
macrael: You can see how I did something like that for the TestClearBasicSectionNos test. I could see us trying something similar to implement all the basics. I think that if we did that we'd probably end up recording them all in the reject.go file instead of split out. As is I kind of like that we succinctly describe what needs to be cleared right next to the actual struct.

macrael:
Should be ready to go now @sojeri !

api/foreign.go
sojeri: just read that commit. I love it! :)

PR #1699, opened 2019-06-05
[EN-2670] Fix Education Reference validation

suzubara:
## Description

When an Education item is within the last 3 years:
- Display all Reference fields (name, phone, email, address)
- Only Name and Email have "I Don't Know" (not applicable) options
- Make sure validation is consistent with this requirement

## Checklist for Reviewer

- [ ] Review code changes


ryanhofdotgov:
@suzubara I think before we merge this in, we should have some education verifiers in one of the complete scenario test cases that exercises this corrected approach (no name, no email), and make sure the existing test cases load up/validate after this change.

There's also a work-around in `history-education.xml` (look for `XXX`) that can be eliminated I believe. @christopherhuii has recently done this type of work (adding/modifying new cases), making simple updates to template XML, re-submitting to e-QIP, so perhaps let him take it over the finish line or create a separate issue as a dependency of this one?

suzubara:
@ryanhofdotgov @christopherhuii Ok, sounds good. I will note that most of the changes in this PR are code cleanup, and the only effective change is to display the rest of the Reference fields regardless of whether or not there's a value in `ReferenceName` (previously, they only showed up if `ReferenceName` had a value). So I think the test case we'll want to add is one where `ReferenceName` is not applicable, but the other fields (telephone/address) have values.

christopherhuii:
@ryanhofdotgov @suzubara I was in the process of updating a test case for this scenario, but I ran into some issues. I'll outline the steps I took.

1. Loaded up test case 3
2. Added a school within the past 3 years (01/01/2018 - present), with a verifier with no name nor email
3. Exported the user's (test03) JSON using [this](https://github.com/18F/e-QIP-prototype/blob/develop/docs/test-scenarios.md#generating-new-test-json-files)
4. Replaced the contents `test3.json` of with `output.json` so that it has the new education item added to the test case


When running `make test-go`, the error is `panic: interface conversion: interface {} is nil, not string [recovered]`. My suspicion is that the errors are from the 18 occurrences of `"props": {}`. There are no instances of `"props": {}` in any of the test cases on `develop`.

ryanhofdotgov:
Test case work has been split into a separate PR: https://github.com/18F/e-QIP-prototype/pull/1780

PR #1700, opened 2019-06-10
[EN-1630] Fix test10

suzubara:
## Description

- To reproduce this issue, load test case 10 and 2 of the employment activities do not pass validation.
- `test10.json` has been fixed so the test data passes validation. There were no effective changes to the test10 XML.
- Add logic to `Location.jsx update` so that if the country is being changed and is not the US or POSTOFFICE, clear out any existing state or zipcode values.
- Ensure state code values are always uppercase
- ESLint fixes on `Location, State, MilitaryState` components

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/api/testdata/complete-scenarios/test10.json
+++ b/api/testdata/complete-scenarios/test10.json
@@ -936,7 +936,6 @@
                       "layout": "Address",
                       "street": "Self Employment APO",
                       "city": "Winnipeg",
-                      "zipcode": "34038",
                       "country": "Canada"
                     }
                   },
@@ -3151,7 +3150,7 @@
                       "layout": "Address",
                       "street": "APO Address",
                       "city": "APO",
-                      "state": "aa",
+                      "state": "AA",
                       "zipcode": "34050",
                       "country": "POSTOFFICE",
                       "validated": true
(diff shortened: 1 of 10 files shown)


suzubara:
@ryanhofdotgov I think it's worth noting that these changes fix the specific issues that caused test case 10 to not pass validation. The larger issue of "ghost" error messages in accordions is due to the code that displays inline errors being suboptimal, and I think we can address that as part of the UI validation refactor work.

ryanhofdotgov:
@suzubara I think the problem might be in the test case itself or something in the frontend that allowed the data in the test case to be created:

1. International addresses shouldn't have zip codes. I would guess that someone started filling out the address as domestic, entered a zip code, then flipped it to an international address. Then  the zip code still got persisted into db, reloaded, etc.

1. State abbreviations need to be in all caps. In theory, some work was done in past 6 or so months to prevent users from being able to enter in lower case abbreviations. I think @christopherhuii did that work.

So I think the frontend validation might be OK? But something else is broken around the entry of these 2 data fields. I would recommend fixing test10 so it conforms to what the frontend produces now and poke around, review the code around zip code/state abbreviation input. What are your thoughts?

suzubara:
@ryanhofdotgov Ok, I agree on both items. I'll update the test case and look into the location & state inputs to make sure those values can't reappear.

suzubara:
This is ready for another review

PR #1702, opened 2019-06-12
[EN-3717] Validation Refactor - Foreign / Direct Activities

suzubara:
## Description

- Models, tests, validators for Foreign CoOwners, Foreign Direct Interests, Foreign Direct Activity

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/models/foreignDirectInterest.js
+++ b/src/models/foreignDirectInterest.js
@@ -1,0 +1,40 @@
+import { foreignCoOwnersModel } from 'validators/foreigncoowner'
+
+const foreignDirectInterest = {
+  InterestTypes: {
+    presence: true,
+    array: {
+      validator: { presence: true },
+      length: { minimum: 1 },
+    },
+  },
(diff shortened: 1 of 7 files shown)


src/models/foreignDirectInterest.js
```diff
@@ -0,0 +1,40 @@
...
+    presence: true,
+    array: {
+      validator: { presence: true },
+      length: { minimum: 1 },
+    },
+  },
+  InterestType: { presence: true, hasValue: true },
+  Acquired: { presence: true, date: { requireDay: false } },
+  HowAcquired: { presence: true, hasValue: true },
+  Cost: { presence: true, hasValue: true },
```
christopherhuii: i'm not too concerned about this because we're restricting non-digits/numbers from being type, but would it be better to use `{ presence: true, hasValue: { validator: { numericality: true } } }` for `Cost` as well as `Value`?

PR #1718, opened 2019-06-14
[EN-3790] Validation Refactor - Military / Disciplinary Procedures

christopherhuii:
## Description
- Updates validations. Pretty straight forward subsection

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/models/militaryDiscipline.js
+++ b/src/models/militaryDiscipline.js
@@ -1,5 +1,8 @@
 const militaryDiscipline = {
-  Date: { presence: true, date: true },
+  Date: {
+    presence: true,
+    date: { requireDay: false },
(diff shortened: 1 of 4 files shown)


src/models/militaryDiscipline.js
```diff
@@ -0,0 +1,9 @@
+const militaryDiscipline = {
+  Date: { presence: true, date: true },
```
suzubara: Based on the UI I think this _should_ use `date: { requireDay: false }`. The validator is working without it because date control seems to pass a default value for day if the day field is hidden, but I think it'd be good to explicitly include this flag so it's not dependent on that.

PR #1743, opened 2019-06-25
[3663] set kickback status

macrael:
## Description

This PR changes the /status api to return a `status` enum instead of a `locked` bool. 

The enum's values are: 
`INCOMPLETE`, `SUBMITTED`, and `KICKBACK`

The `KICKBACK` status is used to display the kickback banner on the front end. 

I've also added a kickback binary that will kickback an application.

## Checklist for Reviewer

- [ ] Review code changes


src/schema/section/submission.js
```diff
@@ -22,7 +22,6 @@ export const submission = (data = {}) => {
     }),
     Attachments: form.general('submission.attachments', {
       Method: form.text(attachmentType)
-    }),
-    Locked: data.Locked
```
macrael: I couldn't find anywhere this was being used so I just removed it. @suzubara @christopherhuii is that OK?

PR #1751, opened 2019-07-02
[EN-3037] Fix error messages showing up too early in Relatives

christopherhuii:
## Description
This PR fixes the errors messages that appear too early. The errors messages would be saying that "Mother And Father are required" as well as "Mother-in-law and Father-in-law are required".

In this PR:
- The errors aren't displayed until AFTER `Do you have an additional relative to enter?` is answered with `No.
- There was uncaught bug where eApp asked for the mother and father in law in all cases. Mother and father in law are only required for applicants who are married or separated.

## Checklist for Reviewer

- [ ] Review code changes


joshmfranklin:
@christopherhuii 
- I think we can add back the second line of text for the mother-in-law and father-in-law error message `You indicated that you are married or separated, please provide entries for mother-in-law and father-in-law.` since that one doesn't reference the "I don't know" checkbox.
- We should also show that there is an error for the subsection on the review page. Right now the error messages show up if you scroll down to the relatives subsection but the errors don't show in the review summary at the top of the page.

src/components/Section/Relationships/Relatives/Relatives.jsx
```diff
@@ -37,6 +37,31 @@ export class Relatives extends Subsection {
...
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.List.branch.value !== this.props.List.branch.value) {
+      this.checkRelativesSubsectionErrors(this.props.List.branch.value)
+    }
+  }
+
+  checkRelativesSubsectionErrors = (value) => {
```
suzubara: What is the purpose of adding this? I would like to move away from this pattern of using `this.constructor.errors` on components, and we shouldn't need to explicitly "check" for errors if the form data changes since that is already done by the validators.

src/components/Section/Relationships/Relatives/Relatives.jsx
```diff
@@ -75,6 +100,7 @@ export class Relatives extends Subsection {
   validRelations = () => new RelativesValidator(this.props).validMinimumRelations()
```
suzubara: This seems like a good opportunity to deprecate some usage of the `RelativesValidator` class. Both `validMinimumRelations` and `validMaritalRelations` could be rewritten as functions.

christopherhuii:
This PR is dependent on EN-3819.

ryanhofdotgov:
EN-3819, which Chris points out this is dependent on, is in PR https://github.com/18F/e-QIP-prototype/pull/1772

ryanhofdotgov:
@christopherhuii Not sure if you've seen the conflict on this PR

christopherhuii:
This work is not relevant anymore. Going in a different direction because of policy

PR #1753, opened 2019-07-03
[EN-3773] Deprecate Validation Classes - part 1

suzubara:
## Description

This PR accomplishes the first 2 steps (listed below) needed to delete the legacy validation classes. Changes include:
- Export `validate` functions from validator files that were missing it
- Replace use of validation classes with new functions in `helpers/validation.js`
- Replace use of validation classes with new functions in `validators/validate.js` (this file is marked for deprecation)
- Ensure the `formType` is passed from state into `validateSection` function call (since some validators require the `formType` param)

---

Requirements to delete validation classes:
- [x] Replace validation class with function in validation helper
- [x] Replace validation class with function in legacy validate file (`this.props.validator`)
- [ ] Replace validation class with functions in Accordion components
- [ ] Replace validation class with functions in other misc instances
- [ ] Ensure test coverage in functions before removing classes & their tests
- [ ] Delete classes & tests

LATER:
- [ ] Add section key to all section components
- [ ] Remove validator prop from section components
- [ ] Section-level validation functions
- [ ] Remove other misc. validator classes

## Checklist for Reviewer

- [ ] Click through & fill out sections, verify that the valid status in the sidebar updates appropriately
- [ ] Load test data and log in, verify the sidebar and score card show the correct valid sections
- [ ] Review code changes

### DIFF
--- a/src/validators/bankruptcy.js
+++ b/src/validators/bankruptcy.js
@@ -15,7 +15,7 @@ const financialBankruptcyModel = {
   },
 }
 
-const validateFinancialBankruptcy = data => (
+export const validateFinancialBankruptcy = data => (
   validateModel(data, financialBankruptcyModel) === true
 )
 
@@ -31,7 +31,9 @@ export default class BankruptcyValidator {
    * Validates the yes/no branching for bankruptcy
    */
   validHasBankruptcy() {
-    return validateModel(this.data, { HasBankruptcy: financialBankruptcyModel.HasBankruptcy }) === true
+    return validateModel(this.data, {
+      HasBankruptcy: financialBankruptcyModel.HasBankruptcy,
+    }) === true
   }
 
   validBankruptcies() {
@@ -87,7 +89,9 @@ export class BankruptcyItemValidator {
   }
 
   validDischargeExplanation() {
-    return validateModel(this.data, { DischargeExplanation: financialBankruptcy.DischargeExplanation }) === true
+    return validateModel(this.data, {
+      DischargeExplanation: financialBankruptcy.DischargeExplanation,
+    }) === true
   }
 
   isValid() {
--- a/src/validators/credit.js
+++ b/src/validators/credit.js
@@ -18,13 +18,19 @@ const creditCounselingModel = {
   },
 }
 
+export const validateFinancialCredit = data => (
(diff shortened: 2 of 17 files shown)


src/validators/bankruptcy.js
```diff
@@ -31,7 +31,9 @@ export default class BankruptcyValidator {
    * Validates the yes/no branching for bankruptcy
    */
   validHasBankruptcy() {
-    return validateModel(this.data, { HasBankruptcy: financialBankruptcyModel.HasBankruptcy }) === true
+    return validateModel(this.data, {
```
sojeri: do we have something like prettier installed to automagically do this & enforce a standard?

src/validators/credit.js
```diff
@@ -18,27 +18,34 @@ const creditCounselingModel = {
...
 export default class CreditValidator {
   constructor(data = {}) {
     this.data = data
   }
 
   validHasCreditCounseling() {
-    return validateModel(this.data, { HasCreditCounseling: creditCounselingModel.HasCreditCounseling }) === true
+    return validateModel(this.data, {
+      HasCreditCounseling: creditCounselingModel.HasCreditCounseling,
+    }) === true
```
sojeri: really bummed that github isn't catching these as whitespace changes. I guess it's the trailing comma.

src/validators/gambling.js
```diff
@@ -38,19 +44,20 @@ export default class GamblingValidator {
    * Validates section of gambling debt
    */
   isValid() {
-    return validateModel(this.data, financialGamblingModel) === true
+    return validateFinancialGambling(this.data)
   }
 }
 
-const validateFinancialGambling = data => (
+export const validateGamblingItem = data => (
```
sojeri: per other method names, should this be `validateFinancialGamblingItem`?

src/validators/bankruptcy.js
suzubara: we do technically have a prettier config, although I think it currently conflicts with some of our eslint rules. Right now eslint happens to staged files as a pre commit hook, that's what we've been using so far. I agree it would be great to just have it taken care of with prettier but I think that's outside the scope of this PR :)

PR #1759, opened 2019-07-10
[EN-3834] Validation Refactor - nested model validations

suzubara:
## Description

This PR ensures that whenever `validateModel` is called from within a custom validator, it passes along both the validator-specific `options` param as well as the `globalOptions` param (which takes whatever options were passed into the originating `validateModel` call). This should allow passing validation parameters such as date limits and form-specific requirements into custom validators that call `validateModel`.

Also adds remaining date limit parameters that are based on other attributes.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/constants/dateLimits.js
+++ b/src/constants/dateLimits.js
@@ -6,8 +6,8 @@ import { DateTime } from 'luxon'
 
 const TODAY = DateTime.local()
 
-const DEFAULT_LATEST = TODAY
-const DEFAULT_EARLIEST = TODAY.minus({ years: 200, days: 1 })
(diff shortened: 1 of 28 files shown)


christopherhuii:
Do `requireTrue`, `requireEmpty`, or `zipcode` validators need the global options? It looks like those are the only ones that don't have them.

suzubara:
@christopherhuii Nope, only the validators that make a nested `validateModel` call need to explicitly pass `globalOptions` through.

christopherhuii:
looks good. pulled down into #1761 and options are getting passed through

PR #1774, opened 2019-07-22
Update purge script to reflect recent db change

ryanhofdotgov:
locked column was changed from bool to enum-ish text column a few weeks ago


ryanhofdotgov:
This is going to be removed in another PR in favor of dbreset.

PR #1777, opened 2019-07-22
Add docs for all dev tools. Remove the sql purge script.

macrael:
## Description

## Checklist for Reviewer

- [ ] Review code changes


macrael:
I couldn't find a reference to the purge script anywhere on confluence, so I think this is sufficient to clear it out. 

I think it's reasonable to remove the purge script, the flush script and the dbreset+dbmigrate scripts cover its use cases in a more managed way.

docs/advanced.md
```diff
@@ -133,8 +139,4 @@ Password: GENERATED-PASS
 Name: GENERATED-DB
 ```
 
-In a second terminal window, you can connect `psql` to the local port, and `cf` will proxy those commands to the remote cloud.gov instance. When prompted for a password, supply `GENERATED-PASS` from the `cf connect-to-service` output:
-
-```
-psql -h localhost -p GENERATED-PORT -U GENERATED-USER -W GENERATED-DB -f purge-all-user-data.sql
-```
+With this data, you can set the associated env vars and then modify the database using tools like `dbmigrate` or `dbreset`
```
ryanhofdotgov:
```suggestion
With this data, you can set the associated env vars and then modify the database using tools like `dbmigrate` or `dbreset`. However, note that `dbreset` will delete *ALL* existing account information. To reset data, while preserving existing accounts, use `psql` to retrieve a list of accounts and use `flush` to clear form data for each user individually.
```

macrael: I switched to having an example for flush, and added a warning to the dbreset docs

PR #1778, opened 2019-07-24
[EN-3845] Create SF85 test scenarios

joshmfranklin:
## Description

1. This PR creates three SF85 specific test scenarios.
2. This PR makes removes unnecessary validations from the SF86 that aren't a part of the SF85.
3. This PR makes updates to the `load-scenario` tool to allow the SF85 test scenarios to be loaded into eApp.

## # 1
`test1.json` is the simplest scenario of the full SF85 form going down as few branched sections as possible. 
`test2.json`is a complete blowout of the full SF85 form going down as many branched sections as possible.
`test3.json`is a complete blowout of the full SF85 form, similar to `test2.json`
*It is also worth mentioning `test2.json` and `test3.json` were created by two different people.

A new folder has also been created in to hold the new SF85 test scenarios: `api/testdata/complete-scenarios/SF85`

### How do I load these tests cases?
1. Make sure to run `make build-cmd`
2. Use `docker exec CONTAINERID sftype USERNAME SF_TYPE SF_VERSION` to update a user to have the SF85 form.
`SF_TYPE` should be `SF85`
`SF_VERSION` should be `2017-12-draft7`
3. After that, you should be set to follow the instructions [here](https://github.com/18F/e-QIP-prototype/blob/develop/docs/test-scenarios.md#loading-existing-test-json-files)

## # 2
Because the SF85 is a subset of the SF86, there are many fields/validations that needed to be removed. During that process, a few validations were overlooked. Those validations prevent the form from being fully validated because those fields aren't on the SF85 to be filled out. Test cases 2 and 3 in the PR will not pass without these changes. This worked is tracked by `EN-3856`. A summary of the issues I found are listed below:

| Section | Form Tracking Row |
| -------- | ------------------- | 
| Substance Use / Illegal Use of Drugs and Drug Activity / Usage | 680 |
| Investigative and criminal history / Police record / Offenses |  624-626 |
| Investigative and criminal history / Police record / Offenses | 639-640 |

## # 3
The `load-scenario` tool is reading from a catalog of sections/subsections. If any of those sections/subsections are missing from the JSON getting imported, it throws a fatal error. To resolve that, it uses `Printf()` instead `Fatalf()`.

## Notes
I think this is outside of the scope of the PR, but there is extraneous data in the SF85 test scenarios JSON. Unnecessary sections are properly excluded from the JSON, but subsections are not excluded from the SF85 JSON (they're empty). Please review the JSON files for further understanding.

~While filling out the form, the Redux store does not populate those subsections, nor are they sent in the payload when saving. At some point in the backend, it is creating those empty data structures. When logging back in, those empty data structures show up in Redux even though they weren't there previously.~

If this doesn't make complete sense, I can write a more thorough issue with pictures and GIFs.

## Checklist for Reviewer

- [ ] Review code changes


ryanhofdotgov:
@christopherhuii When you pick this up, would you rename the bare bones test1 – it would be nice to make test1 the bare bones test across all the form types.

ryanhofdotgov:
@christopherhuii Some example screenshots or links to line numbers to test files in the PR would be good to easily see what you are talking about re: empty sub-sections

Would you rollback your changes to load-scenario? I don't want the switch to Printf merged in because its a step backwards for loading 86 scenarios – it will load bad 86 scenario files when they should fail. That was just a suggested hack to get you moving forward for the short term.

In needs a separate issue/PR, but probably the best 1st pass at this is to add a field to the catalogue entries in application.go – an array that lists what form type/version pair the section should be used in. Then add analogous `Catalogue` call that takes in a form type/version. Then a change to load-scenario that reads the form type from the account to determine what catalog is should use.

christopherhuii:
I reverted the changes in load scenario.

I can provide more context around the JSON issues I'm seeing with the SF85.
Starting from a clean form: the Redux store is populated with this data
```
application: {
    Settings: {
      formType: 'SF85',
      {...}
    },
    Identification: {},
    Financial: {},
    Relationships: {},
    Citizenship: {},
    Military: {},
    History: {},
    Foreign: {},
    TBD: {},
    Legal: {},
    Psychological: {},
    Substance: {},
    Submission: {},
    Completed: {...},
    Errors: {},
    AddressBooks: {}
  }
```
The sections that aren't a part of the SF85  aren't really a problem, but it's a little bit over extra bloat that should get cleaned up at some point. These top-level sections are NOT in the JSON when exported.

In the SF85, Financial only has two subsections: Taxes and Delinquent Payments. As you click through each subsection, eApp saves the subsection it was previously on. However, once you visit the `Review` section page, eApp makes a `POST` request to `/me/save` for each subsection that's a part of the current section in the SF86. I think we might have missed something that handles how we save data.

When you sign back in, the Redux store will be populates with empty values because they were sent when navigating to the `Review` page.

Note: Taxes and Delinquent have `value: 'No'` because they were filled out, but the rest are empty.
```
Financial: {
      Bankruptcy: {
        HasBankruptcy: {
          value: ''
        },
        List: {
          branch: {
            type: ''
          },
          items: []
        }
      },
      Card: {
        HasCardAbuse: {
          value: ''
        },
        List: {
          branch: {
            type: ''
          },
          items: []
        }
      },
      Credit: {
        HasCreditCounseling: {
          value: ''
        },
        List: {
          branch: {
            type: ''
          },
          items: []
        }
      },
      Delinquent: {
        HasDelinquent: {
          value: 'No'
        },
        List: {
          branch: {
            value: ''
          },
          items: []
        }
      },
      Gambling: {
        HasGamblingDebt: {
          value: ''
        },
        List: {
          branch: {
            type: ''
          },
          items: []
        }
      },
      Nonpayment: {
        HasNonpayment: {
          value: ''
        },
        List: {
          branch: {
            type: ''
          },
          items: []
        }
      },
      Taxes: {
        HasTaxes: {
          value: 'No'
        },
        List: {
          branch: {
            value: ''
          },
          items: []
        }
      }
    },
```

christopherhuii:
It looks like the problem in the "empty" subsections saving is in `src/components/Section/sectionData.js`. If you look at `case 'SECTION_NAME/intro'` and `case 'SECTION_NAME/review'`, it contains all the subsections, even ones that don't belong to to SF85 or SF85P.

christopherhuii:
EN-3857 - Ticket to update load-scenario in order to accept SF85 complete scenario
EN-3860 - Ticket to remove subsections from JSON that aren't a part of the SF85/SF85P.

PR #1785, opened 2019-07-30
[EN-3844] Additional Comments section

suzubara:
## Description

- Adds Additional Comments section, navigation, saving in Redux, and sending in save API requests
- Adds backend support for saving Additional Comments section and returning it to the frontend
- Adds additional comments XML template
- Updates all complete scenario test cases to include updated JSON and XML. All have been validated against XSDs and e-QIP
- Test Case 3 has additional comments at the end of the form. I have made a note in `api/xml/xml_test.go`. All other test cases do not have additional comments.

## Checklist for Reviewer

- [ ] Review code changes


christopherhuii:
Do we need validations for the Additional Comments section?

To check if the branch is answered [Yes, No]. If Yes, check for a value in the comment textarea.

api/templates/additional-comments.xml
```diff
@@ -0,0 +1,5 @@
+<AdditionalComments Version="1" Type="Pooled">
```
ryanhofdotgov: I'd probably just put the whole thing in a conditional block that checks if Yes and comments is not empty string. It's a minor detail, but:
* that avoids unnecessary churn on test case files (so fewer things to have to test in eQIP, fewer known good files will have been modified)
* Skips empty element cruft
* Not sure if you've tested on e-QIP or not, but occasionally it complains about useless elements (but sometimes it ignores them too).

api/templates/application.xml
```diff
@@ -34,6 +34,7 @@
             <ValidationResponses Version="1" Type="FormInstanceSpecific" />
           </FormInstanceSpecific>
           <Pooled>
+            {{tmpl "additional-comments.xml" .Package.Comments}}
```
macrael: let's add this at the end of the list instead of the beginning so this list maintains its order relative to the actual sections

ryanhofdotgov:
@christopherhuii Would you re-submit any scenarios that have changed XML as a result of this PR to e-QIP and note that in a comment. Thanks.

christopherhuii:
@macrael @ryanhofdotgov I have updated the PR according to your comments.

ryanhofdotgov:
@christopherhuii Thanks – looks like test 3 is the only one that has XML changes now. Has it been re-submitted to e-QIP successfully?

christopherhuii:
@ryanhofdotgov Yes, that should be correct, right? Test 3 is the only test case that has additional comments. I can add comments to more test cases if we want more.

Yes, I tested all test cases again. Validates against XSDs and submitted successfully to e-QIP again.

api/additionalcomments.go
```diff
@@ -0,0 +1,75 @@
...
+			stack.Append("AdditionalComments", err)
+		}
+	}
+
+	return !stack.HasErrors(), stack
+}
+
+// ClearNoBranches clears any questions answered nos on a kickback
+func (entity *AdditionalComments) ClearNoBranches() error {
+	if entity.HasComments != nil {
```
macrael: Why do you need to set the value to "" here? That's what ClearNo() should do, I think

christopherhuii: Is this enough?
```
func (entity *AdditionalComments) ClearNoBranches() error {
	entity.HasComments.ClearNo()

	return nil
}
```

macrael:
If the test passes, that’s enough.

-MacRae

api/additionalcomments.go
macrael: Here's what ClearNo does:

```
// ClearNo resets a "no" answer to unset
func (entity *Branch) ClearNo() {
	if entity != nil {
		if entity.Value == "No" {
			entity.Value = ""
		}
	}
}
```
Maybe in the test you are trying to clear a Yes?

PR #1790, opened 2019-08-05
F/E session handling

suzubara:
## Description

This PR includes front end refactoring of the authentication flow in order to support the new secure cookie-based sessions, and to decouple async calls from UI code.
- Cleaned up Redux actions around auth and page load, removed async calls from them, added tests
- Added actions to control the session timeout flow and API calls
- Removed timeout logic and async calls from `TimeoutWarning` modal
- Remove logic from `Loader` component
- Added new `initApp` action and corresponding saga
- Added sagas to control app initialization (on page load), handle logged in and not logged in flows, and handle session timeout and logging out
- Added sagas to control fetch requests for status, form data, login, renew session, and logout
- Added error messaging to Login page (network error, SAML failed, invalid credentials, etc.) and error code constants on F/E & API response
- Remove deprecated token code
- Add generic error handling to API service, add `withCredentials` to authenticate calls with secure cookie

1 caveat: we need to re-implement the animation on the Loader screen when the form is done loading. Right now it does a straight redirect to the form. This will likely require react transitions to do correctly. There's a TODO left in the `Loader` component.

## Checklist for Reviewer

- [ ] Review code changes
- [ ] Login, session timeout, and logout flows should be thoroughly tested


src/sagas/initialize.js
```diff
@@ -68,11 +67,60 @@ export function* handleInitError(action) {
...
+    } else {
+      yield call(handleFetchFormSuccess, data, path)
+    }
+  } else {
+    console.warn('Missing response', response)
+    yield call(env.History().push, '/error')
+  }
+}
+
+/** This is a somewhat generic handler for API fetch failures */
```
suzubara: nice

src/components/Form/TimeoutWarning/TimeoutWarning.jsx
```diff
@@ -1,146 +1,84 @@
...
-      () => {
-        // Change the timer interval
-        this.resetInterval()
-        // Refresh the token
-        this.refreshToken()
-      }
-    )
+    // Save the form
+    saveSection(application, section.section, section.subsection, dispatch)
+    // TODO error handling
```
macrael: Does this need to happen for this PR, @suzubara ?

src/views/Form/Form.jsx
```diff
@@ -145,11 +125,34 @@ class Form extends React.Component {
...
           <SavedIndicator interval="30000" />
-          <TimeoutWarning timeout={env.SessionTimeout()} />
+          {showSessionWarning && <TimeoutWarning timeout={env.SessionTimeout()} />}
         </div>
       </div>
     )
   }
 }
 
+// TODO for some reason adding these is breaking this page (:
```
macrael: Does this matter? Should we just delete it?

src/views/Loading/Loading.jsx
```diff
@@ -1,44 +1,24 @@
...
-        getApplicationState(() => {
-          this.animateCloseTimeout()
-        })
-      )
-    })
+    this.setState({ spinner: true })
   }
 
-  animateCloseTimeout() {
+  // TODO - trigger this when load is done
```
macrael: We are pushing to the right route, is this no longer the place that needs to happen?

src/views/Login/Login.jsx
```diff
@@ -199,8 +217,9 @@ export class Login extends React.Component {
             </button>
             <a
               id="forgot-password"
-              href="javascript:;;"
-              title={i18n.t('login.basic.forgot.title')}>
+              href="#TODO"
```
macrael: What's this? Do we actually display any sort of forgot password flow?

src/services/api.js
```diff
@@ -1,130 +1,91 @@
...
-export const deleteCookie = name => {
-  const domain = window.location.hostname
-  Cookies.remove(name, { domain })
-  if (Cookies.get(name)) {
-    console.warn(
-      `${name} cookie couldn't be removed - check that domain matches, etc.`
-    )
-  }
-}
+export const getQueryValue = (queryString, key) => getSplitValue(key, queryString.substring(1), '&', '=')
```
macrael: this is fine, no reason to change it, but I'm surprised there's not a better solution for this than splitting and parsing on our own.

src/views/Form/Form.jsx
```diff
@@ -23,61 +24,78 @@ import { updateSection } from 'actions/SectionActions'
...
-    super(props)
-    this.state = { refreshPending: false }
-  }
-
   componentWillMount() {
     this.defaultRedirect()
   }
 
+  componentDidMount() {
+    this.refreshToken()
```
macrael: I think this is unnecessary. All api calls will refresh the session, so the only time we need to explicitly call /refresh is if someone clicks the "your session is about to expire" button. At the very least we should try not to make this call when we are also making a /save call, which is going to be the bulk of the time we interact with the page.

macrael:
I'm really loving the way sagas make the API calls simple

src/components/Form/TimeoutWarning/TimeoutWarning.jsx
suzubara: No, there wasn't any error handling in place to begin with here. I am going to make a followup issue for converting the saveSection flow into a saga to handle this.

src/views/Form/Form.jsx
suzubara: I don't want to delete it - components should have propTypes and defaultProps defined. So this is a TODO left to flag this file for further attention (but is unrelated to this PR).

src/views/Loading/Loading.jsx
suzubara: The remaining TODO here is the spinner used to animate once the loader was done, _before_ routing to the next page. I'm not sure how important keeping it is to the project, but I wanted to call it out since it is the only deprecation of current functionality in this PR.

src/views/Login/Login.jsx
suzubara: No, we have no forgot password flow. I added the `#TODO` to make it more obvious that this link doesn't do anything but I'm not sure if there is a plan for it or if we should just remove it altogether.

src/services/api.js
suzubara: the browser doesn't provide a native API for parsing queryString params so this kind of code is not uncommon. I'm not saying this is the best solution, but I was just making lint fixes and not changing the code itself.

src/views/Form/Form.jsx
suzubara: sounds good, I'll remove it.

src/views/Loading/Loading.jsx
macrael: oh right, there used to be a minimum 1s delay or something before moving on? That is probably a nice thing to have design wise

src/views/Login/Login.jsx
macrael: I vote we delete it. Dead code tells no tales.

src/services/api.js
macrael: that is surprising!

PR #1792, opened 2019-08-06
Session Management Backend

macrael:
## Description

This PR contains all the backend changes for the new session management work. It includes:

* Making sessions work with SAML
* Adding constraints to the sessions db
* Logging session lifecycle events per the STIG
* hooking up all the handlers to the sessions package
* HTTP level tests for session management

## Checklist for Reviewer

- [ ] Review code changes


api/http/saml.go
```diff
@@ -168,39 +150,46 @@ func (service SamlResponseHandler) serveAuthnResponse(encodedResponse string, w
...
-		Expires:  expiration,
-	}
-	http.SetCookie(w, cookie)
+	service.Cookie.AddSessionKeyToResponse(w, sessionKey)
+
 	http.Redirect(w, r, redirectTo, http.StatusFound)
 }
 
 func (service SamlResponseHandler) serveLogoutResponse(encodedResponse string, w http.ResponseWriter, r *http.Request) {
+	// TODO: validate the SAML Logout Response and tie it to a session via the SLO request ID
```
ryanhofdotgov: Can you clarify what you mean by that?

macrael: I'll expand on the TODO in the comments and make an issue for it. As is, we are not validating the logout response from WSO2. We should be:

1. verifying that the signature is correct
2. checking that the ID in the response is the same as the one we made the logout request for. This will confirm that the logout message was for the intended user.

1. is more important than 2, for sure.

src/services/api.js
```diff
@@ -26,7 +26,7 @@ const getSplitValue = (key, raw, delim1, delim2) => {
 }
 
 export const deleteCookie = name => {
-  const domain = process.env.COOKIE_DOMAIN || window.location.hostname
+  const domain = window.location.hostname
```
ryanhofdotgov: Can you walk me through the usage of this? This doesn't seem like it would work since the cookie would be scoped to backend; e.g., `api.eapp.example.com` and this will be executing in the frontend context?

api/mock/store.go
```diff
@@ -55,7 +56,7 @@ func (s *StorageService) DeleteAttachment(accountID int, attachmentID int) error
 }
 
 // CreateOrUpdateSession creates a new session record in the db
-func (s *StorageService) CreateOrUpdateSession(accountID int, sessionKey string, expirationDate time.Time) error {
+func (s *StorageService) CreateOrUpdateSession(accountID int, sessionKey string, sessionIndex sql.NullString, expirationduration time.Duration) error {
```
ryanhofdotgov: Now that we have `ExtendAndFetchSessionAccount`, can this just be `CreateSession` instead of `CreateOrUpdateSession` (in name and in implementation)?

api/session/session.go
```diff
@@ -1,50 +1,93 @@
...
+		return api.Account{}, api.Session{}, fetchErr
+	}
+
+	return account, session, nil
 }
 
 // UserDidLogout attempts to end the session and returns an error on failure
 func (s Service) UserDidLogout(sessionKey string) error {
-	return s.store.DeleteSession(sessionKey)
+	delErr := s.store.DeleteSession(sessionKey)
```
ryanhofdotgov: Do you have a strategy in mind for where to call `DeleteSession` for general clean up? i.e., avoid perpetual expansion of the session table with stale rows?

src/services/api.js
macrael: This code is removed in the FE branch, sorry for the confusion.

api/mock/store.go
macrael: Yes, I think I will break it out so that it is more explicit when someone logs into an existing session.

api/session/session.go
macrael: I'll make sure that we are deleting rows from this table when DELETE account is called.

macrael: But the real answer is no, I didn't really have a plan for this. The table is limited to the total number of users in the system so once the system has enough users that this is a problem it should be trivial to delete the old ones.

src/services/api.js
ryanhofdotgov: No worries – under this revised paradigm, under what circumstances does the cookie get deleted?

macrael:
Right now it’s not being deleted. Logout just deletes the session, not the cookie. That could be changed so that we don’t have a spurious “request made with invalid session” log line.

ryanhofdotgov:
I've added EN-3863 to track the desire to delete the cookie, not just expire the session.

PR #1793, opened 2019-08-08
add docs on how to configure the FE and BE to serve SSL

macrael:
## Description

Short doc on how to configure SSL in dev.

This should probably be a make command, so if we'd rather make a task for that I can think about how to do that instead of just modifying config files. 

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/docs/https.md
+++ b/docs/https.md
@@ -1,0 +1,16 @@
+# Instructions on running eapp over SSL in dev
+
+1. Configure the api server to use SSL: 
+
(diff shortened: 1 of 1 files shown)


docs/https.md
```diff
@@ -0,0 +1,16 @@
+# Instructions on running eapp over SSL in dev
```
ryanhofdotgov:
```suggestion
# Instructions on running eapp over SSL – for local development only
```

PR #1795, opened 2019-08-09
[EN-3793] Update US Passports to use local state to control suggestions model

christopherhuii:
## Description
This PR fixes the issue where applicants aren't able to exit the "Alternate names" modal by clicking `Use a different name instead`. This modal shows up if an applicant has indicated they have had another name in the Identification section.

This fix leverages local component state rather than manipulating data and sending it up and down the component tree.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/components/Form/Suggestions/Suggestions.jsx
+++ b/src/components/Form/Suggestions/Suggestions.jsx
@@ -1,60 +1,63 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import { newGuid } from '../ValidationElement'
 import Modal from '../Modal'
 
 export default class Suggestions extends React.Component {
-  constructor(props) {
-    super(props)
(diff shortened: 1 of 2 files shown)


src/components/Form/Suggestions/Suggestions.jsx
```diff
@@ -1,60 +1,65 @@
...
-              <i className="fa fa-arrow-circle-right" />
-            </button>
-          </div>
+  suggestions = () => {
+    const {
+      suggestions, suggestionLabel, renderSuggestion, suggestionUseLabel,
+    } = this.props
+    return suggestions.map((suggestion, i) => (
+      <div
+        className="suggestion" key={`suggestion-${i}`}
```
christopherhuii: I don't love that I'm using the index of the current element in the `.map`, but I do think it's a step up from the completely random "guid" that's getting generated. I don't see another unique identifier I could use.

PR #1799, opened 2019-08-12
[3867] CSRF Protection

macrael:
## Description

This PR includes all the support for CSRF attack protection. We use gorilla csrf to generate a csrf and send it to the client as a header on the /status call. It's stored in redux and then sent on all POST calls back to the server. Gorilla verifies the token is present as part of all POST calls. 

Gorilla stores the token to verify it with as an HTTPOnly cookie. 

## Checklist for Reviewer

- [ ] Review code changes


macrael:
This PR includes the "remove JWT code" PR's code as well, it would be excellent if that were merged first before this.

src/sagas/initialize.js
```diff
@@ -91,6 +92,14 @@ export function* handleInitSuccess(action, path = '/form/identification/intro')
     yield put(updateApplication('Settings', 'status', Status))
     yield put(updateApplication('Settings', 'hash', Hash))
 
+    const { headers } = response
+    const csrfToken = headers['x-csrf-token']
```
ryanhofdotgov: I haven't dived deeply here before – what mechanism ensures the headers are all lower case at this point?

api/settings.go
```diff
@@ -189,6 +189,12 @@ const (
 	// Default: *not enabled*
 	UspsAPIKey = "USPS_API_API_KEY"
 
+	// CSRFSecret Random tokens used for CSRF protection are generated using a secret random key of at least 256-bits.
```
ryanhofdotgov: Per Slack convo, correct comment here

docs/CONFIGURATION.md
```diff
@@ -265,6 +266,13 @@ United States Postal Service (USPS) API key for address validation.
 **Target** - Back-end (api)<br>
 **Default** - *not enabled*<br>
 
+## `CSRF_SECRET`
+
+Random tokens used for CSRF protection are generated using a secret random key of at least 256-bits. For example, `openssl rand -base64 32` generates an appropriate key. If this value is not specified, one will be automatically generated unique to the instance.
```
ryanhofdotgov: Per Slack convo, update description of this key here.

src/sagas/initialize.js
macrael: It's a property of Axios, the promise http requests library we are using.

ryanhofdotgov:
@macrael Does this work with the non-SAML authentication workflow? I got a login failure ("no CSRF on POST") – digging deeper now.

If so, it's not the end of the world, as I've been wanting to ditch this basic password flow forever and always use it with WSO2 or some lightweight SAML shim.

ryanhofdotgov:
Screen Shot 2019-08-13 at 5 00 44 PM

These browser console errors are from:
1) using the simple, local authentication (not SAML)
2) then clicking logout

Again, I haven't dived deeply yet, but I think the /auth/basic error is benign, I think the /me/logout is a legit error – my cookie doesn't get deleted. I think some of the issue is from JWT --> cookie switch, and maybe some of it is this CSRF PR.

macrael:
I can't reproduce the /logout error. It's working fine for me.

PR #1811, opened 2019-08-21
[EN-3878] Remove link for unnecessary google fonts

christopherhuii:
## Description
This PR removes the dependencies hosted by 3rd parties for reasons mentioned in `EN-3878`.

We don't actually need to import the font faces ourselves because they are included in the USWDS styles we use. See [here](https://github.com/uswds/uswds/tree/develop/src/fonts)

Please confirm that everything looks as expected and `Merriweather` and `Source Sans Pro` fonts are still being used.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -54,9 +54,6 @@
       <script type="text/javascript" src="/html5shiv.min.js"></script>
       <script type="text/javascript" src="/respond.min.js"></script>
     <![endif]-->
-    <link
-      href="https://fonts.googleapis.com/css?family=Merriweather:700|Source+Sans+Pro:400,700"
-      rel="stylesheet">
   </head>
   <body>
     <table class="print-table">


ryanhofdotgov:
@christopherhuii Would you check IE11 and verify it looks good there?

christopherhuii:
It looks like my VM is failing to work. I'm not sure why, but it redirects me to the `/error` page. However, the error page has text on it!

I think this should suffice, but both fonts load on the `/error` page.

Screen Shot 2019-08-22 at 4 39 11 PM
Screen Shot 2019-08-22 at 4 39 46 PM

christopherhuii:
Screen Shot 2019-08-23 at 9 41 49 AM

Screen Shot 2019-08-23 at 9 42 18 AM

PR #1813, opened 2019-08-21
[EN-3104] Update domestic phone numbers so they can't start with 

christopherhuii:
#  Description
This PR updates the United States (domestic) phone number field so that the phone number cannot start with 0. This also makes a small revision to the error locale so that it makes sense. However that text should be confirmed and approved.

This ONLY focuses on the United States (domestic) phone number as presented in `EN-3104`.
## Checklist for Reviewer

- [ ] Review code changes


michaelccata:
@christopherhuii LGTM

PR #1817, opened 2019-08-24
Update XML generation for 85/85P

macrael:
## Description

Change the templates to generate valid 85/85P xml. 

This PR has everything for 85 in it and half of what is necessary for 85P.

To correctly run the omit_sections_tests you need to install a full set of XSDs in ./api/integration/xsd

## Checklist for Reviewer

- [ ] Review code changes


api/templates/foreign-travel.xml
```diff
@@ -15,6 +15,7 @@
     {{range $index, $item := .props.List.props.items}}
     {{with $Item := $item.Item}}
     <Travel ID="{{inc $index}}">
+      {{ if formType "SF86" }}
```
ryanhofdotgov: ContactedByForeignMilitaryOrganization is present in the 85P – need to move the `if` statement down one element.

api/templates/financial.xml
```diff
@@ -1,9 +1,13 @@
 <FinancialRecord Type="Pooled" Version="1">
-  {{tmpl "financial-bankruptcy.xml" .Financial.Bankruptcy}}
-  {{tmpl "financial-card.xml" .Financial.Card}}
-  {{tmpl "financial-credit.xml" .Financial.Credit}}
+  {{ if formType "SF86,SF85P" }}
```
ryanhofdotgov: Something I'm mulling over (i.e., not asking for a change) – would it be better off for the individual template files to have the `if` block. If you consider them objects/components, the knowledge of where they should be used is maintained at that level, rather than its parent. 

In particular, cases like `financial-credit.xml` where there is an `SF86` conditional would be more clear, as you could just see in that same file that it was already restricted to the `SF86` and `SF85P`.

api/testdata/complete-scenarios/test6.xml
```diff
@@ -977,13 +977,13 @@
               <Investigations>
                 <Investigation ID="1">
                   <Agency>Unknown</Agency>
+                  <InvestigationDate>
```
ryanhofdotgov: These golden file changes look benign, but just to be sure, the 3 SF86 test scenarios that changed should be submitted to e-QIP test and a note on the PR should be added that that was done.

api/templates/legal-investigations-investigated.xml
```diff
@@ -13,6 +13,20 @@
...
     {{end}}
     {{end}}
+
+    <InvestigationDate>
+      {{monthYearOptional $Item.Completed $Item.CompletedNotApplicable}}
+    </InvestigationDate>
+
+    {{ if formType "SF85" }} 
+    <ClearanceGranted>
+      <Answer>{{$Item.ClearanceGranted}}</Answer>
```
ryanhofdotgov: I think this needs to be `{{branch $Item.ClearanceGranted}}`?

api/xml/xml_test.go
```diff
@@ -132,7 +132,10 @@ func TestPackage(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.Schema, func(t *testing.T) {
 
-			result, err := service.DefaultTemplate(test.Schema, test.Data)
+			app := api.BlankApplication(-1, "SF86", "2017-07")
```
ryanhofdotgov: Can we add complete test scenario tests for the 85 and 85P? There are some already in the repo in subdirectories under `complete-scenarios`.

ryanhofdotgov: (Not on this line of course – just used it as a spot to hang my comment on.)

api/templates/foreign-travel.xml
macrael: nice catch!

api/xml/xml_test.go
macrael: I tried this, there's enough places where XML is being inserted at the top level (see foreign.xml for instance) that trying to put the rules in the sub-sections ends up being redundant. I think I'm going to leave things as is rather than making it so sometimes you have to check the top level and sometimes you don't.

api/templates/financial.xml
macrael: I tried this, there's enough places where XML is being inserted at the top level (see foreign.xml for instance) that trying to put the rules in the sub-sections ends up being redundant. I think I'm going to leave things as is rather than making it so sometimes you have to check the top level and sometimes you don't.

PR #1823, opened 2019-08-29
[EN-3914] Fix AlternateAddress Telephone validation

christopherhuii:
## Description
For context, for Self Employment, Other Federal, State Government, Federal Contractor, Non Government, and Other employment types, applicants are asked to fill a physical address. This physical address contains a location as well as a phone number.

When we refactored the validations, we used the same model for physical addresses and alternate addresses. They are only _mostly_ the same. Physical addresses in employment have a `Telephone` field that alternate addresses don't.

This PR adds an `option` to allow the `Telephone` validation to be enabled/disabled depending on the use case.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/models/civilUnion.js
+++ b/src/models/civilUnion.js
@@ -65,14 +65,22 @@ const civilUnion = {
     if (attributes.Address && isInternational(attributes.Address)) {
       return {
         presence: true,
-        model: { validator: physicalAddress, militaryAddress: true },
+        model: {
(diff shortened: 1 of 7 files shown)


christopherhuii:
I'm not exactly sure why code coverage went down. The frontend delta is minimal, but it looks like backend coverage dropped? I don't understand why since I didn't make any backend changes.

PR #1826, opened 2019-08-29
En 3886 allow applicant identification to enter either city or county for birthplace

leonged:
## Description

En 3886 allow applicant identification to enter either city or county for birthplace

## Checklist for Reviewer

- [ ] Review code changes


src/validators/identificationbirthplace.js
```diff
@@ -1,7 +1,15 @@
 /* eslint-disable import/prefer-default-export */
 import { validateModel } from 'models/validate'
-import identificationPlaceOfBirth from 'models/sections/identificationPlaceOfBirth'
+import birthplace from 'models/shared/locations/birthplace'
```
suzubara: This change should be made in the `models/sections/identificationPlaceOfBirth` file instead of this one.

src/validators/location.js
```diff
@@ -278,7 +278,10 @@ export default class LocationValidator {
           valid = valid && this.validStreet()
           break
         case 'city':
-          valid = valid && this.validCity()
+          if (this.layout === Layouts.BIRTHPLACE)
```
suzubara: Shouldn't this be checking for `this.layout === Layouts. IDENTIFICATION_BIRTH_PLACE`?

src/models/shared/locations/birthplace.js
```diff
@@ -1,10 +1,32 @@
 import locationModel from '../location'
+import { isPO, isUS, isInternational } from 'helpers/location'
```
suzubara: it looks like only `isInternational` is used here so the other two imports are unnecessary

leonged:
> Just a couple of minor comments. Ultimately I would also tend towards introducing a prop like requireCityAndCounty to the existing Layouts.BIRTHPLACE instead of adding an entire new layout, but that is a higher level decision and not a blocking change request.

There is requireCity and requireCounty seperately

src/validators/location.js
leonged: Yes, corrected, but I dont' think this code is called anywhere anymore

src/models/shared/locations/birthplace.js
leonged: Removed

src/validators/identificationbirthplace.js
leonged: Ok, updated, but can you show me where/how this gets called?

src/components/Form/Location/ToggleableLocation.jsx
```diff
@@ -167,7 +167,8 @@ export default class ToggleableLocation extends ValidationElement {
               onError={this.onError}
               onFocus={this.props.onFocus}
               onBlur={this.props.onBlur}
-              required={this.props.required}
+              required={this.props.layout === Layouts.IDENTIFICATION_BIRTH_PLACE ? this.props.required && !this.props.county : this.props.required}
+              requireCity={this.props.layout === Layouts.IDENTIFICATION_BIRTH_PLACE ? this.props.required && !this.props.county : this.props.required}
```
suzubara: it doesn't look like this prop is used by the `City` component

src/components/Form/Location/ToggleableLocation.jsx
```diff
@@ -184,7 +185,8 @@ export default class ToggleableLocation extends ValidationElement {
               onError={this.onError}
               onBlur={this.props.onBlur}
               onFocus={this.props.onFocus}
-              required={this.props.required}
+              required={this.props.layout === Layouts.IDENTIFICATION_BIRTH_PLACE ? this.props.required && !this.props.city : this.props.required}
+              requireCounty={this.props.layout === Layouts.IDENTIFICATION_BIRTH_PLACE ? this.props.required && !this.props.city : this.props.required}
```
suzubara: same comment as above but for the `County` component

PR #1829, opened 2019-08-31
Restore USPS validation

ryanhofdotgov:
In https://github.com/18F/culper/pull/1815/commits/b486569bc624ace533224585b767fee020e0d8ed the API behavior for `/me/validate` was unintentionally changed, resulting in USPS geocoding errors returning a web service level error when they should return an HTTP OK and a JSON response body.

This PR reverts that and cleans up some of the logging related to USPS geocoding calls.

### DIFF
--- a/api/integration/validation_test.go
+++ b/api/integration/validation_test.go
@@ -64,7 +64,7 @@ func TestValidateHandlerInvalidAddress(t *testing.T) {
 	api.Geocode = mock.Geocoder{}
 
 	// Pass a bad address
-	location := readTestData(t, "../testdata/bad-address.json")
+	location := readTestData(t, "../testdata/nonstandardized-address.json")
 
 	w, r := standardResponseAndRequest("POST", "/me/validate", strings.NewReader(string(location)), account)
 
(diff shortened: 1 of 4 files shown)


api/integration/validation_test.go
```diff
@@ -64,7 +64,7 @@ func TestValidateHandlerInvalidAddress(t *testing.T) {
 	api.Geocode = mock.Geocoder{}
 
 	// Pass a bad address
-	location := readTestData(t, "../testdata/bad-address.json")
+	location := readTestData(t, "../testdata/nonstandardized-address.json")
```
kilbergr: This is great for testing the nonstandard addresses to make sure validation is still working as expected for those.
Please also write a test to consider when a bad address (such as what was in bad-address.json) is attempted.

ryanhofdotgov: @kilbergr That's already covered by the existing tests; take a look at `TestValidateHandlerBadEntity`, `TestCanNotValidateSomethingElse`.  The main problem here was that the implementation was changed incorrectly and the test was changed to match the incorrect behavior.

kilbergr: OK I see what you're saying. Thanks for addressing.

PR #1832, opened 2019-09-03
EN-3877 (Strict Transport Security Header). Added a new middleware pa…

kdolan-soliel:
## Description

Fix #3877
Added a new middleware package 'security_headers.go' to the http directory. This sets the Strict-Transport-Security header to each response when the api variable 'DevUseInsecureCookie' is set to false.

## Checklist for Reviewer

- [ ] Review code changes


api/cmd/server/main.go
```diff
@@ -122,9 +122,11 @@ func main() {
 
 	// Inject middleware
 	caching := http.CacheHandler{Log: logger}
+	security := http.SecurityHandler{Log: logger, Env: settings}
 	cors := http.CORSHandler{Log: logger, Env: settings}
 	logging := http.LoggingHandler{Log: logger}
-	router := caching.Middleware(cors.Middleware(logging.Middleware(r)))
+	//router := caching.Middleware(cors.Middleware(logging.Middleware(r)))
```
macrael: This line can be deleted now.

ryanhofdotgov: Minor nit, but would you remove this comment reflecting the previous version? Git will track that for us.

api/settings.go
```diff
@@ -352,7 +352,8 @@ const (
 	// Default: ""
 	IndentJSON = "INDENT_JSON"
 
-	// DevUseInsecureCookie wether to use a cookie with Secure set to false in development.
+	// DevUseInsecureCookie controls whether to use a cookie with Secure set to false in development.
```
macrael: I think this makes sense to combine these two behaviors behind the one env var. I think we should probably rename it though. How about DEV_DISABLE_SSL ? 

You'll need to update this both in settings.go and in docs/CONFIGURE.md

Finally, I think that we should use this new environment variable in http/server.go to switch wether we are configuring TLS, instead of just checking to see if the TLS_{CERT,KEY} variables are present. We should throw an error then if DEV_DISABLE_SSL is not set but TLS_CERT is not present.

macrael:
Also, looks like the build is failing with an extraneous "fmt" import.

ryanhofdotgov:
I'm going to slightly disagree with @macrael and say let's do his recommended `DEV_DISABLE_SSL` changes under a different issue/PR. I'll write up the issue and @kdolan-soliel can pick it up as a follow-on task.

macrael:
Ok, I'll approve this once the build failure is fixed, then.

kdolan-soliel:
At this point I've already changed the code to use the DEV_DISABLE_SSL, but haven't tested it yet.

kdolan-soliel:
I checked-in revised code:

1.  Changed DEV_USE_INSECURE_COOKIE to DEV_DISABLE_SSL to provide a more generic name for the flag. It is now used to set the secureCookie in main.go and to set the Strict-Transport-Security header.
 
2.  Added DEV_DISABLE_SSL environment variable to the various Env files. It might be a good idea to check the values I set for each.  Env files changed:
- .env.dev
- .env.example
- .env.production
- .env.staging
- .env.test

3.  Corrected compile error (unused "fmt" import). 

4.  Removed commented-out code.

macrael:
Can you update the docs in docs/CONFIGURE.md as well?

And can you add a test? There's a good example for a middleware test in cors_test.go called "TestAllowedOriginSuccess" it does more than is necessary here but should point you in the right direction for just checking that the middleware does add the header we desire.

kdolan-soliel:
Will do.

kdolan-soliel:
I've created the test file for setting the header.  It's taken awhile because I was setting up a lot of things to make Go work.  That leads to a problem I'm having and I hope you can offer some advice.  When I try to test the test, the compilation fails due to one file:  syslog.go (in github.com/logrus/hooks/syslog).  This file will only work in Linux/Unix systems and has a build constraint (!windows) that blocks it from being compiled.  Are you aware of any alternatives or some other way to check the test?

api/cmd/server/main.go
kdolan-soliel: This line has been deleted.

kdolan-soliel: This has been removed

api/settings.go
kdolan-soliel: Fixed

kdolan-soliel:
Pushed Fix 3877.  Added securityheaders_test.go to verify the setting of the DEV_DISABLE_SSL flag. Updated the CONFIGURATION.md file to include the new DEV_DISABLE_SSL flag.

PR #1833, opened 2019-09-03
[EN-3921] Update test case 3 so direct control validates

christopherhuii:
## Description
This PR updates the complete scenario test case 3 to validate the `Foreign activities -> Direct control` section. This PR only fixes the `Direct control` subsections. This test case is still not 100% valid.

Please see comments in the diff for explanations for each fix.

Note:
XML changes validate against e-QIP

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/api/testdata/complete-scenarios/test3.json
+++ b/api/testdata/complete-scenarios/test3.json
@@ -2638,7 +2638,7 @@
                                     "layout": "Address",
                                     "street": "Apple St",
                                     "city": "Berlin",
-                                    "zipcode": "12333",
+                                    "zipcode": "",
                                     "country": "Germany"
                                   }
                                 },
@@ -2767,9 +2767,9 @@
                   "Relinquished": {
                     "type": "datecontrol",
                     "props": {
-                      "month": "04",
-                      "day": "04",
-                      "year": "1998",
+                      "month": "",
+                      "day": "",
+                      "year": "",
                       "estimated": false
                     }
                   },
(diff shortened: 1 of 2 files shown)


api/testdata/complete-scenarios/test3.json
```diff
@@ -2638,7 +2638,7 @@
                                     "layout": "Address",
                                     "street": "Apple St",
                                     "city": "Berlin",
-                                    "zipcode": "12333",
+                                    "zipcode": "",
```
christopherhuii: I removed the zip code because international addresses do not have zip codes. The generated XML does not include the zip code `12333` that was previously in the test case.

My suspicion is that this data slipped in before we started to clear out irrelevant data when switching between branches. It was never caught as an error by e-QIP because it's not in the generated XML.

api/testdata/complete-scenarios/test3.json
```diff
@@ -2767,9 +2767,9 @@
                   "Relinquished": {
                     "type": "datecontrol",
                     "props": {
-                      "month": "04",
-                      "day": "04",
-                      "year": "1998",
+                      "month": "",
```
christopherhuii: For this `Relinquished` field, it has both a relinquished date as well as the relinquished date marked as not applicable. This was probably a case of data improperly getting cleared out then `Not applicable` was clicked.

api/testdata/complete-scenarios/test3.json
```diff
@@ -2905,7 +2905,7 @@
                     "props": {
                       "month": "01",
                       "day": "6",
-                      "year": "1995",
+                      "year": "2000",
```
christopherhuii: The year was before the initial date so there is an error.

PR #1834, opened 2019-09-03
Close the DB after every test

macrael:
## Description

We've seen issues running `make test-go` where the db refuses to connect because there are too many connections. I was unable to reproduce that reliably, but we have not been closing out our db connections after every test. This PR adds a deferred db.Close() to every test that uses the db. 

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/api/database.go
+++ b/api/database.go
@@ -35,4 +35,5 @@ type DatabaseService interface {
 	Count(model interface{}, condition string, params ...interface{}) int
 	CountExpr(model interface{}, expr string, retval interface{}, condition string, params ...interface{})
 	Array(model interface{}, expr string, retval interface{}, condition string, params ...interface{})
+	Close() error
 }
--- a/api/integration/add_empty_value_test.go
+++ b/api/integration/add_empty_value_test.go
@@ -8,6 +8,7 @@ import (
 
 func TestAddEmptyValue(t *testing.T) {
 	services := cleanTestServices(t)
+	defer services.closeDB()
 	account := createTestAccount(t, services.db)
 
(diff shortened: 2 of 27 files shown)


ryanhofdotgov:
@macrael Is there a follow-on issue that needs to be opened to for db connections for non-test code? 

I only have one or two data points right now, but I think you can reliably cause the tests to fail if you simultaneously have the app running (and log in, use the form a bit), then run the tests.

macrael:
I don't think that the connection close code will ever actually run in this case since the call to ListenAndServe never returns. 

I believe that Postgres is smart enough to close connections from processes that are killed but I haven't tested that.

macrael:
@ryanhofdotgov ^

ryanhofdotgov:
@macrael I haven't dived deeply, but https://github.com/go-pg/pg claims to have automatic connection pooling. In general for a connection pool to serve its intended purpose (i.e., avoid connect/disconnect/connect overhead for both TLS and the database) you want to close the connection in application space rather than the db closing it because the process ended abruptly.

I'll open up an issue to look at it more closely.

macrael:
@ryanhofdotgov I believe that go-pg manages opening/closing connections for the pool for individual db calls automatically. The only Close() function provided is for closing the entire pool, not meant to be called for individual connections. 

From the docs: "It is rare to Close a DB, as the DB handle is meant to be long-lived and shared between many goroutines"

Since we only create a single pool in server/main.go, and that function never returns, I think it's not possible or necessary for us to close the db there. We're only being bit here because we are making a new connection in every single integration test.

PR #1835, opened 2019-09-04
Rek unpackaged tests

kilbergr:
## Description

To create tests for "low hanging fruit" of unpackaged files. Not really all that low-hanging anymore.
This includes a set of entity tests that trigger the unmarshal function on all entities and throw first a json.Unmarshal error and then individual field errors.

## Checklist for Reviewer

- [ ] Review code changes


api/entity_test.go
```diff
@@ -0,0 +1,297 @@
...
+			_, err := payload.Entity()
+			if err == nil {
+				t.Logf("Should have received an error from payload.Entity, instead got nil for type %v", payload.Type)
+				t.Fail()
+			}
+		}
+	}
+}
+
+func selectLinesString(lines []string) []int {
```
macrael: So, the purpose of this hack is to no change the type of anything that is in a collection? What happens when you do that? Does that mean that items that only exist inside of collections don't have their unmarshal error lines called?

kilbergr: Yes that is what that means. So all of an entity's fields are called and trigger errors, but the items within the collections don't trigger errors. This is actually ok because we do trigger the collection type error.

PR #1840, opened 2019-09-05
[EN-2383] Prohibit applicants over age of 100

sasiannlee:
## Description

This PR changes the max age validation in identification date of birth from 130 to 100.

To reproduce this issue, when entering into e-QIP an applicant over 100 years old, e-QIP does not prohibit applicants over age of 100 instead it prohibits only when age is over 130. 

Implement the following logic (extracted from the e-QIP Validation Matrix):
Date must be greater than or equal to NOW-100. Date must be less than or equal to NOW.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/components/ErrorList/ErrorList.scss
+++ b/src/components/ErrorList/ErrorList.scss
@@ -59,4 +59,8 @@
       display: inline-block;
     }
   }
+
+  &.usa-alert-warning {
+    background-size: 3rem;
+  }
 }
--- a/src/components/ErrorMessage/index.jsx
+++ b/src/components/ErrorMessage/index.jsx
@@ -1,11 +1,18 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import classnames from 'classnames'
 
 const ErrorMessage = ({
-  id, title, message, note,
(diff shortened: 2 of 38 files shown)


ryanhofdotgov:
@sasiannlee Just to clarify since I don't think it's written down in any of the repo documentation, but Truss has been following the model where after PR approval, the submitting dev does the merge of the PR and deletes the branch. This way if something unusual happens like a failing build after the merge, the submitting dev gets altered is on the hook to fix.

Also, I'm guessing you rebased off of develop after branching for the feature? I think for a cleaner PR, after rebasing you would do a `git push --force` to your feature branch. That way the PR would only show your deltas.

sasiannlee:
@suzubara on a separate (perhaps maybe related ticket) 
**EN-3633** No validation error on year when user type "1000" on year (this is assigned to you)
**EN-2000** 1000 for Year field
**EN-2975** Date control validates the year 1000 even if before birthday

This validation bug appears to still be an issue and looks like you are assigned to EN-3633. I was curious to know if this bug fix in the same area, or is there more to the changes in the issues I listed above.

PR #1841, opened 2019-09-05
[EN-3927] Fix missing country bug

suzubara:
## Description

To reproduce the issue:
1) Log in as a brand new user with no data (B/E sends no form data, everything is empty)
2) Navigate to a section such as People who know you well. Observe that the phone number fields time of day value defaults to “Both”. Do not interact with any fields
3) Navigate to another section. This triggers a save call to the B/E, which saves an “empty” version of the section that you were just one, which contains empty string values for the country, timeOfDay, etc. This is key
4) Log out/log back in. Now the B/E sends that empty data back in and the form is pre-filled with it, and those empty strings overwrite the defaults for country, timeOfDay, etc. If you go to Relationships / People, now you will see that the phone number time of day field has no default selected.

To resolve, I updated the `Location.go` struct so that it does not send an empty string value for Country fields. This was overwriting the default prop of United States in Location React components, and explicitly setting country to be an empty string.

Additionally just to be safe, I added logic to the update functions for Location components so that if a location value is being updated and is missing a country value and is displaying the domestic country UI, it always sends the United States value as a default.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/api/location.go
+++ b/api/location.go
@@ -39,14 +39,14 @@ const (
 // Location is a basic input.
 type Location struct {
 	ID              int    `json:"-"`
-	Layout          string `json:"layout"`
+	Layout          string `json:"layout,omitempty"`
 	Street1         string `json:"street,omitempty"`
 	Street2         string `json:"street2,omitempty"`
 	City            string `json:"city,omitempty"`
 	State           string `json:"state,omitempty"`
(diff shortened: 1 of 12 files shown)


christopherhuii:
Country bug appears to be fixed. Time of day probably needs a JIRA issue.

PR #1843, opened 2019-09-06
Fix Print Screen

christopherhuii:
## Description
This PR fixes the print screen. Previously after submitting, the print screen would just show the side bar with a blank reading well.

The console error was:
```
Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in. Check the render method of `PackagePrint`.
```

In line 141, the `sectionComponents` object doesn't have a key for `REVIEW_AND_SUBMIT` so the app was trying to render an `undefined` component. To alleviate that, I set a conditional in the `.map` to check if the component exists before trying to render it. I'm not sure if `REVIEW_AND_SUBMIT` should be added to the `sectionComponents` object since it's not really a section in the form. In addition to that, PDF generation will probably replace this code.

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/components/Section/Package/Print/index.jsx
+++ b/src/components/Section/Package/Print/index.jsx
@@ -139,16 +139,16 @@ export class PackagePrint extends React.Component {
     const { printed, attachments } = this.state
 
     const sectionComponents = {
-      [`${sections.IDENTIFICATION}`]: IdentificationReview,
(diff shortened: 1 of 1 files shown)


ryanhofdotgov:
For expediency's sake, skipped the usual process of @christopherhuii clicking the merge button.

PR #1844, opened 2019-09-06
Remove the postgresql package in favor of simplestore

macrael:
## Description

To simplify our db situation, this PR removes the rest of the old go-pg based database code. The last stuff that it was doing related to accounts has been implemented in simplestore using the new patterns. Getting rid of the last vestiges of the db code feels good!

There are a couple things left to do:
* add Transmission support to simplestore
* re-engage the Account.validate code on account save.

## Checklist for Reviewer

- [ ] Review code changes


macrael:
@ryanhofdotgov I've got to go to bed and be done. This PR is big and good, however, the merge into the mainline is still big and to be done. I've gotten started on it but I'm not done. I would really like to get this in before we roll off, but understand if we want to code freeze instead. 

I think that if you're planning to cut a full release and do a merge then it's probably worth waiting on this code, I should probably be the person to do the merge since I've been thinking about it. But I'd love for you to review it and I can address comments when I'm back. 

Take Care,

api/simplestore/accounts.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	return account, nil
+}
+
+// SetAccountPasswordHash sets the password hash for an account
+func (s SimpleStore) SetAccountPasswordHash(account api.Account) error {
+
+	// delete an existing one, if it exists
+	deleteQuery := `DELETE FROM basic_auth_memberships WHERE account_id = $1`
```
ryanhofdotgov: It's not super important since this is code that is just used for dev, but I'd put these 2 statements in a transaction.

PR #1849, opened 2019-09-10
Fix 85P/test2.json to validate against the XSD

macrael:
## Description

test2.json is not validating against the XSDs. There are probably some UI issues to fix that Ryan is going to file, but this gets that file into shape. 

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/api/testdata/complete-scenarios/SF85P/test2.json
+++ b/api/testdata/complete-scenarios/SF85P/test2.json
@@ -4003,7 +4003,7 @@
                           "touched": true
                         }
                       },
-                      "present": true
+                      "present": false
                     }
                   },
                   "Name": {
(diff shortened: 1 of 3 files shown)


ryanhofdotgov:
@macrael I just did a quick audit and there's at least another place in the XSD there the `to` in the date range doesn't permit `Present`, so we should also make the XML code more defensive/explicit and have a specialized template that doesn't try to include it. I'll open up a separate issue for that as well. It also revealed another place in the UI that probably has it wrong.

api/testdata/complete-scenarios/SF85P/test2.json
```diff
@@ -7985,8 +7985,7 @@
                   "Officer": {
                     "type": "radio",
                     "props": {
-                      "name": "officer-na",
-                      "value": "NotApplicable",
+                      "value": "Enlisted",
```
ryanhofdotgov: Just to clarify: the XSD error got a little confused because of the same name `Rank`, but Enlisted is not actually a rank, but since this field is free form, it's just as good as any other string.

macrael:
In the error that the XSD generated it said that “Elisted”, “Officer” were the only two valid options, but that might have been a weird error.

PR #1854, opened 2019-09-12
updated webpack.config.js to support watch on windows machines

leonged:
## Description

allows windows machine and windows docker container to detect changes to frontend code and rebuild
should be reviewed on a windows machine

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -9,6 +9,12 @@ const HtmlWebpackPlugin = require('html-webpack-plugin')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 
 module.exports = {
+  watch: true,
+  watchOptions: {
+    aggregateTimeout: 300,
+    poll: 1000,
+    ignored: './src/node_modules/'
+  },
   mode: production ? 'production' : 'development',
   entry: {
     polyfills: './src/polyfills.js',


leonged:
> You'll note that the code coverage report above shows that the coverage on your branch drops by 5.43% as compared to develop.  I take this as a sign that the build is failing in come respect, most likely as I described.
> I'd like to use PowerShell for everything, but I don't think it's practical under the circumstances.  The 3.81 version of Make for Windows (which is the latest) was last updated in 2006, putting it well behind the current release of the GNU Make utility.  You should consider installing Cygwin to provide a Linux interface.  It you do, just select the make package; installing everything will take up too much disk space and not be useful.
> In the meantime, your branch cannot be merged until we figure out what's going on.  I'll get somebody here to try it and see if he gets the same results that I did, but I'd like you to also look into the issue.

I'm not sure why the code coverage would decrease, probably something we can ask about on Monday. You're right, make shouldn't be relevant. The stuff gets built inside docker containers. The frontend is built with 'docker-compose run --rm js yarn build-js'
What version of docker are you using? I'm running docker desktop community 2.1.0.2

kdolan-soliel:
Sorry I didn't get back sooner, but I've been trying to come up with a explanation for this.  I thought this might be a line termination issue--we had a problem with that earlier--but that does not appear to be the case.

I had another person on our team run the build and it also failed for him.  My develop branch builds without problems, but if I cut and paste your changes into develop, the build fails.  I commented out the lines you added in your branch and the build succeeds.  If I expose just the first line (watch: true) then the build fails, but without the heavy CPU and Ethernet usage we've seen when the watchOptions are also present.

Have you had any luck?

ryanhofdotgov:
@leonged I'm going to close this PR without merging and let you pick it up again as you see fit in the downstream NBIS project.

PR #1858, opened 2019-09-13
En 2383 confirmed date not validating properly in accordion/review summary

leonged:
## Description
addresses a bug described in slack thread https://gsa-tts.slack.com/archives/C9VSX3JNM/p1568312780370000 thought to be caused by original EN-2383 of changing age limit to 100 in identification date of birth, although I think this bug was present regardless of changing the age limit

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/models/sections/__tests__/identificationDateOfBirth.test.js
+++ b/src/models/sections/__tests__/identificationDateOfBirth.test.js
@@ -42,7 +42,7 @@ describe('The identification date of birth section', () => {
       .toEqual(expect.arrayContaining(expectedErrors))
   })
 
-  it('requires a birthdate greater than 16 years', () => {
+  it('requires a confirmation if birthdate less than 16 years', () => {
     const currentYear = new Date().getFullYear()
     const testData = {
       Date: {
@@ -52,7 +52,7 @@ describe('The identification date of birth section', () => {
         estimated: false,
       },
     }
-    const expectedErrors = ['Date.date.date.datetime.DATE_TOO_LATE']
+    const expectedErrors = ["Confirmed.presence.REQUIRED"]
 
     expect(validateModel(testData, identifcationDateOfBirth))
       .toEqual(expect.arrayContaining(expectedErrors))
@@ -78,7 +78,6 @@ describe('The identification date of birth section', () => {
         day: '1',
         year: '2018',
         estimated: false,
-        name: "birthdate",
       },
       Confirmed: {
         checked: true,
@@ -87,7 +86,26 @@ describe('The identification date of birth section', () => {
 
     expect(validateModel(testData, identifcationDateOfBirth)).toBe(true)
   })
+/* TODO validate checked property to ensure it is set to true
+  it('does not allow an age less than 16 birthdate if confirmed not checked', () => {
+    const testData = {
+      Date: {
+        month: '1',
+        day: '1',
+        year: '2018',
+        estimated: false,
+      },
+      Confirmed: {
+        checked: false,
+      }
+    }
 
+    const expectedErrors = ["Date.date.date.datetime.DATE_TOO_LATE"]
+
+    expect(validateModel(testData, identifcationDateOfBirth))
+      .toEqual(expect.arrayContaining(expectedErrors))
+  })
+*/
   it('requires a birthdate less than 100 years and 1 day even if confirmed', () => {
     const testData = {
       Date: {
@@ -95,7 +113,6 @@ describe('The identification date of birth section', () => {
         day: '1',
         year: '1905',
         estimated: false,
-        name: "birthdate",
       },
       Confirmed: {
         checked: true,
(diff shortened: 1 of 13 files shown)


christopherhuii:
I quickly did some manual testing and I don't think the validations are functioning the way we want them to, but I'll defer to @michaelccata or @ryanhofdotgov.

My understanding is that we **SHOULD NOT** allow applicants to be over 100 years old because it is not accepted by e-QIP. With the proposed implementation in this PR, applicant's can override the validation as long as they confirm that the DOB is correct.

See this error I got from e-QIP when submitted DOB as `1/1/1990`:
```
WARN[0004] gov.opm.eqip.ws.webservice.ApplicantDataException: SF86_5_S01-S04-IdentifyingInformation: Provided "Your date of birth" must not be earlier than 100 years before today's date, 9/13/1919.
```

michaelccata:
@christopherhuii that's correct. We should not allow anyone to submit an age over 100.

@leonged can you make the update? We can keep the checkbox for people under 16.

leonged:
> @christopherhuii that's correct. We should not allow anyone to submit an age over 100.
> @leonged can you make the update? We can keep the checkbox for people under 16.
@michaelccata @christopherhuii yup, i addressed that in 3633, this ticket was just to update the accordion validator by updating the model

src/models/sections/__tests__/identificationDateOfBirth.test.js
```diff
@@ -70,4 +70,22 @@ describe('The identification date of birth section', () => {
 
     expect(validateModel(testData, identifcationDateOfBirth)).toBe(true)
   })
+
+  it('allows an "invalid" birthdate if confirmed', () => {
```
christopherhuii: name of the test needs to be updated because birthdays over 100 are not allowed.

src/models/sections/__tests__/identificationDateOfBirth.test.js
```diff
@@ -70,4 +70,22 @@ describe('The identification date of birth section', () => {
...
+        year: '1900',
+        estimated: false,
+      },
+      Confirmed: {
+        checked: true,
+      }
+    }
+    const expectedErrors = ['Date.date.date.datetime.DATE_TOO_LATE']
+
+    expect(validateModel(testData, identifcationDateOfBirth)).toBe(true)
```
christopherhuii: I don't think this should validate because the applicant is over 100.

src/models/sections/__tests__/identificationDateOfBirth.test.js
leonged: currently the model allows anything to pass if the new field confirmed is true, i think we can add the restrictions in EN-3633 since that ticket is dealing with the age limit 16-100 constraints

christopherhuii: Not quite sure I understand. EN-3633 does not mention of anything about age minimums or limits. It's specifically referencing a use case where there is no error then the year is `1000`.

This PR was created to fix a defect that was introduced by #1840.

leonged: Ya, EN-3633 was 1000 but that's where I discovered the age limit issue, which is related to entering a year 1000, because entering a 1000 year means you're over the age of 100, which should error, but wasn't. Perhaps we should create a separate issue?

christopherhuii:
@leonged I'm not sure if I'm understanding the scope of this PR and #1859 and how they compliment each other. I'm still having a few issues with this PR, but look like they're resolved in #1859. The best thing to do here might be to combine this PR and #1859 because the files are directly conflicting each other.

In addition to that, I found this bug:
1. Enter birthdate `1/1/1900`
It still asks `I confirm this date is correct`. It shouldn't. I believe this is fixed in #1859 though

2. Refresh the page
3. Notice that the date is now `1/1/09`

leonged:
> @leonged I'm not sure if I'm understanding the scope of this PR and #1859 and how they compliment each other. I'm still having a few issues with this PR, but look like they're resolved in #1859. The best thing to do here might be to combine this PR and #1859 because the files are directly conflicting each other.
> In addition to that, I found this bug:
> 
> 
> Enter birthdate 1/1/1900
> It still asks I confirm this date is correct. It shouldn't. I believe this is fixed in #1859 though
> 
> 
> Refresh the page
> 
> 
> Notice that the date is now 1/1/09

@christopherhuii merged my other branch into this one

christopherhuii:
looks good to me

error messages are consistent with DOB year 99, 1000, 1900
applicant cannot have age 100+
applicant needs to confirm 16 and under
something we shouldn't do now, but might consider for the future is insrc/components/Form/DateControl/DateControl.jsx. maybe the year field we should change the min back to 1000 then add a minLength to 4 or something. or enforce that the length needs to be 4.

if it doesn't have 4 characters/numbers, the error message will read, "there aren't enough digits" or something of that variety. something to think about, but this PR looks good.

PR #1859, opened 2019-09-13
EN-3633 fix birthdate component validation to not

leonged:
## Description

addresses a bug where component validation allows '1000' to be entered in the identification date of birth year

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/components/Form/Number/Number.jsx
+++ b/src/components/Form/Number/Number.jsx
@@ -9,6 +9,7 @@ export default class Number extends ValidationElement {
     this.state = {
       uid: `${this.props.name}-${super.guid()}`,
       max: props.max,
+      min: props.min,
     }
 
     this.handleError = this.handleError.bind(this)
@@ -19,9 +20,10 @@ export default class Number extends ValidationElement {
       return
     }
 
-    const old = this.state.max
-    this.setState({ max: next.max, value: next.value }, () => {
-      if (old !== next.max) {
+    const oldmax = this.state.max
+    const oldmin = this.state.min
+    this.setState({ max: next.max, min: next.min, value: next.value }, () => {
+      if (oldmax !== next.max || oldmin !== next.min) {
         this.handleValidation(
           {
             fake: true,
@@ -114,6 +116,7 @@ Number.defaultProps = {
   disabled: false,
   value: '',
   max: '',
+  min: '',
   prefix: '',
   pattern: '^(\\s*|\\d+)$',
   errorCode: null,
--- a/src/components/Section/History/dateranges.js
+++ b/src/components/Section/History/dateranges.js
@@ -3,6 +3,7 @@
  * Date object or null.
  */
 export const extractDate = dateObj => {
+  
   if (dateObj instanceof Date) {
     return dateObj
   }
@@ -10,8 +11,10 @@ export const extractDate = dateObj => {
   if (!dateObj || !dateObj.month || !dateObj.day || !dateObj.year) {
     return null
   }
+  var d = new Date(`${dateObj.month}/${dateObj.day}/${dateObj.year}`)
+  d.setFullYear(dateObj.year) // addresses an issue where two digit years are assumed to be in 20th or 21st centry, ex: 99 -> 1999 01 -> 2001
 
-  return new Date(`${dateObj.month}/${dateObj.day}/${dateObj.year}`)
+  return d
 }
 
 /**
@@ -189,7 +192,7 @@ export const validDate = date => {
   const y = parseInt(year || 0)
 
   return (
-    y > 1000 &&
+    y >= 0 &&
     y < 10000 &&
     (m > 0 && m < 13) &&
     (d > 0 && d <= daysInMonth(m, y))
--- a/src/components/Section/History/dateranges.js
+++ b/src/components/Section/History/dateranges.js
@@ -189,7 +189,7 @@ export const validDate = date => {
   }
   const m = parseInt(month || 0)
   const d = parseInt(day || 0)
-  const y = parseInt(year || 0)
+  const y = parseInt(year || -1)
 
   return (
     y >= 0 &&
--- a/src/components/Section/Identification/ApplicantBirthDate/ApplicantBirthDate.jsx
+++ b/src/components/Section/Identification/ApplicantBirthDate/ApplicantBirthDate.jsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import { i18n } from 'config'
-import { DateTime } from 'luxon'
+import { TODAY } from 'constants/dateLimit'
 import {
   Field, DateControl, Show, Checkbox,
 } from 'components/Form'
@@ -170,7 +170,7 @@ export class ApplicantBirthDate extends Subsection {
     )
   }
 }
-const TODAY = DateTime.local()
+
 ApplicantBirthDate.defaultProps = {
   Date: {},
   Confirmed: {},
--- a/src/components/Section/Identification/ApplicantBirthDate/ApplicantBirthDate.jsx
+++ b/src/components/Section/Identification/ApplicantBirthDate/ApplicantBirthDate.jsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import { i18n } from 'config'
-
+import { DateTime } from 'luxon'
 import {
   Field, DateControl, Show, Checkbox,
 } from 'components/Form'
@@ -133,6 +133,8 @@ export class ApplicantBirthDate extends Subsection {
             name="birthdate"
             {...this.props.Date}
             relationship="Self"
+            minDate={TODAY.minus({ years: 100, days: 1 })}
+            maxDate={TODAY.minus({ years: 16 })}
             overrideError={(this.props.Confirmed || {}).checked}
             onUpdate={this.updateDate}
             onError={this.handleError}
@@ -156,7 +158,7 @@ export class ApplicantBirthDate extends Subsection {
     )
   }
 }
-
+const TODAY = DateTime.local()
 ApplicantBirthDate.defaultProps = {
   Date: {},
   Confirmed: {},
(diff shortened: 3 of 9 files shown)


src/config/locales/en/error.js
```diff
@@ -79,6 +79,21 @@ export const error = {
         'Please confirm the date is correct by checking the box below.'
       ],
       note: '',
+      min: {
+        title: 'The applicant age is not approved',
+        message: [
+          'Your date of birth indicates you are over the age of 100.',
```
christopherhuii: Let's try adding this line.
```
Your date of birth indicates you are over the age of 100. Please contact your security office representative or the individual who initiated your investigation regarding further instruction.
```

See [this](https://gsa-tts.slack.com/archives/C9VSX3JNM/p1568658106033800?thread_ts=1568657409.032200&cid=C9VSX3JNM)

src/components/Form/Number/Number.jsx
```diff
@@ -19,9 +20,10 @@ export default class Number extends ValidationElement {
       return
     }
 
-    const old = this.state.max
-    this.setState({ max: next.max, value: next.value }, () => {
-      if (old !== next.max) {
+    const oldmax = this.state.max
```
christopherhuii: This is an instance where local state can be removed from a component. I'm not completely convinced that local state is required in this component. However, this is refactor is probably outside the scope of this PR. I just wanted to point out this anti-pattern in the codebase.

christopherhuii:
Something I find a little strange is that the error message changes from year `1000` to `1001`. Should we just have one error message? Is there a reason why it changes?

Screen Shot 2019-09-16 at 12 52 54 PM


Screen Shot 2019-09-16 at 12 52 47 PM

leonged:
> Something I find a little strange is that the error message changes from year 1000 to 1001. Should we just have one error message? Is there a reason why it changes?

@christopherhuii ya, I didn't know what to do, basically the Number is giving the error when < 1001 and the BirthDate DateControl is giving the age error I think. The root of the issue is that the DateControl validates a date with year 1000 or less, because maybe the date object behind it doesn't like it and gives null which validates in the model, I couldn't figure out why. Maybe you have some insight? I plan to ask Suz about that during her office hours.

christopherhuii:
Well, before we figure out how or why this is happening, we need to understand what should be happening first.

Should the applicant see different errors for different years?

leonged:
> Well, before we figure out how or why this is happening, we need to understand what should be happening first.
> Should the applicant see different errors for different years?

I don't have an authoritive answer for that. I think it's ok and easy for the applicant to identify the mistake, but having one message is simpler. I don't know what eQip is currently like. Before this bug was identified, there were two messages, one when year was 999 or less, and one where 1001 or more. The current fix leaves the error messages as is.

ryanhofdotgov:
@leonged I think with our conversation with Brian, we've got a good idea of what the requirements are for kids under 16, and folks less than 100, and folks older than 100.

After all the discussion, we've not really uncovered/confirmed any requirements for a special error message for year 1000 or less. It seems like this was an arbitrary rule that a programmer decided to add.

At this point, I would rip it out, and just make sure the validation conforms to the rules that have been vetted by the product owner. However, for due diligence, I would use the blame function and do a little bit of software archeology to find out when and why this code for 1000 year validation got added, just to make sure there isn't something we haven't considered. 

In this case, the odds are extremely low we will find something important, but this is a process that is generally applicable as you uncover odd, unintuitive behavior. There's a lot of weird, one-off things that are mandated by the SF-86 and policy.

ryanhofdotgov:
@leonged  If it is not in the validation matrix, not in the content guide, the product owner doesn't think it is required, an e-QIP test indicates that it is not required, it's not one of the one or two new things that ESBO has asked eApp to deviate from e-QIP – it's very likely a hold over from the prototype days 2+ years ago when eApp was built without having access to these resources. Those should be far and few between, as the past year was spent excising them from the code base.

leonged:
Ok cool thanks for the feedback. I will just show the age limit message for any errors.

leonged:
@christopherhuii updated to have single error message for age limit

src/components/Section/History/dateranges.js
```diff
@@ -3,15 +3,18 @@
...
 export const extractDate = dateObj => {
+  
   if (dateObj instanceof Date) {
     return dateObj
   }
 
   if (!dateObj || !dateObj.month || !dateObj.day || !dateObj.year) {
     return null
   }
+  var d = new Date(`${dateObj.month}/${dateObj.day}/${dateObj.year}`)
```
christopherhuii: Is there a reason why this is changed? My understanding was that this PR is for the applicant's birthdate, but this change has to do with something else. Can you please clarify what this fixes?

src/components/Section/Identification/ApplicantBirthDate/ApplicantBirthDate.jsx
```diff
@@ -156,13 +170,13 @@ export class ApplicantBirthDate extends Subsection {
     )
   }
 }
-
+const TODAY = DateTime.local()
```
christopherhuii: Can we move this inside of the component so that it's not in the global scope? This way, it'll be scoped within the component and it'll be defined because the variable is used.

src/components/Section/History/dateranges.js
leonged: this fixes an issue with two digit years, so, if I were to enter 99 into the year textbox, it would validate since the date year will be 1999, it gets kinda screwy to the user I think, so this will make it the year 0099, in some cases like 15 becomes 2015 instead of 1915 so I think it's just confusing all over, I think this addresses that

src/components/Section/Identification/ApplicantBirthDate/ApplicantBirthDate.jsx
leonged: done

src/components/Section/Identification/ApplicantBirthDate/ApplicantBirthDate.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import { i18n } from 'config'
-import { TODAY } from 'constants/dateLimit'
+import { TODAY } from '../../../../constants/dateLimit'
```
christopherhuii: i think it should be `constants/dateLimits`. You also need to `export` the `const` from `constants/dateLimits`.

leonged: ya, I just fixed it, sorry
I forgot that I had exported it in my other branch and not this one

src/components/Section/History/dateranges.js
ryanhofdotgov: @leonged I'm just skimming this PR, but this doesn't sound like the type of change that should be done in this PR. A separate issue should be made, someone thinking about the product/usability should consider whether it is a good change to make, etc.

I can almost guarantee that if eApp automatically changes `99` to `1999` someone spent some time thinking about it, possibly even doing usability tests with users.

leonged: I don't think eApp is doing it intentionally, I think that's how JavaScript date objects are... so previously it would give a date too far into the past error when something < 1000 was entered for year... so now that we decided it should just give a "age too old error", people can enter 99 and expect to get an age too old error, but that is not the case if we don't make this change since JavaScript dateobject will be 1999

christopherhuii: I don't think this should be included in this PR, especially since it hasn't been addressed with usability testing or required by the PO. If we want to add this kind of feature to eApp, it needs to be consistently addressed throughout the application.

I also don't think it works as intended. Maybe I'm executing it wrong..
// const d = new Date()
// d.setFullYear('01')
// Wed Sep 19 0001 08:36:08 GMT-0752
// d.setFullYear('99')
// Sat Sep 19 0099 08:36:08 GMT-0752

I think we're getting close enough to `2020` where `20` can mean '0020', `1920` or `2020`. `1920` can represent a birth date for a relative (mother or father, for example) OR it can represent the year `2020`, when someone get married recently (in the near future). From the example above, i suspect entering `20` will mean `0020`, but I don't know the exactly JS spec.

leonged: That is how it's intended. Do you have a minute to chat?

leonged:
Closed - code merged into EN-2383 PR #1858

src/components/Section/History/dateranges.js
christopherhuii: Okay @ryanhofdotgov. I just caught up with @leonged and I misunderstood the intent of that code. I'll continue to test it and see what comes out of it.

When a user enters the month as `10`, day as `21`, and year as `99`, we expect the year to be `0099`. When Javascript creates that date with `99`, it converts `99` to `1999`. So in this PR the `d.setFullYear('99')` sets it as `0099`.

The reason why he did this was to trigger the more specific error message (age must be 16-100; you're too old), rather than the less specific (this year is too far in the past)

christopherhuii: However, I think going forward, it might be worth thinking about or looking into enforcing date years to have 4 digits. That's not happening now.

leonged: I think that's good, we can create a ticket to have the option of enforcing a length on the numbers, I don't know if that exists currently. We could also update the text to say enter DOB in mm/dd/yyyy format.

christopherhuii:
looks good to me
- error messages are consistent with DOB year 99, 1000, 1900
- applicant cannot have age 100+
- applicant needs to confirm 16 and under

something  we shouldn't do now, but might consider for the future is in`src/components/Form/DateControl/DateControl.jsx`. maybe the year field we should change the `min` back to `1000` then add a `minLength` to `4` or something. or enforce that the length needs to be  `4`.

if it doesn't have 4 characters/numbers, the error message will read, "there aren't enough digits" or something of that variety. something to think about, but this PR looks good.

PR #1862, opened 2019-09-16
[EN-3934] Fix refresh api to allow extending sessions about to expire

macrael:
We were getting an error where when the "you're session is about to expire" dialog appeared, clicking the refresh button led to logging out instead of extending the session. By default, the gorilla CSRF HTTPOnly cookie sets its path to be the path of the request where it is set. I believe that is whatever request first invokes the middleware, which in the case of basic auth is /me/status. 

Since the path is set to /me/, our calls to /refresh were not sending the cookie up, which is why the /refresh call was erring. That error caused the front end to logout. 

This PR changes the path of the /refresh api to /me/refresh so it sits at the same level as the other state modifying requests. This gets it covered by the CSRF cookie and the refresh button works as expected. 

Just to reduce the number of possible future errors, we could set the path of the CSRF cookie explicitly, as well. 

## Checklist for Reviewer

- [x] Review code changes
- [ ] Verify change works with SAML
- [x] Verify change works in IE browser


kdolan-soliel:
The fix provided by MacRae works:  clicking on the 'Still Working' button takes the user back to the page he was working on.  Since I'm the new kid, I won't comment on whether Ryan's suggestion is better or not.

However, in testing, I encountered another login-related bug, but it appears to be a separate issue (it also appears in the current develop branch), so I'll enter a new Jira ticket for it.

macrael:
Switched to just changing the Path of the cookie. This works.

ryanhofdotgov:
@macrael Thanks – good to go!

PR #1866, opened 2019-09-18
Modified the clear command so it wouldn't effect the country string f…

owyche:
…unction and removed the return from the update name function that ignored user input

This branch fixed the bug on the cohabitant page that would through an error when the "This is my spouse" button was clicked when the cohabitant had the same name as the users spouse/ Additionally when the "This is not my spouse button was clicked the user was no longer allowed to change the cohabitants name .

## Checklist for Reviewer

- [ ] Review code changes

### DIFF
--- a/src/components/Section/Relationships/RelationshipStatus/Cohabitant.jsx
+++ b/src/components/Section/Relationships/RelationshipStatus/Cohabitant.jsx
@@ -58,7 +58,7 @@ export default class Cohabitant extends ValidationElement {
     const state = {
       Name: {},
       Birthdate: null,
-      BirthPlace: null,
+      BirthPlace: {country: undefined},
       ForeignBornDocument: null,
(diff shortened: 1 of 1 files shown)


src/components/Section/Relationships/RelationshipStatus/Cohabitant.jsx
```diff
@@ -58,7 +58,7 @@ export default class Cohabitant extends ValidationElement {
     const state = {
       Name: {},
       Birthdate: null,
-      BirthPlace: null,
+      BirthPlace: {country: undefined},
```
christopherhuii:
```suggestion
      BirthPlace: {},
```

Don't need to add that because `this.props.BirthPlace.country` will be `undefined` against an empty object. Please make sure you have your IDE/editor setup with ESLint.

PR #1868, opened 2019-09-18
[EN-3969] Increase F/E test coverage

suzubara:
## Description

This PR:
- Deletes dead/legacy code that is not being used any more
- Adds some low hanging fruit tests where there were none
- Finish moving section model definitions from `src/validators` to new files in `src/models` and add corresponding tests

## Checklist for Reviewer

- [ ] Review code changes


ryanhofdotgov:
When this gets picked up for review, can we do a more hand-on review than might have been done normally? i.e., for each section where validation was changed, in a fresh form, test out that the validation is still working as expected. Thanks!

ryanhofdotgov:
Also: as you are doing the hands-on review, be checking the JavaScript console for any exceptions, etc.

leonged:
I've started reviewing this, do you have any recommendations on how I should approach this? I'm running through the application form start to finish. I'm finding issues but hard to tell if they are caused by changes or if they are bugs before the changes. I've been cross checking it to see if the same issue is on the deployed version at https://eqip-prototype-dev.fr.cloud.gov/

suzubara:
@leonged Because the majority of changes in this PR are adding new tests and removing dead code, there are no intentional functional changes, so if you are finding issues beyond filling out the form at a basic level (such as fatal JS errors, preventing a page from rendering, crashing the app completely, etc.) it is very likely those issues exist on `develop` already. I would focus on verifying the deleted code is not in fact referenced anywhere and reviewing the new tests to verify they reflect the expected functionality.

The other significant change was moving any validation models that were still located in `src/validators` files into `src/models` files. If any of these were moved incorrectly, the top-level form section would not return errors when validating, so in order to review this you could make sure that Redux contains errors for all sections when loading an empty form.

leonged:
@suzubara I did a flush on test01 which I believe clears out their form data. I don't see any errors in redux 
image

suzubara:
@leonged Sorry, I misspoke. If you login as a new user with an empty form, navigate to each of the form sections without filling out anything. This will save each section as an empty structure to the backend, and when you refresh or log back in, they should all be populated in Redux.

leonged:
@suzubara ok, i'm seeing it now, i'll approve

PR #1876, opened 2019-09-25
[WIP] Deprecate src/config/locales/i18n.js

christopherhuii:
## Description
This PR deprecates the homegrown implementation if `i18n` located in `src/config/locales/i18n.js`. Instead of the homegrown version, we have started replacing it with open source software. This new implementation is initialized in `src/util/i18n.js`.

Migrating all `i18n` imports to `util/i18n` has started, but has not fully deprecated the old method.

### WIP
99% of instances have been committed, but there is still one outstanding area that needs to be visited. `src/components/Form/Field/Field.jsx` still uses the old `i18n` import. It's not as straight forward because it appears the error adds additional markup that should not be present.

The new `i18n` method also doesn't like the fact that there are missing keys in the locale files. If there's a missing key, it will display the unreadable `i18n` key. The easy, but tedious way to alleviate this is to ensure that all keys are in the locale files, but it's not sustainable in the long run. Be sure to click through the application, trigger validations/errors, in order to find all the missing keys.

It breaks snapshots and should be reviewed a little more closely before deprecating `src/config/locales/i18n.js`

## Checklist for Reviewer

- [ ] Review code changes


ryanhofdotgov:
Replaced with PR #1886 – not quite ready to say I'll never look into this yet.

PR #1889, opened 2020-01-09
Bump birthdates a year for 2020

ryanhofdotgov:
These date-oriented tests aren't written well and are tied to the present year from the system clock vs a fixed clock being passed into the test. This is a minimum viable change to get the tests to pass again. 

If it's not properly fixed, the tests will have to be changed again in 2021, etc.


ryanhofdotgov:
There are likely quite a few similar date values that need to be updated in the SF-85, SF-86, etc. complete test scenarios as well, as that portion of the test suite is brittle with respect to date calculations.

